Issue #139, opened 2026-06-26
Adopt public helper APIs from build123d-drafting-helpers v0.12.1


pzfreo:
## Status
Upstream helper work is complete:

- `pzfreo/build123d-drafting-helpers#169` is closed.
- `pzfreo/build123d-drafting-helpers#170` merged on 2026-06-26.
- The public helper contracts shipped in `build123d-drafting-helpers` `v0.12.1`, published on 2026-06-26.
- Release: https://github.com/pzfreo/build123d-drafting-helpers/releases/tag/v0.12.1

## Problem
`draftwright` still imports private names from `build123d-drafting-helpers`, including:

- `_full_cyls` and `_spec_key` from `build123d_drafting.features`
- `_TB_COL_FRACTIONS` from `build123d_drafting.helpers`

Private upstream APIs can change without a compatibility guarantee, making ordinary dependency upgrades risky.

## Public replacements
Use the stable APIs added in `build123d-drafting-helpers` `v0.12.1`:

- `_full_cyls(cyls)` -> `full_cylinders(cyls)`
- `_spec_key(hole)` -> `HoleSpec.from_hole(hole)`
- `_TB_COL_FRACTIONS` and local title-block cell geometry -> `TitleBlock.cell_bbox(name)` or `TitleBlock.drawn_by_cell_bbox()`

`full_cylinders` and `HoleSpec` are exported from the package root. The title-block bbox helpers are instance methods on `TitleBlock` and return build-frame bounding boxes.

## Suggested fix

- Bump the minimum `build123d-drafting-helpers` dependency to `>=0.12.1`.
- Replace all private imports and call sites with the public APIs above.
- Prefer direct replacement over a compatibility layer unless `draftwright` still needs to support helper versions older than `0.12.1`.
- Update or add focused tests around hole grouping, cylinder filtering, and title-block drawn-by placement.

## Acceptance criteria

- No imports from private `build123d_drafting` names remain in `draftwright`.
- The helper dependency requires a version that provides the public APIs.
- Existing tests pass.
- Any changed title-block placement behavior is covered by a regression test or verified fixture output.

## References

- `src/draftwright/make_drawing.py`
- `src/draftwright/annotate.py`
- `src/draftwright/_core.py`
- https://github.com/pzfreo/build123d-drafting-helpers/issues/169
- https://github.com/pzfreo/build123d-drafting-helpers/pull/170

Issue #138, opened 2026-06-26
Refactor drawing pipeline architecture and module boundaries


pzfreo:
## Problem
This should not be a line-count refactor. The large files are a symptom, not the design target:

- `src/draftwright/make_drawing.py`: ~3,900 lines
- `src/draftwright/annotate.py`: ~2,250 lines

The deeper maintainability problem is that draftwright's build pipeline has several real architectural stages, but many of them currently communicate through `Drawing`'s private mutable dictionaries and helper state:

- `_named`
- `_anno_view`
- `_pinned`
- `_pattern_callouts`
- `_patterned_holes`
- `_build_issues`
- `_dropped_callout_diams`
- `_view_edge_cache`
- `_analysis`

`Drawing` is currently both the public editable object and the internal state bus for projection, annotation ownership, lint, repair, repack measurement, tables, balloons, and export. That makes unrelated changes harder to isolate and makes future feature work depend on private incidental state.

## Architecture observed in the code
The code already wants to be a deterministic drawing compiler pipeline:

```text
input part / STEP
  -> geometry + feature analysis
  -> sheet planning: scale, page, view blocks, strips, zones
  -> projection assembly: front / plan / side / iso / section views
  -> annotation passes: holes, envelopes, turned diameters, PMI, tables
  -> critique: lint_summary / lint issues
  -> repair: deterministic mechanical fixes
  -> export: SVG / DXF / PDF
```

Relevant current anchors:

- `_analyse()` builds the geometry/layout model in `make_drawing.py`.
- `Analysis` in `_core.py` is the typed build-time contract.
- `_assemble()` projects views, annotates, and fits the iso view.
- `_auto_annotate()` in `annotate.py` is the annotation orchestration pass.
- `Drawing` is the public facade, but also owns internal annotation registry/state.
- `layout.py` already owns the solver abstraction: `Placeable`, `LayoutSolver`, `fit_box`.

ADR direction already supports this: deterministic generation, lint/repair, solver-backed placement, and compose-then-pack. The refactor should move toward that model instead of creating arbitrary smaller files.

## Target module boundaries
Move toward this shape:

```text
draftwright/
  __init__.py
  make_drawing.py        # compatibility facade / public re-exports during migration
  builder.py             # build_drawing, make_drawing orchestration
  analysis.py            # STEP import, feature analysis, Analysis construction
  sheet.py               # page sizes, choose_scale, ViewBlock, StripDepths, repack planning
[...]

PR #147, opened 2026-06-27
Refactor groundwork (#138): ADR 0005, helper 0.12.1, golden-output gate

pzfreo:
Groundwork for the #138 pipeline refactor. **No production code changes** — this is the architecture decision, the helper-version baseline, and the behaviour-equivalence safety net that every later #138 PR will lean on. Public engine behaviour is unchanged (`387 passed` before and after).

## What's here

**1. `chore(#139)` — set the `build123d-drafting-helpers` floor to 0.12.1**
This PR only **sets the dependency floor**; it does **not** complete the migration. Production code still imports the helper's private symbols (`_full_cyls`, `_spec_key`, `_TB_COL_FRACTIONS`) — swapping those for the public replacements (`full_cylinders`, `HoleSpec.from_hole`, `TitleBlock.cell_bbox()`/`drawn_by_cell_bbox()`) is #139's actual work, on a later PR. The floor bump is the prerequisite: I verified against both PyPI sdists that those public symbols **ship in 0.12.1 and are absent from 0.12.0**, so the previous `>=0.12.0` floor could resolve a release lacking them. Floor + lock bumped; suite green on 0.12.1.

**2. `docs(adr)` — ADR 0005: pipeline boundaries and state ownership**
Records *why* #138 is shaped the way it is, so it doesn't get re-derived:
- Reshape the package along the compiler-pipeline stages (`builder`/`analysis`/`sheet`/`projection`/`linting`/`repair`/`export`/`annotations/`); `layout.py` unchanged.
- Give the build state that currently lives on `Drawing`'s private dicts **explicit owners — three, not one** (annotation registry / lint-side coverage state / pipeline build context), so the refactor dissolves the state-bus rather than renaming it.
- Behaviour-equivalence gated by the golden harness (Step 0); compat aliases carry deletion deadlines; helper-version risk recorded.
- Cross-refs into ADR 0003 (pin state → `registry.py`) and ADR 0004 (sheet/projection anchors relocate; decision unchanged); indexed in CLAUDE.md with a note that the migration is unstarted.

**3. `test(#138)` — golden-output regression gate (ADR 0005 Step 0)**
The existing tests assert *local* properties; none pins the *whole* drawing, so a refactor that shifts a dimension or reorders a pass can pass them all. This gate snapshots a canonical per-part digest:
- **drawing** — per-view edge counts + bboxes, every annotation's type/label/bbox, lint summary. Built from the **public surface only**, so the oracle survives the Step-2 registry move without edits.
- **svg** — structural tag histogram + page size, version/metadata-agnostic, to guard the export path (Step 3).

Coords rounded to 1e-4 mm. Four cases: `cylinder`/`plate`/`stepped` (fast) + `CTC-01` (slow, a real part with holes). CTC-02 excluded — too heavy for a routine gate, and its overlap acceptance is already pinned by `test_e2e_standards`. Regenerate intended changes with `UPDATE_GOLDEN=1`.

## Verification
- Full suite green on 0.12.1: `387 passed, 1 skipped`.
- Golden gate round-trips: generate → compare passes deterministically (a digest-determinism test guards the gate itself).
- Marginal cost: ~20s fast tier / +44s CI slow tier.

## Decisions deferred
- **Gate lifecycle after #138** — left open (revisit when the refactor completes); framed here as the refactor's Step 0 only.
- This PR is **scaffolding + decisions**, not the refactor. Steps 1–8 (the actual module moves) follow as separate PRs, each held byte-identical against these goldens.


pzfreo:
I reviewed the PR and the current CI state. There are no existing PR comments or review threads, but I think this needs a few fixes before merge:

- **Blocking:** the golden-output gate is not platform-stable. Ubuntu CI is failing `tests/test_golden.py::test_golden[cylinder]`, `[plate]`, and `[stepped]` because the committed goldens include exact annotation/text bboxes rounded to `1e-4` mm. The same tests pass on macOS, so this looks like font/rendering bbox variance rather than a drawing regression. The gate should either avoid exact `label_bbox` coordinates in cross-platform snapshots, quantize text bboxes more coarsely, pin the rendering/font environment, or split geometry assertions from platform-sensitive label extents.

- **Blocking:** lint is red because `tests/test_golden.py` is not ruff-formatted. `ruff format --check` wants the `anns = [...]` comprehension reformatted.

- **ADR/test mismatch:** ADR 0005 says Step 0 renders the corpus including `NIST CTC-02` and snapshots SVG + DXF geometry. The actual test excludes `ctc02`, and although it exports DXF, the comparison ignores `_dxf`. Either the ADR should describe the implemented gate accurately, or the test should add the promised DXF digest / coverage.

- **Golden export digest is weak:** `digest_svg()` currently records page size plus a tag histogram. That would miss many export regressions, including path coordinate changes, styles/classes, labels, layers, and all DXF output. If this is meant to guard the export path during #138, it should hash normalized structural attributes or add a more meaningful entity digest.

- **Minor docs drift:** `CLAUDE.md` still says `build123d-drafting-helpers>=0.9.1`, while this PR moves the actual floor to `>=0.12.1`.

One scope note: the PR body says the helper bump “adopts the public helper APIs”, but production code still imports helper private APIs (`_full_cyls`, `_spec_key`, `_TB_COL_FRACTIONS`). I verified helper `0.12.1` exposes the public replacements, so this PR is probably best described as setting the dependency floor for that migration rather than completing it.

pzfreo:
Thanks — all six points are addressed in `0db9e5b`. Point by point:

1. **Platform-stable golden gate (blocking) — fixed.** Confirmed root cause from the Ubuntu diff: every mismatch was in `.drawing.annotations[].bbox` and was a 0.04–0.26 mm shift; views/lint/SVG matched across OS. That's font-metric variance in dimension *text* boxes, not a drawing regression. The digest no longer records `label_bbox` or glyph coordinates — text-bearing annotations now pin their **value** (label string); only glyph-free geometry annotations keep a bbox. So the gate is portable (runs on macOS *and* Linux, no pinned font env needed). Linux confirmation is on this run's CI.

2. **Lint (blocking) — fixed.** `tests/test_golden.py` is ruff-formatted; `ruff format --check` and `ruff check` are clean locally.

3. **ADR/test mismatch — fixed both ways.** ADR 0005 Step 0 now describes the *implemented* gate (CTC-01, not CTC-02 — excluded as too heavy, its overlap acceptance already pinned by `test_e2e_standards`; and the SVG **+ DXF** digests). And the DXF is no longer ignored — see #4.

4. **Weak export digest — strengthened, with one deliberate boundary.** SVG digest now keys element counts by `(tag, class)` for per-layer granularity (`part`/`hidden`/`dims`) instead of a bare tag histogram, and a new **DXF digest** pins entity counts by type and layer. I did **not** hash path coordinates: that would re-introduce exactly the cross-platform text variance from #1 (glyphs are coordinate paths). So the gate pins counts + geometry, never text extents — that's the portability boundary, called out in the test docstring and the ADR.

5. **CLAUDE.md drift — fixed.** Dependency line now `>=0.12.1` (was a stale `>=0.9.1`).

6. **Scope wording — fixed.** You're right: this PR only sets the floor; `_full_cyls`/`_spec_key`/`_TB_COL_FRACTIONS` are still imported in production. PR body and commit reworded to say "set the dependency floor for #139," not "adopt the APIs." The actual swaps are #139's later PR.

Honest caveat: I can't run Linux locally (no Docker here), so the portability fix is verified by logic (removed the sole field that differed) + this CI run, not a local Linux pass. Watching the Ubuntu jobs now.

Issue #153, opened 2026-06-27
Speed up the local test tier (the 'fast' tier is ~8 min)


pzfreo:
## Problem
The default ("fast", `-m 'not slow'`) tier takes **~8 min** (471–489 s) for 408 tests — too slow for local iteration. The slow tier (CTC builds) is already correctly gated at ~19 min and CI-owned; this is about the *default* tier.

Root cause: nearly every test does a **real OCC build**. There are **202 `build_drawing`/`make_drawing` calls** in the suite, and the same simple parts are rebuilt many times to assert different *read-only* properties:

```
38×  Box(60…)
11×  Box(30…)
 5×  Box(80…)         9×  _multi_hole_plate(…)
 4×  Cylinder(30…)    5×  _holed_plate(…)
```

Building once and sharing the result across the read-only assertions would remove dozens of redundant builds. `--cov` (on by default via addopts) adds further overhead.

## Approach
1. **Session/module-scoped shared-build fixtures for the hot parts.** Provide built `Drawing`s for the common fixtures (`Box(60,40,20)`, the holed/stepped/multi-hole helpers, …) and have the *read-only* assertion tests consume them instead of rebuilding. There's precedent already (`scope="module"`/`scope="class"` in `test_make_drawing.py`).
   - **Isolation caveat:** only tests that don't mutate the drawing (add/remove/pin/repair/clear) may share a fixture. Mutating tests keep building their own. Audit per test; a shared fixture must be treated as read-only (consider a guard or a defensive copy for the few that need a tweak).
2. **A `smoke` marker** for a curated ~30 s subset — a handful of representative end-to-end builds + the golden primitives + lint-clean checks — so local "did I break something obvious" is seconds, not minutes. Default local run = `smoke`; full fast tier runs in CI.
3. **Coverage off by default locally.** Drop `--cov…` from the default `addopts` (or move it to a CI-only invocation); coverage is a CI concern and adds per-test overhead.
4. **CI keeps running the full fast + slow tiers** — no coverage loss there.

## Acceptance
- A `smoke` selection runs in **≤ ~45 s** locally and covers the main build/annotate/export/lint paths + the golden primitives.
- The full fast tier wall-clock drops materially (target: well under half of today's ~8 min) via shared-build fixtures, with **no loss of assertions** and no cross-test state leakage.
- CI still runs fast + slow with coverage; no behavioral test removed.

## Notes
[...]

PR #152, opened 2026-06-27
#138 Steps 1–2: public helper APIs + AnnotationRegistry

pzfreo:
Closes #139. **Step 1 of the #138 refactor.** Replaces draftwright's remaining private imports from `build123d-drafting-helpers` with the stable public contracts. Pure internal swap — **no output change**.

## Swaps
| Private | Public | Evidence |
|---|---|---|
| `_full_cyls` | `full_cylinders` | `full_cylinders is _full_cyls` → `True` in 0.13.0 |
| `_spec_key(h)` | `HoleSpec.from_hole(h)` | `_spec_key`'s entire body is `return HoleSpec.from_hole(h)` |
| `_TB_COL_FRACTIONS` cell math | `TitleBlock.drawn_by_cell_bbox()` | identical rect for every `legal_owner=""` block (see below) |

No private `build123d_drafting` names remain in `src/` **or** `tests/` (the title-block link-rect regression test now derives its expectation from the public accessor too).

## Adversarial review
Swept the title-block link rect (old hardcoded formula vs new public bbox) across `tb_w` ∈ {120…250}, several author strings, and with/without a legal-owner row:
- **`legal_owner=""` (every block `_add_title_block` builds): identical**, all configs.
- The only divergences are with a legal-owner row present, where the **old formula was latently wrong** (assumed a 2-row block, `top = block_h/2 = 12`; the real cell is `8`). The public accessor is correct, so this swap also **hardens** an edge case draftwright doesn't currently hit.

## Verification
- Fast tier local: **408 passed, 1 skipped**.
- **Golden gate passes against the committed snapshots with no regeneration** — independent proof of zero behavior change (exactly what the gate is for).
- Slow tier → CI.

No dependency or CHANGELOG change (the public APIs shipped in 0.12.1; we're already on 0.13.0, and there's no user-visible behavior change).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CoG24RWpmNApX4cMiucS7v

PR #154, opened 2026-06-27
Faster tests: smoke tier, coverage→CI, shared-build fixtures (#153)

pzfreo:
Part 1 of #153. Two low-risk wins toward a faster local loop; the structural lever (shared-build fixtures) is a deliberate follow-up — see below.

## Why
The default ("fast") tier is **~8 min** (471 s, 408 tests) because nearly every test does a real OCC build. That's too slow for tight iteration.

## What
- **`smoke` marker** — `uv run pytest -m smoke` runs a curated, build-light subset in **~34 s**: the layout solver unit tests, the e2e build→annotate→export→lint→standards path on three primitives, and the golden build→digest path. Fast "did I break something obvious" confidence.
- **Coverage out of the default addopts** — it adds **~13%** locally (measured: 49.5 s vs 43.2 s on a build-heavy slice) and instruments every line. The CI fast-tier step now passes `--cov` explicitly, so **CI coverage is unchanged**; local runs stay lean. (No coverage upload step existed, so nothing downstream is affected.)
- CLAUDE.md documents the three tiers (smoke / fast / slow) and that coverage is CI-only.

## Deliberately deferred
The big lever — **session/module-scoped shared-build fixtures** for the hot repeated parts (`Box(60…)` is built 38×, etc.) — is *not* here. It heavily edits `tests/test_make_drawing.py`, which **#152 also touches**, so doing it now would create a painful rebase conflict. It lands as part 2 once #152 merges. (The `test_registry.py` tests from #152 will also join the smoke set then.)

## Verification
Per the local workflow (targeted local + CI for the full tier): smoke run **36 passed (~34 s)**, layout slice 32 passed, lint clean. The **full fast tier + slow tier run in CI** — this PR changes test *selection/config*, not test logic, so CI is the comprehensive check.

https://claude.ai/code/session_01CoG24RWpmNApX4cMiucS7v

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,9 @@ jobs:
         # No --timeout here: it would override the 300 s set in
         # [tool.pytest.ini_options]. The heaviest test (full specimen-sheet SVG
         # export) can exceed 120 s on the slower macOS runners.
-        run: uv run pytest tests/ -v
+        # Coverage runs in CI; it is kept out of the default addopts so local
+        # runs stay lean (see pyproject [tool.pytest.ini_options]).
+        run: uv run pytest tests/ -v --cov=src/draftwright --cov-report=term-missing --cov-report=xml
 
   test-slow:
     # Heavy end-to-end CTC fixture builds (minutes each); run once, not across
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,8 +97,18 @@ Current ADRs:
 
 ## Testing
 
-`uv run pytest`. Tests are geometry-level — edge counts, bbox placement, face counts,
-lint clean checks. Target is 100% passing.
+Tests are geometry-level — edge counts, bbox placement, face counts, lint clean
+checks. Target is 100% passing. Tiers (#153):
+
+- **`uv run pytest -m smoke`** (~30 s) — curated build-light subset for a quick
+  local "did I break something obvious" check.
+- **`uv run pytest`** — full fast tier (`-m 'not slow'`, ~8 min; nearly every test
+  does a real OCC build). Prefer **targeted** selections (`-k`, node ids) locally.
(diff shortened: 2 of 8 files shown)


pzfreo:
## Added: part 2 — shared-build fixture + contamination guard

Rebased onto main (which now has #152), and added the structural piece.

`Box(60, 40, 20)` is built **32×** in `test_make_drawing.py`, mostly for read-only assertions. New **module-scoped `dwg_box_60_40_20`** fixture builds it once for those tests, with a **teardown contamination guard**: it snapshots annotation count / names / pins / per-view `(vis, hid)` identities and asserts they're unchanged at module teardown — so a consumer that accidentally mutates the shared drawing fails *loudly* instead of silently corrupting neighbours.

**Why the guard matters (a real finding):** an AST pass that classified tests by mutation calls got `TestViewBounds` **wrong both ways** — it missed that `test_view_bounds_includes_hidden_lines` reassigns `dwg.views["front"]` (a mutation) and false-flagged `.at()` (a read). So conversions are done **by hand, verified read-only only**, and the guard backstops the rest. I confirmed a mutating consumer trips the guard (`AssertionError: ...mutated dwg_box_60_40_20`).

Converted the verified read-only `Box(60,40,20)` tests in `TestViewBounds` (4) and `TestLintSuggestions` (2); the mutating siblings keep their own build. `test_registry.py` joins the `smoke` set.

**Honest scope:** this lands the *mechanism* + a verified conversion subset. The remaining ~8 RO conversions are mechanical follow-up (the guard makes them safe). The full-tier gain is bounded — the tier is build-bound and CI-owned — so the headline local win remains the **`smoke` tier (~40 s)** from part 1. I stopped at a verified subset rather than grinding all conversions blindly, given the classifier-unreliability finding.

Smoke: **45 passed (~40 s)**. Converted classes pass with the guard quiet.

Issue #135, opened 2026-06-26
Dimension non-cylindrical machined features (slots, flats, pockets) + a slot-completeness metric


pzfreo:
## Gap

draftwright dimensions holes (size + now location #93/#133), diameters (#77/#131),
and overall envelope — but **not non-cylindrical machined features**: slots,
machined flats, pockets, keyways. They appear in the views but carry no width /
length / depth / position dims, so the part isn't manufacturable.

Concrete: the `pzfreo/gramel` **shaft** has a **blade slot** (a rectangular
through-cut) and a **machined flat** — both completely undimensioned. The shaft
nonetheless scores 1.00 lint and (after #133) 1.00 location-coverage, because the
honest metric only inspects *holes*. So both the engine and the metric are blind
to this class.

## Two pieces of work

1. **Engine** — recognise and dimension non-cylindrical features:
   - slots / elongated cutouts: width, length, end radii, position;
   - machined flats: across-flat dimension and axial extent;
   - rectangular pockets: width, length, depth, position.
   Feature recognition likely needs a helpers-side primitive (build123d-drafting
   feature analysis) — scope that first.
2. **Metric** — extend `drawing-service/eval/score.py` with a feature-completeness
   axis beyond holes: of the machined features recognised, what fraction have a
   defining size dim AND a location dim. Until the engine can dimension slots,
   this axis will (correctly) drop the gramel shaft below 1.0 again.

## Priority
After the hole work (#93/#131/#133). This is the next manufacturability axis; it
also needs the metric so progress is measurable, same loop as before.

pzfreo:
Closing — the manufacturability gap this was filed for is fixed.

**Engine (piece 1): done.** The #148 machined-feature epic recognises AND dimensions the non-cylindrical features #135 called out — slots (#604 cross-slots, #607 arc/curved-stock), machined flats (#605), pockets (#603), plus turned/circlip grooves (#606). A slotted + flatted shaft (the gramel case) now carries size/location callouts instead of scoring a blind 1.00.

Two residuals carved out so nothing is lost:
- **#613** — radiused-end (obround) slot reports flat-wall length, not overall (end-radii/length gap).
- **#616** — piece 2's *metric*: the machined-feature **dimensioning-completeness** axis in the `drawing-service/eval/score.py` eval harness. Note #608 (`draftwright/score.py`) delivered a recognition *census*, which is a different measurement from the dimensioning-coverage axis #135 specified.

PR #185, opened 2026-06-28
feat: vendor feature recognition into draftwright (ADR 0007 PR-A)

pzfreo:
First migration step of **ADR 0007**: draftwright owns feature recognition;
`build123d-drafting-helpers` becomes the rendering library.

## What

Vendor `build123d_drafting.features` into a new **`recognition/`** subpackage and
switch every consumer to it. Deprecate-and-vendor (strangler fig): the upstream
copy stays in place — it'll be frozen/deprecated in a separate helpers release
(ADR 0007 PR-D) — and this becomes the source of truth.

- **`recognition/_features.py`** — 1:1 vendored copy (provenance note in the
  docstring) of the hole/boss/cylinder/pattern recognisers
  (`find_holes`/`find_bosses`/`analyse_cylinders`/`feature_diameters`/
  `find_hole_patterns`/`full_cylinders` + the feature/pattern types).
- **`recognition/slots.py`** — the existing draftwright-local slot recogniser
  (#135), `git mv`d in so `recognition/` is the single recognition home. Its
  docstring's "upstream this someday" note is reversed (ADR 0007).
- **`recognition/__init__.py`** — the public surface; all 8 consumers import here.
- **`tests/test_recognition.py`** — helpers' `test_features.py`, vendored with
  imports rewired to the package, so recognition is iterable here in isolation.
- **CLAUDE.md** — DAG + dependency framing updated per ADR 0007.

## Why

Recognition is *reasoning*, not *rendering*. Keeping it upstream fixed its
semantics by the wrong owner (e.g. `find_bosses`' chamfer-shortened length, wrong
for axial dims) and forced a cross-repo release for every tweak. This unblocks
PR-C (turned-part step lengths) to iterate recognition in one repo.

## Verification

Pure move, no logic change. Full fast tier **491 passed, 1 skipped**; the
vendored recognition tests **70 passed**; `ruff check`, `ruff format --check`,
and `mypy src/draftwright/` all clean.

## Not in scope

- helpers-side deprecation warnings (separate repo — ADR 0007 PR-D).
- Vendoring the lint engine (ADR 0007 PR-B).
- The turned-part step-length feature (ADR 0007 PR-C).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,16 +10,18 @@ produces a fully-annotated multi-view technical drawing (orthographic views, dim
 section A–A, ISO hatching, title block) ready for DXF/SVG export.
 
 It sits on top of two Apache 2.0 libraries:
-- `build123d-drafting-helpers` — annotation primitives (`Dimension`, `Leader`, `HoleCallout`, …)
+- `build123d-drafting-helpers` — annotation primitives (`Dimension`, `Leader`, `HoleCallout`, …).
+  The *rendering* library; draftwright owns feature recognition and linting (ADR 0007).
 - `build123d` — the underlying CAD kernel
 
 ## Architecture
 
 The dependency graph is a DAG (the #138 / ADR 0005 split is complete). Bottom to
-top: leaf modules (`layout.py`, `registry.py`, `linting.py`, `fonts.py`) → `_core.py`
-→ stage modules (`export.py`, `repair.py`, `projection.py`, `sheet.py`, `analysis.py`,
-`drawing.py`, the `annotations/` subpackage) → `builder.py` → the `make_drawing.py` /
-`annotate.py` compat facades. No lower module imports an upper one.
+top: leaf modules (`layout.py`, `registry.py`, `linting.py`, `fonts.py`, the
+`recognition/` subpackage) → `_core.py` → stage modules (`export.py`, `repair.py`,
+`projection.py`, `sheet.py`, `analysis.py`, `drawing.py`, the `annotations/`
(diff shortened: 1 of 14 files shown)


pzfreo:
Review finding:

- `src/draftwright/features.py` is removed by the rename to `src/draftwright/recognition/slots.py`, so existing downstream code using `from draftwright.features import find_slots` or `Slot` now fails with `ModuleNotFoundError`. The old module was non-private and the test suite used that import path before this PR, so this is a compatibility break rather than just an internal move. Please keep a small compatibility facade at `draftwright/features.py` that re-exports `Slot` and `find_slots` from `draftwright.recognition`, optionally with a deprecation warning pointing callers at the new package.

I checked the rest of the migration and did not find another issue. CI is green. Local spot checks:

- `uv run pytest tests/test_recognition.py -q` -> 70 passed
- `uv run pytest tests/test_make_drawing.py::TestFindSlots -q` -> 13 passed
- compared vendored `_features.py` against installed `build123d_drafting.features`; only the docstring/import example changed.

pzfreo:
Good catch — fixed in bf671a7.

Restored `draftwright/features.py` as a thin compat facade re-exporting `Slot`/`find_slots` from `draftwright.recognition`, with a `DeprecationWarning` pointing callers at the new package (consistent with the deprecate-and-vendor approach in ADR 0007 and the existing `make_drawing.py`/`annotate.py` facades). You were right that I was inconsistent — deprecate-and-vendor for the helpers move but a hard rename internally.

Verified: `from draftwright.features import find_slots, Slot` works and emits the deprecation warning; `TestFindSlots` 13 passed; ruff/format/mypy clean.

PR #187, opened 2026-06-28
fix: address adversarial review of the lint vendoring (ADR 0007)

pzfreo:
Post-merge adversarial review of **PR-B** (#186, lint vendoring) turned up two
issues. The review otherwise confirmed the vendoring is sound.

## Fixes

1. **Test exercised the wrong `lint_drawing`.** `test_repetition_label_passes_measured_check`
   imported `lint_drawing` from `build123d_drafting` (package root, no `.helpers`),
   which the earlier import switch missed — so it still tested the **upstream** lint,
   not the vendored one. Now imports from `draftwright.linting`.
2. **Magic number.** `Drawing.lint()` hardcoded `10` for the page margin in
   `page_bbox`; `_core` already defines `_MARGIN = 10.0`. Use the constant
   (numerically identical, no behaviour change).

## Review findings (no action needed — recorded for the trail)

- **Behavioural equivalence:** vendored `lint_drawing` produces identical issues
  to upstream across normal / overlap / out-of-bounds cases.
- **Source diff:** 13/16 vendored functions byte-identical; the 3 changes are the
  intentional `LintIssue.suggestion` field and two `bool()` wraps that are no-ops
  on already-boolean expressions.
- **Severed global is safe:** `_DRAWING_PAGE` is read *only* by `lint_drawing` in
  helpers — nothing else (export/render) depends on it.
- **`LintIssue.suggestion`:** all constructions use keywords (no positional shift),
  nothing compares it by equality/hash, serialization unchanged.
- **End-to-end:** goldens incl. the slow `ctc01` NIST part are stable.

## Verification

Corrected test passes against the vendored module; build+lint clean; `ruff`,
`ruff format --check`, `mypy` all green.

### DIFF
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -34,6 +34,7 @@ from build123d_drafting.helpers import (
 )
 
 from draftwright._core import (
+    _MARGIN,
     _STRIP_GAP,
     _STRIP_SPACING,
     Analysis,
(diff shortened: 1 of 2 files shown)


pzfreo:
Adversarial review result: no blocking findings.

I checked the two-file diff closely:

- `Drawing.lint()` now uses `_MARGIN` for `page_bbox`; this is behavior-equivalent to the previous hardcoded `10` because `_MARGIN == 10.0`, and `drawing.py` already depends on `_core`.
- `test_repetition_label_passes_measured_check` now imports `lint_drawing` from `draftwright.linting`; a local probe confirms that resolves to `draftwright.linting.structural`, not the upstream helper.
- The patch does not change the structural lint implementation itself, only the caller constant and the test import.

Local checks:

- `uv run pytest tests/test_make_drawing.py::TestHolePatternAnnotations::test_repetition_label_passes_measured_check -q` -> passed
- `uv run ruff check src/draftwright/drawing.py tests/test_make_drawing.py` -> passed
- `uv run ruff format --check src/draftwright/drawing.py tests/test_make_drawing.py` -> passed
- `git diff origin/main...HEAD --check` -> clean

CI fast matrix is green. Only residual note: PR CI skips `test-slow`, so this PR relies on the limited blast radius plus the targeted fast checks rather than a slow-suite rerun.

Issue #207, opened 2026-06-28
render: section/detail views via the new pipeline (future)


pzfreo:
Part of #195 (ADR 0008 **out-grow**) — **future**. Let the new pipeline trigger section/detail views from features that need them (blind/stepped bores) and route their dims accordingly. The engine keeps its current section pass; this adds the capability to the new path when a shape it owns needs a section. Likely follows the turned-via-new-pipeline work (#223).

pzfreo:
Done (PR #271). The section A–A **trigger** + cut-plane row are now decided by the planner (`plan_sections(model, feature_keys)` → `SectionPlan`) from the IR's Z hole/pattern features with cbore/spotface/blind bores; `_add_section_view` is the shared rendering it feeds (Amendment 4 — planner triggers, infra renders). Behaviour-equivalent; visual confirmed. The detail view stays user/lint-triggered (not feature-driven), so the feature-trigger scope of #207 is complete.

Issue #201, opened 2026-06-28
Phase 3: planner in production — holes (callouts + location dims)


pzfreo:
Part of #195 — ADR 0008 migration ([roadmap](../blob/main/docs/plans/0008-compiler-migration-roadmap.md)).

**Phase 3.**

Route hole callouts + location dims through `plan_dimensions` → existing layout; retire the hole placement in `annotations/holes.py`. First production use of the planner→layout seam; holes are axis-agnostic so parity is natural.

**Acceptance:** golden gate green (or intentional diff reviewed with rationale); dedicated X/Z parity test; no axis-gated back-end branch; existing suites + ruff/format/mypy green. **One PR.**

pzfreo:
**Hard prerequisite — planner view/routing must be model-aware (from #211 review).**

The prototype planner chooses a hole group's view from `feature.frame.axis` alone (`z→plan` etc.). That is too coarse for turned parts. Verified divergence on `_z_stepped_bored()`:

```
engine:  ø8 concentric axial bore → ldr_z0 on FRONT (turned-bore leader; depth via section A–A)
model:   hole group               → 'plan' (generic end-on), bore diameter + depth
```

For a **concentric axial bore on a turned part** the engine routes **diameter → front bore leader** and **depth → section view** (depth is not a plain end-on number), and excludes it from generic per-hole plan callouts. For a **prismatic blind hole** it's **ø + depth as one end-on callout**. So placement is **part-context-dependent** (turned vs prismatic; axial vs radial), not a function of axis alone.

When wiring holes here, the planner's view/routing must take the full `model` + feature + params:
- concentric axial bores on a turned model → the **turned-bore path** (front `ldr_z`), not the end-on rule;
- depth/cbore/spotface placement must be longitudinal/section-aware where the convention requires it (interplay with #207 section triggers);
- generic (prismatic, radial, simple) holes → the end-on callout.

Gate-protected: `bored_stepped_z`/`bored_stepped_x` in the migration corpus enforce this once holes are wired. Add a regression on `_z_stepped_bored()` asserting the blind axial bore is NOT placed as a simple end-on `plan` callout.

pzfreo:
Superseded by the ADR 0008 strategy pivot (out-grow, not reproduce-and-swap — #214). Disposition of its points:
- **Hole callouts** through plan→layout: **done** (seam #213, slice #215/#216).
- **Location dims**: carried to #220 (render completeness).
[...]

PR #211, opened 2026-06-28
feat: revise IR/planner contract (role + grouping + feature-aware dedup) + counterbore (ADR 0008)

### LINKED ISSUE #197: Phase 2: detector — counterbore/spotface as DimParameters on HoleFeature
Part of #195 — ADR 0008 migration ([roadmap](../blob/main/docs/plans/0008-compiler-migration-roadmap.md)).

**Phase 2.**

Enrich `HoleFeature` so cbore/spotface steps surface as `DimParameter`s; `build_part_model` emits them. No engine behaviour change (parallel model).

**Acceptance:** golden gate green (or intentional diff reviewed with rationale); dedicated X/Z parity test; no axis-gated back-end branch; existing suites + ruff/format/mypy green. **One PR.**

pzfreo:
Closes #197. Reworked after an **adversarial review** of the counterbore work,
which falsified the prototype's `DimParameter`/planner contract. Fixes the
foundation before building on it. Still prototype — **not wired into
`build_drawing`**, so the migration gate is unchanged (18 passed).

## Review findings (all fixed here)
1. **Dedup collapsed distinct same-value diameters** — a `counterbore ø16` + a
   `boss ø16` → the boss was silently dropped.
2. **Two identical counterbored holes → one callout** — value-dedup ate the count.
3. **Flat stream lost per-hole grouping** — bore/cbore/depth became independent
   dims with no owner, so a compound callout couldn't be reassembled.
4. **Baked GD&T glyph labels** (`⌴`/`⌵`/`↧`) **aren't in the pinned font** (verified
   with fontTools) — and the engine draws those symbols as *geometry*, not text.

## Contract revision
- `DimParameter` gains a semantic **`role`** (bore/counterbore/spotface/step/boss/…)
  and **drops the baked label** — formatting/symbols are a renderer concern;
  `display()` gives font-safe debug text.
- The planner returns **`DimensionGroup`s** (one per feature), so a hole's
  bore+counterbore+depth stay **one compound callout**.
- **Redundancy is feature-aware** — params that merely share a value but differ in
  `role` both survive; only genuine duplicates within a feature dedup. Count
  aggregation (`3× ø8`) is upstream (pattern detection), not here.
- Counterbore (#197) reworked onto the new contract.

## Verification
Prototype suite **8 passed** — incl. new tests for compound grouping,
feature-aware redundancy, and font-safe labels. Migration gate **18 passed**
(model-only). ruff/format/mypy clean.

### DIFF
--- a/src/draftwright/model/__init__.py
+++ b/src/draftwright/model/__init__.py
@@ -28,13 +28,15 @@ from draftwright.model.ir import (
     HoleFeature,
     PartModel,
     StepFeature,
+    display,
 )
-from draftwright.model.planner import PlannedDimension, plan_dimensions
+from draftwright.model.planner import DimensionGroup, PlannedDimension, plan_dimensions
 
 __all__ = [
     "BossFeature",
     "Datum",
     "DimParameter",
+    "DimensionGroup",
     "Feature",
     "Frame",
     "HoleFeature",
@@ -42,5 +44,6 @@ __all__ = [
     "PlannedDimension",
     "StepFeature",
     "build_part_model",
+    "display",
     "plan_dimensions",
 ]
--- a/src/draftwright/model/detect.py
+++ b/src/draftwright/model/detect.py
@@ -49,7 +49,7 @@ def build_part_model(part) -> PartModel:
     bbox = part.bounding_box()
     features: list[Feature] = []
 
-    # Holes — any orientation.
+    # Holes — any orientation; counterbore/spotface steps come along as params.
     for h in find_holes(part):
         features.append(
             HoleFeature(
@@ -57,6 +57,8 @@ def build_part_model(part) -> PartModel:
                 diameter=h.diameter,
                 depth=h.depth,
                 through=(h.bottom == "through"),
+                cbore=(h.cbore.diameter, h.cbore.depth) if h.cbore else None,
+                spotface=(h.spotface.diameter, h.spotface.depth) if h.spotface else None,
             )
         )
 
--- a/src/draftwright/model/ir.py
+++ b/src/draftwright/model/ir.py
@@ -107,7 +107,7 @@ class HoleFeature:
     kind: ClassVar[str] = "hole"
 
     def parameters(self) -> list[DimParameter]:
-        loc = (self.frame.origin, self.frame.origin)
+        # Location is the group's anchor (Feature.frame), not a parameter.
         ps = [DimParameter("diameter", "bore", self.diameter)]
         if not self.through and self.depth is not None:
             ps.append(DimParameter("depth", "bore", self.depth))
@@ -119,7 +119,6 @@ class HoleFeature:
             sd, sdp = self.spotface
             ps.append(DimParameter("diameter", "spotface", sd))
             ps.append(DimParameter("depth", "spotface", sdp))
-        ps.append(DimParameter("location", "location", 0.0, span=loc))
         return ps
 
     def references(self) -> list[Datum]:
--- a/src/draftwright/model/ir.py
+++ b/src/draftwright/model/ir.py
@@ -73,13 +73,18 @@ class Feature(Protocol):
 
 @dataclass(frozen=True)
 class HoleFeature:
-    """A drilled hole — bore diameter, depth (blind), and location from datums."""
+    """A drilled hole — bore diameter, depth (blind), optional counterbore /
+    spotface steps, and location from datums. ``cbore``/``spotface`` are
+    ``(diameter, depth)`` or ``None`` (kept as plain tuples so the IR doesn't
+    depend on the recogniser's types)."""
 
     frame: Frame
     diameter: float
     depth: float | None
     through: bool
     count: int = 1
+    cbore: tuple[float, float] | None = None
+    spotface: tuple[float, float] | None = None
     kind: ClassVar[str] = "hole"
 
     def parameters(self) -> list[DimParameter]:
@@ -87,6 +92,14 @@ class HoleFeature:
         ps = [DimParameter("diameter", self.diameter, f"{n}ø{_fmt(self.diameter)}")]
         if not self.through and self.depth is not None:
             ps.append(DimParameter("depth", self.depth, f"↧{_fmt(self.depth)}"))
+        if self.cbore is not None:
+            cd, cdp = self.cbore
+            ps.append(DimParameter("diameter", cd, f"⌴ø{_fmt(cd)}"))
+            ps.append(DimParameter("depth", cdp, f"⌴↧{_fmt(cdp)}"))
+        if self.spotface is not None:
+            sd, sdp = self.spotface
+            ps.append(DimParameter("diameter", sd, f"⌵ø{_fmt(sd)}"))
+            ps.append(DimParameter("depth", sdp, f"⌵↧{_fmt(sdp)}"))
         ps.append(
             DimParameter("location", 0.0, "loc", span=(self.frame.origin, self.frame.origin))
         )
--- a/src/draftwright/model/ir.py
+++ b/src/draftwright/model/ir.py
@@ -3,16 +3,19 @@
 The narrow waist between recognition and dimensioning. Two protocols carry the
 weight:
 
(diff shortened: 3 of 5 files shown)


pzfreo:
Adversarial review findings:

1. `src/draftwright/model/planner.py:71` drops legitimate same-valued dimensions within one feature. The dedup key is only `(kind, role, round(value, 2))`, so two distinct spans like a 10x10 square pocket collapse to one dimension. I reproduced this with a custom feature emitting two orthogonal `length/pocket/10.0` params; `plan_dimensions()` returned one dim, not two. The key needs to include geometry identity such as `span`/`refs`, or the IR needs an explicit parameter identity.

2. `src/draftwright/model/planner.py:55` / `:79` makes compound hole groups internally inconsistent for rendering by assigning view per parameter kind. A counterbored hole group gets bore and counterbore diameters in `front`, but counterbore depth in `plan`, so the "one compound callout" group cannot actually be rendered as one callout. View selection should be feature/group-aware, or `DimensionGroup` should own the view.

Open question: the group still loses the feature anchor because `location` params are skipped and `DimensionGroup` only keeps `feature_kind` plus dims. That may be acceptable prototype scope, but it should be fixed before wiring this into drawing output.

Verification: `uv run pytest tests/test_part_model.py -q` passes locally (8 passed). CI is green for the fast matrix; `test-slow` is skipped.

pzfreo:
Good catches — all three fixed in $(git rev-parse --short HEAD).

1. **Pocket dedup** — dropped the planner's value-dedup entirely (keying on geometry would also work, but features own their parameters and shouldn't emit spurious duplicates; genuine repeated-feature count belongs upstream in pattern detection). A `PocketFeature` emitting two orthogonal 10 mm lengths now yields two dims (regression test added).
2. **Split compound group** — `view` moved onto `DimensionGroup` (one view per group, feature-aware via `_group_view`), so a counterbored hole's bore/cbore/depth all land in one view → one callout.
3. **Missing anchor** — `DimensionGroup` now carries `anchor` (`Feature.frame.origin`); the `location` stub param is removed, so a group is placeable.

Verified: prototype 9 passed (incl. one-group/one-view/with-anchor for compound holes); migration gate unaffected (model-only); ruff/format/mypy clean.

pzfreo:
Hard re-review of my own planner turned up an **X/Z asymmetry** — fixed in $(git rev-parse --short HEAD).

`_group_view` accepted `orientation` but ignored it and keyed only on `feature_kind`, so **every hole's callout was forced to `plan`** regardless of axis. Reproduced: a Z-axis bore and the same part turned to X **both** returned `plan`, when the X hole is seen end-on in the **side** view. That's precisely the orientation-as-branches bug the IR is meant to eliminate.

Fix: a diameter callout's view is now derived from the feature's axis by one rule for all axes — `z→plan`, `x→side`, `y→front`; a turned step's length+OD stay on the lengthwise `front` view. Regression test asserts Z-hole→`plan`, X-hole→`side`.

10 passed; gate unaffected (model-only); ruff/format/mypy clean.

pzfreo:
Re-review finding:

The previous dedup, split-group view, and missing-anchor issues are fixed, but the new group-level view selection is still too coarse for non-Z holes. `src/draftwright/model/planner.py:57-60` only looks at `feature_kind`, and `plan_dimensions()` passes only `feature.kind` at `src/draftwright/model/planner.py:77`, so every `hole` group is assigned `view="plan"` regardless of `Feature.frame.axis`.

This is a real model path, not just synthetic: the existing cross-axis fixture `Box(60, 60, 20) - Cylinder(4, 60, rotation=(0, 90, 0))` builds a `HoleFeature` with `frame.axis == "x"`, but the updated planner returns a hole group with `view == "plan"`. The current annotation layer maps hole axis to view as `z -> plan`, `y -> front`, `x -> side`; an x-axis hole is circular in the side view, not the plan view. As written, the new contract would place side-drilled hole callouts on the wrong view once this IR is wired into drawing output.

Suggested fix: make `_group_view` accept the feature, not just `feature_kind`, and use `feature.frame.axis` for holes, e.g. `{"z": "plan", "y": "front", "x": "side"}`. Then add a regression test with the existing cross-axis-hole geometry or an explicit `HoleFeature(Frame(..., "x"), ...)`.

pzfreo:
This crossed with my previous commit — we found the same bug within a minute, and **`229e71d` already implements your exact suggested fix**: `_group_view` now takes the feature and uses `feature.frame.axis` with `{"z": "plan", "y": "front", "x": "side"}`; a turned step's length+OD stay on `front`.

Verified against your exact fixture:
```
Box(60, 60, 20) - Cylinder(4, 60, rotation=(0, 90, 0))
  -> HoleFeature frame.axis == "x"
  -> hole group view == "side"   ✓ (was "plan")
```
Regression test `test_group_view_follows_axis_not_hardcoded` covers Z→plan and X→side. So this finding is resolved on HEAD — please re-run against `229e71d`.

pzfreo:
Detailed re-review on current HEAD `229e71d`:

The earlier findings are resolved: value-blind dedup is gone, groups now carry one `view`, groups have an `anchor`, and the X/Y/Z axis mapping fix is present. CI is green for the fast matrix, and `uv run pytest tests/test_part_model.py -q` passes locally (10 passed).

I still see one contract-level blocker before this IR/planner can be safely wired into drawing output:

`src/draftwright/model/planner.py:61-69` now maps every non-step feature to the view where `feature.frame.axis` is seen end-on (`z -> plan`, `x -> side`, `y -> front`). That is correct for a simple through-hole diameter callout, but it is too coarse for the grouped hole contract because `HoleFeature.parameters()` can include axial depth/counterbore/spotface parameters (`src/draftwright/model/ir.py:111-121`). Once a depth parameter is in the same group, the end-on view is often the wrong single view for the compound callout.

Concrete repro using the PR's own turned/boring fixture shape:

```py
shaft = Cylinder(15, 30) + Pos(0, 0, 30) * Cylinder(8, 30)
part = shaft - Pos(0, 0, 45) * Cylinder(4, 20)
model = build_part_model(part)
# model.orientation == "z"
# HoleFeature: frame.axis == "z", through == False, depth == 10.0
plan_dimensions(model)
# -> DimensionGroup(feature_kind="hole", view="plan",
#      dims=[("diameter", "bore", 8.0), ("depth", "bore", 10.0)])
```

That group says the blind bore's diameter and depth are one placeable callout in `plan`, but the depth is along Z and is not visible in the end-on plan view. This also conflicts with the existing drawing behavior for turned parts: concentric axial bores are part of the turned base set, get front-view bore leaders, and are explicitly excluded from generic per-hole plan callouts to avoid duplicates/wrong placement (`annotations/orchestrator.py` around the `ldr_z` and `feature_holes` logic).

Suggested fix: make group view selection use the full `model` plus the feature and its planned parameters, not only `feature.frame.axis`. At minimum, hole groups with depth/counterbore/spotface need a longitudinal/section-aware placement, and concentric axial bores on a turned model need to follow the turned-bore path rather than the generic end-on hole rule. Add a regression using the `_z_stepped_bored()` shape or an explicit `PartModel(orientation="z", features=[HoleFeature(Frame(..., "z"), through=False, depth=...)])` asserting the blind axial bore group is not placed in `plan` as a simple end-on callout.

pzfreo:
Verified and agreed — thanks, this is a real one. On the PR's own fixture the current engine routes the concentric axial bore to a **front** `ldr_z` leader (depth via the section), while the model puts it in `plan`:

```
engine: ø8 bore → ldr_z0 on 'front'      model: hole group → 'plan' (+ depth)
```

It's a **planner-maturity / wiring** concern, not a #211 contract-structure bug: the per-feature group + single-view + anchor contract is fine, but the *view/routing selection* must become model-aware (turned concentric bore → turned-bore path; depth/cbore section-aware; generic holes → end-on). That's exactly Phase 3's job and it's gate-protected (`bored_stepped_z`/`_x` in the corpus fail the gate the instant holes are wired).

Recorded as a hard prerequisite on #201 (with the #207 section interplay). #211 stays scoped to the contract + counterbore; not expanding it again.

Issue #198, opened 2026-06-28
Phase 2: detector — hole patterns → PatternFeature


pzfreo:
Part of #195 — ADR 0008 migration ([roadmap](../blob/main/docs/plans/0008-compiler-migration-roadmap.md)).

**Phase 2.**

Adapt `find_hole_patterns` (bolt-circle / linear / grid) into a `PatternFeature` emitting pitch/grid/count `DimParameter`s. Parallel model only.

**Acceptance:** golden gate green (or intentional diff reviewed with rationale); dedicated X/Z parity test; no axis-gated back-end branch; existing suites + ruff/format/mypy green. **One PR.**

PR #213, opened 2026-06-28
feat: renderer seam — DimensionGroup → placed HoleCallout (ADR 0008, #201 start)

pzfreo:
Builds the **renderer seam** — the first real consumer of the planner output —
validating the IR/planner contract against a real consumer instead of review
guesswork. Part of #201 (planner in production for holes).

## What
`model/render.py` reads a hole/pattern `DimensionGroup` **purely from its planned
parameters** (+ the feature's `count`) and builds a placed `HoleCallout` leader via
the real projection (`Drawing.at`) and helpers:
- `hole_callout_spec(group)` — the renderer's reading of the plan (bore/counterbore
  values by role, through inferred from absent depth param, count from the feature).
- `render_callouts(dwg, groups)` — placed `HoleCallout` leaders, projected into each
  group's view.

GD&T symbols (⌴/↧) are the helper's geometry — which is exactly why the IR carries
semantic `role`s, not glyph strings. This is the payoff of that contract decision.

## Why this, now
The last 5 review rounds found contract gaps by imagining the renderer's needs. A
real consumer settles them: this proves the plan carries what's needed (bore +
counterbore + blind/through + **pattern count → one `6× ø6` callout, not six**), and
any remaining gap now surfaces against real code, not speculation.

## Scope
- **Does not yet replace** the engine's hole placement — kept out of
  `model.__init__` (pure IR stays helpers-free) and not wired into `build_drawing`.
  The swap-in under the migration gate (turned-bore routing, etc.) is the rest of
  #201.
- Gate unaffected (engine untouched).

## Verification
`tests/test_render_seam.py`: spec-level (through / blind / counterbore / count) +
render (placed leaders; bolt circle → 1 leader). 6 passed; ruff/format/mypy clean.

### DIFF
--- a/src/draftwright/model/render.py
+++ b/src/draftwright/model/render.py
@@ -1,0 +1,80 @@
+"""render — the renderer seam: DimensionGroup → placed annotations (ADR 0008).
+
+The **first real consumer** of the planner output, validating the IR/planner
+contract end-to-end. A hole/pattern group is read *purely from its planned
+parameters* (+ the feature's `count`) and turned into a placed `HoleCallout`
+leader via the existing projection (`Drawing.at`) and rendering primitives. GD&T
+symbols (⌴/↧) are the helper's geometry, which is exactly why the IR carries
+semantic `role`s, not glyph strings.
+
+This is the planner→layout seam. It does **not** yet replace the engine's hole
+placement — that swap-in, under the migration gate, is #201. Here it proves the
+contract carries what a renderer needs, and surfaces gaps before wiring. Kept out
+of `draftwright.model.__init__` so the pure IR stays free of any helpers import.
+"""
+
+from __future__ import annotations
+
+from build123d_drafting.helpers import HoleCallout, Leader
+
+from draftwright.model.planner import DimensionGroup
+
+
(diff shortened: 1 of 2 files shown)


pzfreo:
Self-review of the seam against the engine's actual hole-callout convention found three gaps — fixed in $(git rev-parse --short HEAD):

1. **Spotface dropped** — `hole_callout_spec` read only the counterbore role. Now `step = counterbore or spotface` (counterbore precedence), mirroring `_build_callout` in `annotations/holes.py`.
2. **Pattern suffix missing** — the engine puts the BCD/grid in the HoleCallout *suffix* (`EQ SP ON ø50 BC`, `(3×3)`), not a separate dim. The seam now reproduces that from the feature's `bcd`/`rows`/`cols`.
3. **Pattern leader tipped at the empty centre** — a bolt-circle leader pointed at `anchor` (the axis centre, no hole there). It now tips at a **member hole** (`members[0]`); a single hole still tips at itself.

These validate the contract carries what the engine's convention needs (and that the #198 arrangement-facts were worth keeping — the suffix + member tip use `bcd`/`rows`/`cols`/`members`). 7 tests pass; gate unaffected (seam is parallel, not wired); ruff/format/mypy clean.

Issue #218, opened 2026-06-29
lint: location + centermark coverage (drawing-derived) — make lint a real completeness bar


pzfreo:
Part of #195 (ADR 0008 out-grow). From the end-to-end slice review, **Finding 1: lint-clean ≠ complete.** A plate with hole callouts but **no location dimensions and no center marks** lints `1.0`. Lint is draftwright's single correctness judge (ADR 0002 lint→repair, 0007 own-linting) — the fix is to **strengthen lint, not add a parallel checker** (pzfreo: "improve lint rather than a different check").

Add coverage codes, **derived from the drawing** (placed annotations vs detected features), so they judge *any* producer — engine and new pipeline alike:
- `feature_not_located` — a hole/feature with no locating dimension to a datum.
- `feature_no_centermark` — a hole/circular feature with no center mark.

This closes the long-standing "Location coverage remains out of scope" note (`src/draftwright/linting/coverage.py:136`, which pointed at the now-closed #93). It also makes the pivot's "judged by correctness = lint-clean" actually meaningful, and surfaces the *engine's* own location gaps where they exist.

Pairs with #219 (drive coverage off the drawing, not `CoverageState`) and #220 (render side).

Issue #227, opened 2026-06-29
lint: refine location coverage — per-axis precision + per-hole granularity


pzfreo:
Follow-up to #218 (location/centermark coverage). The first cut is deliberately coarse; track the refinements:

- **Per-axis location.** A hole counts as located if *any* dimension witness aligns in x **or** y, so a partially-located hole (x dimensioned, y not) isn't flagged. Errs toward under-warning (safe), but a complete check should require *both* in-plane coordinates located.
- **Per-hole granularity.** Today the check emits **one aggregated issue per code** ("N hole(s) …") to avoid flooding lint on many-hole parts. Consequence: the score penalty applies once, not ×N, and you can't tell *which* holes are uncovered. Consider per-hole (or per-distinct-location) issues, consistent with how `lint_feature_coverage` reports per diameter.
- **End-on-view assumption.** Coverage is checked only in the view normal to the hole axis. Holds for the engine and the model pipeline today; revisit if a producer marks/locates a hole in another view.

Not blocking — the coarse check already catches the real gaps (the new pipeline's missing dims, and engine #225).

PR #226, opened 2026-06-29
feat: lint location + centermark coverage (drawing-derived) — #218

pzfreo:
Closes #218. Acts on the slice review's **Finding 1: lint-clean ≠ complete** — a drawing with hole callouts but no location dims / center marks lints `1.0`. Fix is to **strengthen the single judge**, not add a parallel check.

## What
`lint_location_coverage(part, dwg)` derives coverage **from the drawing itself**, so it judges *any* producer (engine + model pipeline). For each hole, project its centre into the end-on view and check the placed annotations:
- **`feature_no_centermark`** — no `CenterMark` at the projected hole centre;
- **`feature_not_located`** — no `Dimension` witness aligned to the hole centre.

Exemptions (calibrated against engine output, no false positives):
- **patterned holes** — a bolt circle/array is located by its BCD/pitch, not per-hole dims;
- **coaxial holes** — a hole on the part centre/turning axis is located by centrelines.

Wired into `Drawing.lint()`. Closes the long-standing "location coverage out of scope" gap (`coverage.py:136`).

## Fallout from strengthening the judge — all real signal
Full fast suite **577 passed / 1 skipped**; exactly 3 tests touched:
- **2 `e2e_slice` tests reframed** — the new pipeline is error-free + size-complete, and the strengthened lint now correctly flags its missing center marks/location (the **#220** gap). That's the point.
- **1 engine test narrowed** — it surfaced a *real* engine gap: the second side-drilled hole is unlocated. Filed as **#225**; the test (about callout placement) filters that one warning with a pointer.

## Verification
577 passed / 1 skipped · ruff/format/mypy clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #228, opened 2026-06-29
feat: derive axial coverage from the drawing, not CoverageState — #219

### LINKED ISSUE #219: lint: derive coverage from the drawing, not the CoverageState side channel
Part of #195 (ADR 0008 out-grow). From the slice review, **Finding 4.** `lint_axial_coverage` (and parts of feature coverage) read `CoverageState` — a counter the **engine's** orchestrator sets as it places dims. So the new pipeline can't satisfy them no matter what it renders: a stepped shaft drawn via the new path always flags `axial_length_missing` because nothing set `cover_axial`.

Refactor `linting/coverage.py` to derive coverage from the **drawing itself** (placed dimensions/annotations vs detected features), removing the `CoverageState` side-channel dependency. Then lint judges any producer on its actual output.

Companion to #218 (the new location/centermark coverage must also be drawing-derived). Related: #97 (coverage recomputes find_holes each lint — fold the reuse in here).

pzfreo:
Closes #219. Companion to #218: lint should judge a drawing by its **output**, not a build-time side channel set only by the engine's passes.

## Problem
`lint_axial_coverage` read `CoverageState.axial_covered` — a counter the engine's step-length pass increments via `cover_axial`. A stepped shaft drawn through the **model pipeline** therefore *always* tripped `axial_length_missing`, no matter what it actually dimensioned, because nothing called `cover_axial`.

## Fix
`lint_axial_coverage(part, dwg)` now counts dimensioned step lengths **from the drawing**: `_axial_covered_from_drawing` marks a step covered when a front-view `Dimension` has witness vertices at both of its shoulders' page-x positions (same drawing-derived technique as #218's location check). It judges any producer.

Removed the now-dead axial side channel — `CoverageState.cover_axial` / `axial_covered`, and the `cover_axial` call in `annotations/turned.py`.

## Scope (what stays, and why)
`CoverageState` remains for two **genuinely engine-internal** signals that have *no drawing trace* to derive from, and which don't block the model pipeline:
- `dropped_diams` — per-view-cap suppression for `feature_not_dimensioned` (a *dropped* callout leaves no annotation, so it can't be read back from geometry);
- pattern coverage (`cover_pattern` …) — hole-table escalation coordination, not a lint judgment.

The axial channel was the only one producing false lint for the new pipeline.

## Verification
Drawing-derived count matches the old side channel exactly on engine output (2/2, 3/3 on stepped / three-step X shafts). The 3 axial tests now build real drawings (bare scaffold → flagged; engine chain → clean) instead of synthetic counts. Full fast suite **581 passed / 1 skipped**; ruff/format/mypy clean.

### DIFF
--- a/src/draftwright/annotations/turned.py
+++ b/src/draftwright/annotations/turned.py
@@ -276,4 +276,3 @@ def _annotate_turned_lengths(dwg, a: Analysis, prof: TurnedProfile | None) -> No
             f"dim_len{i}",
             view="front",
         )
-    dwg._coverage.cover_axial(len(steps))
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -982,7 +982,7 @@ class Drawing:
             )
             issues += lint_axial_coverage(
                 self.part,
-                self._coverage.axial_covered,
+                self,
                 assembly=self.assembly,
             )
             issues += lint_location_coverage(
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -168,6 +168,7 @@ _GEOMETRY_AWARE_CODES = frozenset(
         "feature_count_mismatch",
         "feature_not_located",
         "feature_no_centermark",
+        "axial_length_missing",
         "missing_principal_dimension",
         "label_vs_measured",
         "dim_inside_part",
--- a/src/draftwright/linting/coverage.py
+++ b/src/draftwright/linting/coverage.py
@@ -58,10 +58,6 @@ class CoverageState:
         # redundant feature_not_dimensioned for them. Reset at the top of
         # _auto_annotate so re-annotation does not accumulate.
         self._dropped_callout_diams: list = []
-        # Count of turned-part axial step lengths the step-length pass dimensioned,
-        # so lint_axial_coverage knows how many shoulders are located. Reset with
-        # the dropped diameters at the top of _auto_annotate.
-        self._axial_covered: int = 0
 
     # -- pattern coverage -----------------------------------------------------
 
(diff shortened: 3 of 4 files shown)


pzfreo:
## Adversarial self-review

**Verdict: sound, one fix pushed.** The drawing-derived count matches the old side channel exactly on engine output, and the side-channel removal is clean.

### Checks that passed
- **Clean removal** — no orphaned refs to `cover_axial` / `axial_covered` anywhere in src/tests/docs.
- **Repair-loop interaction** — coverage stays clean after build+repair and is idempotent across repeated `lint()`s. Repair moves the dim line/label but the witness *tips* stay anchored to geometry, so a covered step stays covered. Verified on x-2-step, three-step, chamfered, and bored shafts. Locked with a regression test.
- **Edge cases** — chamfered shoulders and the bored-shaft phantom floor don't break detection (`find_turned_steps` already excludes the internal floor; coverage rides on its shoulders).
- **Scope** — the two remaining `CoverageState` signals (`dropped_diams`, pattern coverage) genuinely have no drawing trace and don't block the new pipeline; leaving them is correct, not laziness.

### Fixed (pushed)
- **`axial_length_missing` wasn't in `_GEOMETRY_AWARE_CODES`** — I'd flagged this in the #226 review and said to fold it in when a PR reworks axial coverage. This is that PR; I missed it. Added it (+ a test asserting the classification), so `lint_summary.geometry_issues` counts it correctly.

### Limitation (documented, consistent with #218)
- Same `tol=0.6` page-mm witness match and **one aggregated issue** per part as the location check — under-warns rather than false-alarms. Tracked for refinement in #227.

Issue #288, opened 2026-06-30
cli/install: PDF as a core dep + default output, and a --format selector


pzfreo:
Three coupled changes to install + CLI output policy. **No engine change** — the `Drawing.export()`/`export_pdf()` library API stays flexible; this is about install weight and CLI defaults. Don't-code-yet; capturing scope + trade-offs first.

## 1. Make PDF a core install (drop the `[pdf]` extra)

Today `cairosvg` is an optional extra (`pyproject.toml:32-33`, `pip install draftwright[pdf]`). Move `cairosvg>=2.7` into core `dependencies` and delete the extra.

- **Trade-off:** `cairosvg` → `cairocffi` pulls in native **libcairo**. But build123d already drags the (much larger) native OCP/OpenCASCADE wheel, so cairo is *incremental* weight on an already-heavy package — the "keep it dependency-light" argument is largely already spent. Acceptable.
- Sweep the `export_pdf` docstring (`drawing.py:1146-1149`) that says \"install with `pip install draftwright[pdf]`\".

## 2. Output PDF by default; drop SVG + DXF from the default set

Today the CLI emits **SVG + DXF** by default and PDF only behind `--pdf`. Flip it: **PDF is the default output.**

- **Caveat — this is changing the *default set*, not removing the capability.** SVG and DXF stay fully supported and reachable via the `--format` selector (#3). This keeps the deliverable-first default without losing outputs.
- **Push-back worth a decision:** DXF is the *machine* deliverable — editable/measurable in CAD — arguably more important for an engineering-drawing tool than a flat PDF. Defaulting to PDF-only means a CAD user must now pass a flag to get DXF. Is PDF-only the right *default*, or should the default be e.g. `pdf,dxf` (human deliverable + CAD interchange)? **Recommend deciding this explicitly before implementing.**
- **Implementation note:** PDF is currently *rendered from the SVG* (`export_pdf` runs cairosvg over the SVG). So a PDF build still produces an SVG internally — when SVG isn't requested, render it to a temp and discard, rather than emitting it.

## 3. Add a `--format` output-type selector

Replace the boolean `--pdf` flag with a format selector.

- **Multi-valued, not single-choice:** engineering users routinely want several at once (PDF for review + DXF for CAD). Prefer a comma-list / repeatable flag — e.g. `--format pdf,dxf` — over `--format {one-of}`. `all` as a convenience alias.
- **Default:** whatever #2 settles on (`pdf`, or `pdf,dxf`).
[...]

Issue #276, opened 2026-06-29
cli: rich TUI over a structured pipeline-event stream (engaging per-stage progress + live lint readout)


pzfreo:
## Idea

The ADR 1 (was 0008) pipeline is now beautifully legible (`detect → IR → plan → render → layout → export`). The CLI should match it: a compact, engaging **rich** TUI that shows live progress *through the named pipeline stages*, the feature inventory as it's recognised, and the lint/build-issue readout — instead of today's opaque single blocking call (`_cli` in `builder.py:616`) whose only feedback is log lines behind `-v`.

A drawing build has real wall-clock (OCC build, the fit/pack search, export), so this fills a genuine 10–30s silence — it is not vanity.

## The load-bearing framing (do this, not "dress up argparse")

The valuable core is **a structured pipeline-event stream**, with the TUI as *one renderer* over it:

- **Engine emits typed events** (stage start/end + timing, the recognised feature inventory, and the existing `_build_issues` lint records — codes/severities are already structured). The CLI subscribes and renders.
- **Presentation stays out of the domain.** Do **not** thread a `console`/renderer into `build_drawing` or the render passes — that re-pollutes the ADR 1 (was 0005/0008) separation just fought for. One narrow event/observer sink the engine writes to; the CLI is a subscriber. The engine must remain headless.
- **Reusable beyond the CLI.** The same event stream is exactly what the Maquetto web-service direction will want (events → websocket → browser progress). Framed as instrumentation it is reusable; framed as "make the CLI pretty" it is throwaway. Build the stream first.

## Design constraints / non-obvious traps

1. **rich, not textual.** `rich` (Progress + Live + Tree) is pure-Python, no native deps — appropriate for an AGPL library dependency. `textual` is a full app framework + event loop and is overkill for a fire-and-forget batch command (you would run an app loop around synchronous compute). Reach for textual only if a genuinely *interactive* use case appears (live re-run with tweaked scale/page, drawing browser) — a separate product.
2. **No fake linear bar.** Stage costs are wildly uneven — most stages are sub-second; OCC-build / pack-search / export dominate. A "N equal steps" bar will sit at one tick for ~90% of the run and *lie*. Prefer a spinner + current-activity line + per-stage elapsed, or weight by real cost. Engaging != fake-accurate.
[...]

PR #291, opened 2026-06-30
feat: Typer CLI — completion, rich help, --version (#289)

### LINKED ISSUE #289: cli: adopt Typer (over argparse) — completion, rich help, and a real command-body home
Replace the argparse `_cli` (`builder.py:616`) with a proper CLI framework, extracted into its own `cli/` module. Don't-code-yet — this is the framework decision + scope.

## Why now (honest framing)

The driver is **not** that the event-stream/TUI work (#276) *needs* a framework — the contextvar `using_sink()` sink is framework-agnostic and would work under argparse too. The real wins are:

- **Shell completion** (bash/zsh/fish) — argparse needs a bolt-on (argcomplete); modern frameworks ship it.
- **Rich-integrated help + errors** — matches the #276 TUI aesthetic; a consistent look from `--help` through to the progress display.
- **A clean command-body home** to wire `using_sink()` + the rich renderer (#276) and the `--format` selector (#288) without the argparse boilerplate.
- **Headroom for a growing surface** — event-stream/TUI now, plausibly `watch` / `inspect` / `completion install` later (and the Maquetto direction).

argparse is a dead end for all four. This is the right moment to switch — it touches the same CLI entry layer as #276 and #288.

## Framework decision

**Recommend: [Typer](https://typer.tiangolo.com/).** Runner-up: Click.

| Criterion | Typer | Click | argparse + argcomplete |
|---|---|---|---|
| Type-hint-driven (matches mypy-clean codebase ethos) | ✅ signatures *are* the CLI | ➖ decorators | ❌ |
| Native `rich` help/tracebacks (serves #276) | ✅ built-in | ➖ via `rich-click` | ❌ |
| Shell completion | ✅ inherited from Click + installer | ✅ built-in | ➖ bolt-on |
| Boilerplate | lowest | medium | highest |
| Magic / layers | thin layer over Click (occasionally leaks Click) | none | none |
| Maturity / longevity | high (widely used) | highest | stdlib |

- **Typer** wins on fit: type-hint-driven matches the codebase, native rich aligns with #276 (which already commits to `rich`, so no *new* weight), completion is free.
- **Click** is the conservative fallback — one fewer abstraction layer, total control; choose it if the team prefers no magic. (Typer pulls Click + rich anyway.)
- **Rejected:** argparse+argcomplete (keeps the verbosity, bolts completion on); cyclopts and other newer/smaller libs (longevity/ecosystem risk for a published library's CLI).

## Scope

- Extract the CLI into its own **`src/draftwright/cli/`** package (also the home #276 wants for the rich renderer — \"renderer up by the CLI\"). Update the entry point (`pyproject.toml` `[project.scripts]`, currently `draftwright.make_drawing:_cli`).
- Port the existing flags 1:1 (`--out`, `--title`, `--number`, `--tolerance`, `--drawn-by`, `--scale`, `--page`, `--script`, `--pmi`, `-v/--verbose`).
- Wire **shell completion** (and a `completion install` path if a subcommand model is adopted).
- Rich-integrated **help + error** output.
- Preserve the **stdout contract**: output paths printed one-per-line for CI/scripts (the TUI/rich chrome goes to stderr or is TTY-gated, never corrupting stdout).
- Leave a clean seam for #276 (`using_sink()` + renderer wrap the command body) and #288 (`--format`).

## Non-goals (resist over-engineering — CLAUDE.md §2)

- **Don't invent speculative subcommands.** Port the single existing command; the structure permits growth, but YAGNI on actual new commands (`watch`, `inspect`) until there's a real need.
- No behavioural change to the drawing or the engine.

## Ordering

Land **before or together with #288** (`--format`) so the new flag is written in the new framework, not implemented in argparse and immediately re-ported. #276 builds on the `cli/` home this creates.

## Open questions (decide before implementing)

1. **Typer vs Click** — ratify the pick. (Recommend Typer.)
2. **Invocation model** — keep bare `draftwright foo.step`, or move to a subcommand namespace (`draftwright build foo.step`, enabling `draftwright completion install` / future `watch`)? The latter is a **breaking UX change** but unlocks the growing surface. Typer/Click can keep a default command, but it's slightly awkward — worth a deliberate call.
3. Whether to expose `completion install` as a first-class command (implies the subcommand model in Q2).

## Cross-refs

- #276 — event stream + rich TUI (shares the `cli/` home; command body wires the sink + renderer).
- #288 — PDF default + `--format` selector (implement in the new framework per Ordering).

pzfreo:
Implements the core of #289: replace the argparse `_cli` with a **Typer** app in a dedicated `src/draftwright/cli.py`. This is the framework foundation the event-stream/TUI work (#276) and the `--format` selector (#288) build on. Typer was chosen in #289 and validated against `estampo/estampo`, which runs the same `typer + click + rich` stack in production.

## What's in

- **New `cli.py`** — a thin Typer front-end over `builder.build_drawing` / `generate_script`. The engine stays headless; no presentation concern leaks below this module (so it's also the home #276 will wrap its sink + renderer around).
- **Shell completion** — `--install-completion` / `--show-completion`, free from Typer (argparse needed a bolt-on). Confirmed present in `--help`.
- **Rich help** — and it **degrades to ASCII box-drawing on a cp1252 stream**, so the existing Windows-safety guard still holds.
- **`--version`** — reports the installed distribution version via `importlib.metadata` (the PyPI version once pip-installed; `unknown` from a bare source tree). Eager, so `draftwright --version` works without a STEP file.

## Scope discipline

- **Framework only.** Every existing flag is ported **1:1** (`--out`, `--title`, `--number`, `--tolerance`, `--drawn-by`, `--scale`, `--page`, `--script`, `--pmi`, `--pdf`, `-v/--verbose`). No behavioural change to the drawing. `--format` (#288) and the TUI (#276) are deliberately **not** here.
- `builder._cli` becomes a **thin shim** to the Typer app, so `python -m draftwright.make_drawing` and existing `from draftwright... import _cli` imports keep working. The `[project.scripts]` entry point now targets `draftwright.cli:app`.
- Removed the now-unused `argparse` / `logging` imports from `builder.py`.
- Kept a single-command app (bare `draftwright foo.step` preserved — **not** a breaking move to `draftwright build …`). The #289 Q2 mixed-model decision is left for when a second command actually arrives.

## Verification

- End-to-end CLI run (SVG + DXF written, paths to stdout, exit 0); `--script`; the `--script`+`--scale` guard (exit 2); `--version`; completion options present.
- `ruff check` + `ruff format` clean; `mypy` clean.
- Entrypoint-help test updated for Typer's capitalised `Usage:` (ASCII guard retained); new `--version` test added.
- **Full fast tier green — 589 passed.**

## Note

`typer` is now a core dependency (it brings `click` + `rich` + `shellingham`). `rich` was already slated to become a base dep by #276/#289, so this isn't net-new weight there.

Closes #289.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,7 @@ dependencies = [
     "build123d-drafting-helpers>=0.13.0",
     "kiwisolver>=1.4,<2",
     "numpy>=1.24",
+    "typer>=0.12",
 ]
 keywords = ["build123d", "cad", "drafting", "engineering-drawing", "technical-drawing", "automation"]
 classifiers = [
@@ -33,7 +34,7 @@ classifiers = [
 pdf = ["cairosvg>=2.7"]
 
 [project.scripts]
-draftwright = "draftwright.make_drawing:_cli"
+draftwright = "draftwright.cli:app"
 
 [project.urls]
 Homepage = "https://github.com/pzfreo/draftwright"
--- a/src/draftwright/builder.py
+++ b/src/draftwright/builder.py
@@ -9,8 +9,6 @@ stage modules -- never make_drawing -- so the graph stays a DAG.
 
 from __future__ import annotations
 
-import argparse
-import logging
 from dataclasses import replace
 from pathlib import Path
 from typing import Literal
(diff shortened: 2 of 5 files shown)


pzfreo:
## Adversarial review — applied (commit 9696b92)

Reviewed for parity with the old argparse CLI, error/edge behaviour, and typing. Two parity fixes applied:

1. **`step_file: Path` → `str`** — typing the argument as `Path` silently normalised it (`str(Path("./sub//x.step"))` → `sub/x.step`; `/`→`\` on Windows) before it reached the engine. argparse passed the raw string; restored. We stringified immediately anyway, so `Path` bought nothing.
2. **`pretty_exceptions_enable=False`** — Typer's rich source-panel traceback buried clean domain errors (`could not read STEP file '…'`) in a developer-style source dump. The old CLI showed a plain traceback on an engine error; matched it. (Verified no locals were ever leaked either way.)
3. Minor: de-unicoded the app `help=` string (`->` not `→`) — latent cp1252 risk if a subcommand ever surfaces the top-level help.

**Not changed (deliberately):**
- `pmi.value` → `Literal[...]` is mildly unsound (`.value` is `str`) but mypy is green (full run: 37 files) and the runtime value is exactly the literal.
- CLI test coverage stays at `--help` + `--version` (build/`--script`/guard paths verified manually); the CLI is thin and the engine is well-covered.

No correctness bug that produces a wrong drawing was found. CLI tests + smoke green; ruff/mypy clean.

Issue #307, opened 2026-06-30
arch: unify detail views — renderers report sub-legible regions → plan_details → generic detail (retire bespoke detailers)


pzfreo:
## Problem

Detail views are currently **ad-hoc, one bespoke detailer per crowding case**:
- `_add_detail_view` — prismatic Z-step crowding (#42), triggered by `_legible_steps`.
- `_add_turned_detail_view` — X-turned head crowding (#304), triggered by sub-floor steps.

They share machinery (crop / project / place / caption) but each hand-codes its
own trigger, region selection, and dimension drawing. A third crowding case (a
dense hole cluster, fine knurling, a tight chamfer chain) would mean a third
bespoke detailer. That doesn't scale and accretes debt.

## Principled approach (mirror the section pipeline)

Sections already do detection → `plan_sections` → `SectionPlan` → `_add_section_view`.
Details should follow the same shape:

1. **One detection signal — "unplaceable at scale".** Renderers already compute a
   legibility floor and skip/stagger/cram. Instead of silently skipping, a renderer
   emits a structured record: `(region_bbox, dropped features/dims, scale_needed)`.
   This is feature-agnostic — turned steps, hole clusters, prismatic steps all
   funnel through the same signal. Trigger becomes "did anything fail to place?",
   not "is this a turned head?".

2. **`plan_details` → `DetailPlan`.** Spatially cluster the flagged regions → one
   detail region per cluster, each carrying the scale it needs. No per-feature logic.

3. **Generic render + re-run the normal renderers.** Crop to the region (axis-
   agnostic box crop), project on the parent view's axis at the detail scale, place
   via largest-empty-rect / repack, then **call the ordinary renderers again in the
   detail view's coordinate system** — `render_step_lengths` / `render_diameters` /
   the hole pass — instead of hand-coding "draw the step chain in the detail". The
   detail is just another view the renderers target.

Result: one detail path, every feature type, the next crowding case free. Squarely
ADR-0008 (detection → plan → render).

## Scope

Fold `_add_detail_view` + `_add_turned_detail_view` into the unified path; renderers
gain a "report unplaceable region" return; add `plan_details`/`DetailPlan`. Do this
*before* a third bespoke detailer is added.

Related: #42, #304, #306 (detail page escalation), #207 (section pipeline precedent).

pzfreo:
Delivered by #311 (merged).

The unified pipeline is in: renderers emit a `DetailRequest` ("I can't draw this
[...]

Issue #306, opened 2026-06-30
layout: detail view should drive page/scale escalation for ultra-fine heads


pzfreo:
## Context

#304 added an auto enlarged detail view for an X-turned part's crowded head. The
detail is placed into the **largest leftover rectangle** on the current sheet, and
its scale is shrunk to fit that rectangle. This works for moderately fine heads,
but an **ultra-fine** feature can out-run the available room:

- GRM-03's 0.5 mm step needs ~14:1 to separate its shoulders to the legibility
  floor (`_MIN_STEP_SEP_MM`). On A4's leftover space the detail is room-limited to
  ~4:1, where the 0.5 mm step is still only ~2 mm — better than in line, but not
  fully clean.
- A head with several sub-floor steps can also drop individual detail dims when
  even the detail scale leaves them too close (the per-step placement skips).

## Proposal

Let the detail view participate in sheet selection instead of only consuming
leftover space:
- When the detail can't reach the scale its smallest gap needs, allow the
  measure-and-repack to **escalate the sheet** (A4 → A3 → …) so the detail gets a
  rectangle big enough, OR
- give the detail its **own sub-sheet / second page**, OR
- cap honestly: if no sheet can host a legible detail, log + lint the residual so
  the user knows (rather than silently landing a too-small detail).

This is the detail-view counterpart of the scaler under-fill (#300): both are the
"use the space / escalate when needed" gap in the layout engine.

## Acceptance

GRM-03's 0.5 mm step is shown at a legible scale in the detail (every detail step
dim placed), or the shortfall is logged/linted. The layout-cleanliness invariant
(#301) still passes.

Found during #304. Related: #300, #301, #302.

Issue #304, opened 2026-06-30
feat: auto detail view for turned heads with sub-floor step lengths (the 'tech-drawing-101' fix)


pzfreo:
## Motivation

Some turned parts have a cluster of fine steps at one end plus a long shaft — the
gramel GRM-03 thumbwheel drive screw is the canonical case (steps 3.2 / 0.5 / 2 /
3 / 20 mm). At any practical scale the fine head **cannot** be dimensioned cleanly
as an in-line step-length chain:

- The 0.5 mm step is ~1 mm on the page at 2:1; two facing arrowheads need ~5.4 mm,
  so a 1 mm-wide dimension bow-ties *within itself* — unfixable by tiering, text
  placement, or arrow flipping (it's geometry, not a placement bug).
- Three shoulders fall within ~6 mm, so even the #293 staggered (two-tier) chain
  leaves two of them sharing a tier with near-touching arrows.
- Legible spacing would need ~11:1 — a 300 mm+ drawing of a 28 mm part.

The #293 staggering fixes *moderately* crowded chains (every gap above the
arrowhead floor). Sub-floor heads need the textbook solution a draughtsman would
reach for: an **enlarged detail view**.

## Proposal

Auto-generate a detail view ("DETAIL A — SCALE n:1") of the turned head when
`find_turned_steps` reports any shoulder gap below the page legibility floor
(~2x arrow_length at the chosen scale). The main front view then carries only the
overall length(s); the detail view carries the fine step lengths + diameters at a
magnification where they're legible.

Scope notes:
- `make_drawing(detail_view=True)` exists but does not currently fire for this
  case — wire the turned-head trigger into it (and consider auto-enabling when the
  chain would otherwise be sub-floor).
- Pick the detail region from the sub-floor shoulder cluster; pick the detail
  scale from the smallest gap (legible floor / smallest gap).
- Place the detail view via the existing layout/repack (it's another view block).
- The step-length renderer should then dimension the head in the detail view, and
  the main chain should carry only the legible (above-floor) lengths + overall.

## Acceptance

GRM-03 renders the head's 0.5/2/3 steps legibly in a detail view, the main views
stay clean, and the layout-cleanliness invariant (#301) passes. Verify visually.

Found during #293. Related: #298 (the ø6 band that also lives in this head),
#300 (scaler under-fill), #301.

PR #308, opened 2026-06-30
feat: auto detail view for turned heads with sub-floor steps (#304)

pzfreo:
## What

A turned part with a fine head and a long shaft — the gramel **GRM-03 drive
screw** — can't have its head dimensioned legibly in line at any sensible scale:
the shoulders are physically too close (its 0.5 mm step is 1 mm on the page at
2:1). The textbook fix is an enlarged detail view; this generates one
automatically.

Result for GRM-03: the main front view locates the head as one `8.7` block (with
an **A** marker box) + the `20` shaft, and **DETAIL A — SCALE 4:1** breaks the
head down into `3.2 / 0.5 / 2 / 3`. No cramming, no double-dimensioning.

## How

- **`_turned_head_band`** (`sections.py`): the crowded X-band — the contiguous run
  of steps below the page arrowhead floor, grown over near-floor neighbours but
  **stopping at a clearly larger step** (so it never swallows the shaft) — plus the
  detail scale that separates the closest shoulder pair.
- **`_add_turned_detail_view`** (`sections.py`): the X-axis sibling of the
  prismatic `_add_detail_view` — crop the band, project front-on at the detail
  scale into the largest free rectangle, draw the head's staggered chain there,
  mark the region, caption it. Skips-with-log (never aborts) when there's no room.
- **Coordination** (`render_step_lengths` + orchestrator): the band is computed
  once; the main chain drops the head steps and shows the head as one block dim
  (located against the shaft), the detail breaks it down. `head_band=None` keeps
  the full chain (unchanged default).
- **Axial coverage** (`linting/coverage.py`): credit shoulders located in EITHER
  the front or the detail view, so a detailed head isn't falsely flagged.

## Latent bug fixed (surfaced here)

`_dim_vertices` read **every** geometry vertex — including the text-glyph outline —
so a wide dim whose centred label sat over interior shoulders could *fake* shoulder
coverage. It now reads the recorded `_dw_spec` witness endpoints. (This is why the
no-room fallback now honestly reports `axial_length_missing` instead of silently
passing.)

## Tests

- `test_subfloor_head_gets_detail_view` — sub-floor head → detail placed, main =
  block + shaft, detail breaks it down, coverage satisfied across both views.
- `test_crowded_chain_staggers_into_two_tiers_at_current_scale` — retargeted to a
  *moderately* crowded chain (above floor) so it asserts staggering, not detail.
- Full fast tier: **610 passed**.

## Adversarial review (tried to refute)

- **`_dim_vertices` ripples to ALL coverage (axial + location).** The 91-test
  coverage cohort and the full 610 tier pass — no location-coverage regression;
  dims without `_dw_spec` fall back to `vertices()` unchanged.
- **Detail fails to place → head undimensioned?** Verified honest: the coverage
  fix makes `axial_length_missing` fire (not silent) when the detail can't fit.
- **`detail_a` vs the repack trigger?** `detail_a` isn't an ortho view, so it's
  exempt from `_annotation_view_overlaps` / `_measure_blocks`, and it's placed in
  the largest rect avoiding every view — no cross-view collision.
- **Double-detail?** An x-turned part has empty `step_zs`, so the prismatic
  `_add_detail_view` can't also fire on `detail_a`.
- **Band swallows the shaft?** Fixed: cluster growth stops at a step ≥ 2× floor
  (the long shaft), verified on both GRM-03 and a synthetic `[1,1,1,1,30]`.

## Known limit / follow-ups

The detail is placed into **leftover sheet space**, so an ultra-fine head (GRM-03's
0.5 mm needs ~14:1) can out-run the room and land lower (4:1 — better than in line,
not perfect). Tracked as **#306** (detail-view page escalation). The two bespoke
detailers (prismatic + turned) should fold into one generic
detection→plan→render path — **#307**.

### DIFF
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -584,7 +584,7 @@ def render_envelope(dwg, groups, a) -> int:
     return n
 
 
-def render_step_lengths(dwg, groups) -> int:
+def render_step_lengths(dwg, groups, head_band=None) -> int:
     """Unified turned step-length chain (ADR 0008 #223) — one IR-driven path that
     replaces the engine's asymmetric X-chain / Z-ladder. Each `StepFeature`'s length
     span is projected into the front view; the chain runs *along the projected axis*
@@ -594,8 +594,14 @@ def render_step_lengths(dwg, groups) -> int:
 
     The chain is collinear (all segments share one offset line) and tiles end to
     end, so every shoulder is located. Crowded labels are spread along the line by
-    the ADR-0003 strip solve (the primitive the engine's X chain already used)."""
-    segs = []  # (page_lo, page_hi, value), in axis order
+    the ADR-0003 strip solve (the primitive the engine's X chain already used).
+
+    ``head_band`` ``(band_lo, band_hi, _)`` (#304): when an X-turned head is too
+    crowded to dimension in line and gets an enlarged detail view, the steps inside
+    the band are NOT chained here — the main view shows the head as one block dim
+    (located against the shaft), and the detail breaks it down. ``None`` (default)
+    keeps the full chain."""
+    segs = []  # (page_lo, page_hi, value, world_lo, world_hi), in axis order
     for g in groups:
         if g.feature_kind != "step":
             continue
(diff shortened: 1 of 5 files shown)


pzfreo:
Closing in favour of the generalized detail-view pipeline (#307).

The high-effort adversarial review of this branch (25 agents) confirmed the
bespoke approach is the wrong shape: alongside real correctness bugs (an
ordering hazard that leaves the head undimensioned when the detail bails, a
lint-crash on Vector dim points, a band-growth heuristic that swallows the
shaft), it flagged ~70 lines copy-pasted from the prismatic `_add_detail_view`.

Rather than patch a second bespoke detailer, we'll build the unified
detect→plan→render path (#307): renderers emit a typed "unplaceable at this
scale" signal, `plan_details` clusters them into a `DetailPlan` (mirroring
`plan_sections`/`SectionPlan`), one generic detailer crops/projects/places, and
the ordinary renderers run again inside the detail view. Both existing detailers
fold into that path, and the layout-visibility gaps (#4/#8 in the review) dissolve.

The working crop/project/place + staggered-chain-in-detail code here, and the
`_dim_vertices` endpoint coverage fix, are good reference for #307. Branch kept.

Issue #426, opened 2026-07-04
Design: record/finalize remaining manual intent verbs through the corridor solve


pzfreo:
## Summary

The #400 Ph2 add verbs (`callout`/`locate`/`furniture`/`dimension`/`section`) place **live** — each call funnels into a per-feature placement helper and does `dwg.add(...)` immediately, **corridor-free**. That is the simple v1, but it means a script-reconstructed drawing is placed incrementally (greedy, avoid-existing-occupancy) rather than by the auto-pass's full ADR-0009 batch solve, so on dense parts it can be busier / less optimally packed and lacks the global crossing-free guarantee (see #424).

This issue proposes the clean end-state: **record-then-finalize batch placement**, and records why it is the natural convergence point that also subsumes #388 Ph2.

## Proposal

Give the verbs a deferred mode:

1. Each verb call **records a deferred intent** (a candidate — e.g. `("callout", featureX)`, `("locate", featureY, datum="A")`) on the drawing instead of placing immediately.
2. A terminal **`dwg.finalize()`** (implicitly run by `export()` if not called) drains all recorded intents through the **same ADR-0009 collect-then-solve orchestration the auto-pass runs** — as one batch.

Result: the emitted script stays fully editable (commenting a verb line removes exactly that recorded intent) **and** placement reaches auto-pass quality (global ordering, crossing-free, optimal packing). This is strictly better than live placement for reconstruction fidelity.

## Why this is the right end-state (not just a re-pack)

- It **subsumes #388 Ph2** (`finalize_drawing` / recompose-around-edits). Once verbs are recorded intents, "recompose the drawing around a user's manual edits" simply *is* "finalize the recorded intent set." The two collapse into one mechanism. The historic blocker for #388 — manual edits weren't re-runnable, so `_repack` would discard them — is exactly what the intent verbs remove. `finalize()` is what makes recompose safe.
- It realizes the ADR 0001 Amendment 1 property ("edits survive a re-solve") at full quality, rather than the soft "runs + lint-clean + droppable" bar the live verbs meet.

## The real obstacle (why this is substantive, not a quick flag)

The auto-pass batch is **not** one uniform "solve this bag of intents" function. `_auto_annotate` (annotations/orchestrator.py) is an **orchestrated sequence of specialized solvers with ordering dependencies**:

- the corridor collect-then-solve (locations + slots; crossing-free by feature order) — `_common.py::place_strip_candidates` / `drain_corridors`;
[...]

PR #441, opened 2026-07-04
#426 tail cleanup: slot_ids filter (#439) + escalation clear (#440) + round-trip coverage (#436)

### LINKED ISSUE #436: test: STEP→script→drawing round-trip is under-covered — parametrize the subprocess run + assert lint across part families
## Summary

The STEP → `generate_script` → run-the-`.py` → drawing round-trip has a coverage seam: the **only** test that executes the emitted script end-to-end (`test_generated_script_runs_and_preserves_pmi`) runs a single simple part (box + one hole) and asserts only *"runs + produces SVG/DXF"* — not that the resulting drawing is lint-clean or correct.

All the richer correctness assertions (lint-error-free, no-worse-than-auto-pass, comment-drops-exactly-that) live on the **in-process** reconstruction tests, which drive `TestFeatureEdits._reconstruct` — a helper that **hand-mirrors** `builder._feature_listing` rather than executing the emitted script text. If the emitted text and the mirror drift, nothing catches it, and no subprocess test covers turned / pattern / sectioned / all-gaps parts end-to-end.

This matters more after #426 Phase 5 (#435): the emitted script now wraps the verbs in `with dwg.deferred():` and relies on `finalize()` running on block exit — a behavior that only the *executed* script exercises, not the mirror.

## Proposal

Parametrize a subprocess round-trip test across representative part families and assert the **executed script's output** is correct — making the emitted text itself the thing under test, not its mirror.

Part families (one fixture each):
- prismatic plate + scattered holes + envelope
- turned multi-step shaft (diameter + step-length chain)
- bolt-circle pattern + central hole (furniture)
- side-drilled bore (flagged-comment gate)
- counterbored Z-hole → section A–A
- all-gaps part (rotational-only, exercises the flat-fallback branch)

For each: `generate_script` → run the `.py` in a subprocess → assert exit 0, SVG+DXF written, **and re-load / lint the produced drawing is error-free** (or assert a lint summary emitted by the script itself, to avoid re-building in the test process).

## Notes / cost

- Subprocess + real OCC build per family is slow — mark `slow` (CI post-merge tier) or keep to 2–3 families in the fast tier + the rest slow.
- Consider having the generated script optionally emit its `lint_summary()` to stdout (behind an env flag) so the test asserts on that rather than re-building, halving the build cost.
- Alternatively (cheaper, weaker): assert the emitted script text and `_reconstruct` stay in sync via a shared single source, removing the drift risk without a subprocess.

Follow-up to #426 Phase 5 (#435). Refs #388 (the `--script` editing surface), #424.

### LINKED ISSUE #439: finalize(): slot_ids lacks the param/role filter its siblings have (live-vs-deferred validation divergence)
Surfaced by the PR #437 (Phase 2b) adversarial review — **low severity, follow-up.**

In `Drawing.finalize()`, `slot_ids` matches *any* `it.kind == "dimension"` whose `feature.kind == "slot"`, with **no `param`/`role` check** — unlike its siblings `len_ids` (requires `param == "length"` AND `role == "step"`) and `dia_ids` (axis filter). A `SlotFeature` exposes only `slot_width`/`slot_length` (both `kind == "length"`), so any other param is invalid.

**Divergence:** the live `Drawing.dimension()` path validates and raises `ValueError` for e.g. `dwg.dimension(slot, "diameter")` or a bare `dwg.dimension(slot, "length")` (ambiguous — needs `role=`). In the **deferred** path there's no record-time validation, so such a bogus intent is silently swallowed: skipped in leg A and routed through `render_slots`, which unconditionally emits width+length+pos regardless of the requested param.

**Impact:** none on the intended flow — the `--script` emitter only ever writes valid `slot_width`/`slot_length` lines, so this is unreachable by generated scripts. It only bites a **hand-edited** script with a malformed slot dimension line, where the deferred path is merely more lenient than the live one. Auto-pass output unchanged; no renumber/dedup/double-place/regression.

**Suggested fix (one-line consistency tightening):** narrow `slot_ids` to
```python
and it.kwargs.get("param") == "length"
and it.kwargs.get("role") in ("slot_width", "slot_length")
```
mirroring `len_ids`, so an invalid slot intent falls through to live replay and surfaces the same `ValueError` the live verb raises (the documented #432 principle for `dia_ids`).

### LINKED ISSUE #440: _auto_annotate leaves consumed _escalations on the drawing → a later deferred edit can re-fire leg D
Surfaced by the PR #438 (Phase 4c) adversarial review — **low severity, benign, follow-up.**

`_auto_annotate` resets `dwg._escalations = []` at start (`orchestrator.py:134`), appends during the passes, but **never clears them after `_maybe_tabulate_holes`** (`orchestrator.py:314`). `finalize()` only *initializes* `_escalations` when the attribute is absent (`drawing.py:967-968`), so escalations from a preceding full `build_drawing(part)` **survive** into a later deferred/finalize batch.

**Failure scenario:** `dwg = build_drawing(dense_part)` (full auto-pass — table placed, ~3 `callout` escalations retained on `dwg._escalations`); then `with dwg.deferred(): dwg.furniture(a_hole)`. `_intents` is non-empty, so finalize's **leg D** runs, reads the STALE auto-pass escalations, passes the callout/location gate, and re-invokes `_maybe_tabulate_holes` — re-adding the fixed-name `hole_table_plan` (identity changes, forced to a different corner as the old table is now an obstacle) and re-placing the balloon ring. The re-add also drops any pin those names carried (`registry.py:147`).

**Impact:** benign in every reproduction — no duplicate `hc_plan` callouts (the auto-pass already removed them, so `replaced` is empty on the second run), no crash, no mis-tabulation, auto-pass byte-identity intact, lint not worsened (`view_annotation_overlap` actually dropped), `_escalations` ends at 0. Harm is **wasted work + table churn / relocation + potential pin loss**, not output corruption. Outside the intended `auto_dims=False` reconstruction flow (where escalations start empty). #438's leg-D `_escalations = []` clear runs AFTER the re-fire, so it only guards the intra-finalize `record→finalize→record→finalize` case.

**Suggested fix (one line):** have `_auto_annotate` clear `dwg._escalations = []` after `_maybe_tabulate_holes` (`orchestrator.py:314`), so no build ever leaves consumed escalations on the drawing. Safe: nothing reads `_escalations` after the table pass in the auto-pass (section rendering, which reads them, runs before it), and lint doesn't read them.

Follow-up to #426 Phase 4c (#438). Will be fixed in the #426 tail-cleanup PR alongside #439.

pzfreo:
Three low-risk follow-ups from the #426 record-then-finalize epic — two one-line finalize/escalation robustness fixes (each surfaced by a prior adversarial review) plus the round-trip test-coverage gap. One commit per issue.

## #439 — tighten `finalize()` `slot_ids` to param/role *(from the #437/Phase 2b review)*
`slot_ids` matched *any* dimension intent on a slot feature, unlike its siblings `len_ids`/`dia_ids`. A `SlotFeature` exposes only `slot_width`/`slot_length`, so a malformed slot dim (`dimension(slot, "diameter")`, or a bare `dimension(slot, "length")` with no role) was silently swallowed on the deferred path instead of raising the `ValueError` the live path raises. Narrowed to `param == "length"` + `role in ("slot_width","slot_length")` so an invalid intent falls through to live replay and surfaces the error. Never reached by the emitter — only a hand-edited script.

## #440 — clear `_escalations` after the auto-pass hole-table pass *(from the #438/Phase 4c review)*
`_auto_annotate` never cleared `_escalations` after `_maybe_tabulate_holes`, and `finalize()` only inits the attribute when absent — so `build_drawing(part)` followed by `with dwg.deferred(): dwg.furniture(f)` made leg D read the *stale* auto-pass drops and re-fire the resolver, relocating the fixed-name `hole_table_plan` (wasted work + potential pin loss). Benign but unintended. One-line clear so no build leaves consumed escalations on the drawing. Safe — nothing reads `_escalations` after the table pass.

## #436 — parametrized STEP→script→drawing round-trip
The only end-to-end test that *executed* the emitted script ran one simple part and asserted only "runs + SVG". The richer assertions lived on the in-process `_reconstruct` mirror (which hand-mirrors `builder._feature_listing`), so drift went uncaught and no subprocess test covered turned/pattern/sectioned/all-gaps parts — matters more since Phase 5 made the emitted script rely on `with dwg.deferred():` + `finalize()` on block exit. New slow parametrized test: one fixture per emitter branch (prismatic holes, turned ø+length, bolt-circle furniture, counterbored→section, all-gaps flat fallback); each exports STEP → `generate_script` → appends a lint epilogue so the **executed** script prints its own drawing's error-lint (no in-process rebuild) → subprocess → asserts exit 0 + SVG/DXF + no lint errors. The emitted text is now the thing under test, not its mirror.

## Verification
Full local gate clean: ruff + `ruff format` + full-package `mypy src/draftwright/`. Fast: #439 malformed-slot + slot-finalize regression (5) green; byte-identity corpus + escalation + hole-table + pattern-balloon (20) green — the #440 orchestrator change is byte-identical. Slow: the 5 round-trip families + the #440-extended Phase-4c WIN test (6) all green.

Closes #439, closes #440, closes #436.

### DIFF
--- a/src/draftwright/annotations/orchestrator.py
+++ b/src/draftwright/annotations/orchestrator.py
@@ -312,6 +312,11 @@ def _auto_annotate(dwg, a: Analysis, *, detail_view: bool = False):
     # every hole — runs last so the table avoids every placed annotation
     # including the title block (#93).
     _maybe_tabulate_holes(dwg, a)
+    # Don't leave the consumed escalations on the drawing: a later deferred edit
+    # (`build_drawing(part)` then `with dwg.deferred(): …`) would otherwise inherit
+    # them and re-fire finalize's leg D against stale drops, relocating the table
+    # (#440). Nothing reads _escalations after the table pass.
+    dwg._escalations = []
 
 
 def _maybe_tabulate_holes(dwg, a: Analysis):
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -1025,10 +1025,17 @@ class Drawing:
         # position dim is model-derived, not recorded). Slots share the location corridor,
         # so they register alongside B2's locations and drain in the SAME solve (the #345
         # dedup of a slot position coincident with a hole location needs one combined pass).
+        # Match on param/role like len_ids above (#439): a slot exposes only the two length
+        # params, so a malformed slot dim (e.g. dimension(slot, "diameter")) falls through to
+        # live replay, where the verb raises the same ValueError instead of being swallowed.
         slot_ids = {
             id(it)
             for it in self._intents
-            if routable and it.kind == "dimension" and getattr(it.feature, "kind", None) == "slot"
+            if routable
+            and it.kind == "dimension"
+            and getattr(it.feature, "kind", None) == "slot"
+            and it.kwargs.get("param") == "length"
+            and it.kwargs.get("role") in ("slot_width", "slot_length")
         }
         only_loc = {it.feature for it in self._intents if id(it) in corridor_ids}
         only_callout = {it.feature for it in self._intents if id(it) in callout_ids}
--- a/tests/test_e2e_standards.py
+++ b/tests/test_e2e_standards.py
@@ -239,3 +239,105 @@ def test_dense_scattered_reconstruction_rebuilds_the_hole_table():
     # avoid an incidental view_annotation_overlap the auto-pass leaves.)
     fin_codes = {i.code for i in dwg.lint() if i.severity in ("warning", "error")}
     assert fin_codes <= auto_codes
+
+
+def _rt_prismatic_holes():
+    from build123d import Cylinder, Pos
+
+    part = Box(80, 60, 20)  # a row of Z-holes → callouts + location dims + centre marks
+    for x in (-30, -10, 10, 30):
+        part -= Pos(x, 20, 0) * Cylinder(3, 30)
+    return part
+
+
+def _rt_turned_shaft():
+    from build123d import Cylinder, Pos
+
+    return Cylinder(15, 30) + Pos(0, 0, 30) * Cylinder(8, 30)  # Z-turned ladder: ø + length
+
+
+def _rt_bolt_circle():
+    import math
+
+    from build123d import Cylinder, Pos
+
+    part = Box(60, 60, 15)  # 6-hole bolt circle → pattern furniture (centre-cross + pitch)
+    for i in range(6):
+        ang = i * math.pi / 3
+        part -= Pos(20 * math.cos(ang), 20 * math.sin(ang), 0) * Cylinder(2.5, 30)
+    return part
+
+
+def _rt_counterbored_section():
+    from build123d import Cylinder, Pos
+
+    part = Box(60, 40, 20)  # a counterbore → the emitter records dwg.section()
+    part -= Cylinder(4, 30)
+    part -= Pos(0, 0, 2) * Cylinder(7, 20)
+    return part
+
+
+def _rt_rotational_boss():
+    from build123d import Cylinder
+
+    # A plain cylinder: a rotational boss (ø via callout()) plus a flagged "rotational"
+    # gap comment for the OD/centrelines the verbs don't reach (#419). Exercises the
+    # boss-callout path + a gap-comment line, distinct from the turned-shaft step chain.
+    return Cylinder(15, 40)
+
+
+# One fixture per emitter path: hole verbs (callout/locate/furniture), turned step
+# diameter + length chain, pattern furniture, section, and a rotational boss + gap comment.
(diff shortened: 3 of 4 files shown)


pzfreo:
Adversarial review verdict: **safe to merge**, all three load-bearing claims confirmed (valid slots route + malformed raise; escalation clear byte-identical; round-trip test meaningful + non-flaky). One LOW finding — the `rotational_gap` fixture comment falsely claimed it exercised the flat no-`with` emitter branch, but a plain cylinder detects as a **boss** and emits `callout(f)` inside the `with` block. Fixed: renamed to `rotational_boss` and corrected the comment to describe what it actually covers (boss-callout + rotational gap-comment), honestly noting the flat-branch + side-drilled gate are covered by unit tests. Comment/rename only — no behavior change (amended into the #436 commit).

Issue #472, opened 2026-07-06
Sheet-DSL validity gap: generated script should reproduce the direct SVG exactly


pzfreo:
## Goal

Running the generated declarative `Sheet` script (`--script --style sheet`) should produce a drawing identical to a direct `draftwright part.step` SVG build. This is the invariant that makes mode-3 generation trustworthy, and it needs to hold before the proposed #473 default flip is low-risk.

## Where we are (measured 2026-07-06)

Compared the direct `build_drawing` annotation set against the generated-sheet-script drawing for three fixtures:

- **Prismatic plate** (holes): **identical** annotation set.
- **Counterbored plate** (carries a `step_level`): **identical** annotation set; the `# ... no declarative verb yet` comment for `step_level` does not cause a drop here.
- **Turned shaft** (stepped cylinder): **two real gaps**:
  1. the front/side **centre lines are dropped** because the declared-model path does not emit the turned axis centrelines;
  2. the base diameter renders as a **`Leader`** in the sheet path versus a **`Dimension`** in the direct path.

So prismatic parts already meet the invariant; the known gap is specific to **turned / rotational** parts. PMI is expected to need the same treatment because it is also currently flagged as `# ... no declarative verb yet`.

## Work

1. Reproduce turned-shaft **centre lines** through the declared-model render path. They come from detection today; a declared model skips detection.
2. Match the base-diameter rendering (**`Dimension` vs `Leader`**) between the two paths.
3. Give `step_level` / `rotational` / `pmi` a **declarative verb** or guarantee the auto-pass reproduces them faithfully from the declared model, so the inline `# auto-dimensioned` comment is truthful.
4. Add a **round-trip parity test**: for a corpus (prismatic + turned + PMI), assert the generated-script drawing's annotation signature equals the direct build's.

## Context

- #473 proposes flipping `--script` to `--style sheet` by default; this issue tracks the fidelity gap that should be closed before that becomes the normal path.
- Emitter: `src/draftwright/sheet_emit.py`; declared-model render: `builder.build_drawing(model=...)` -> `annotations/from_model.py`.
- ADR 0011 Amendment 1 (three authoring modes).

pzfreo:
## Scoped 2026-07-07 (re-measured against current `main`)

[...]

PR #473, opened 2026-07-06
feat: make `--script` default to the declarative Sheet DSL

pzfreo:
## What

`draftwright … --script` now writes the **declarative `Sheet` script by default** (`--style sheet`), for both STEP and `module:attr` object input. The imperative edit-verb reconstruction stays available via `--script --style imperative`.

The declarative DSL — one commentable line per feature — is the natural editable surface (ADR 0011 Amendment 1, mode 3). This flips it to the default; imperative becomes opt-in.

## Fidelity — measured before flipping

Compared the direct `build_drawing` annotation set vs the generated-sheet-script drawing:

| fixture | result |
|---|---|
| prismatic plate (holes) | **identical** |
| counterbored plate (has a `step_level`) | **identical** |
| turned shaft | centre lines dropped; base ⌀ as `Leader` not `Dimension` |

So prismatic parts already reproduce the direct SVG; the gap is turned/rotational (and, expected, PMI). Still a valid dimensioned drawing — just not pixel-identical yet. **Tracked as #472** (goal: sheet script == direct SVG), the agreed follow-up.

## Also

- `--style imperative` now errors clearly on a `module:attr` object spec (imperative reads a STEP file; the default `sheet` references the object) instead of an opaque `import_step` failure.

## Tests

+3 CLI tests (default-is-sheet, imperative-still-available, imperative+object-spec rejected). Full emitter suite (39) + script/CLI e2e (15) green; lint + mypy clean.

## Release

Intended for the **v0.2.8** patch (pyproject already at `0.2.8.dev0`).

Closes nothing on its own; #472 tracks closing the validity gap.


pzfreo:
Assessment after reviewing latest main, #472, and #474: I would not merge the default flip until either the sheet path reaches parity or the rollout is narrowed. The warning is good, but making `--style sheet` the default currently routes users to a path with known turned/rotational fidelity gaps (#472) and ignored title/layout flags (#474). Safer options: keep `imperative` as the default for now, or make the sheet default conditional on supported/prismatic + no ignored flags.

Issue #477, opened 2026-07-06
Umbrella: finish the ADR 0009 unification — one collect-then-solve per shared strip, across all passes


pzfreo:
## What this tracks

ADR 0009 (`docs/adr/0009-boundary-labeling-strip-placement.md`) replaced imperative,
per-pass strip placement with a **collect -> solve -> emit** stage. The phase epic
(#320) is closed and the legacy `Strip.allocate` cursor is gone.

This umbrella tracks the remaining difference between "uses the shared strip/carve
helpers" and the stronger end state: **one `register_corridor` -> `solve_corridor`
collect-then-solve per shared strip, across every pass that writes it**.

## End goal ("fully unified")

A single solve sees the full candidate set for any strip that multiple producers share.
That lets the engine dedup coincident spans, order ladders crossing-free, and rank
over-capacity drops by priority. A pass that runs its own per-strip carve is
obstacle-aware, but it still only sees already-placed output from earlier passes.

## Current state after #524

**Bucket A — in the unified cross-pass corridor solve** (`register_corridor` /
`solve_corridor`):

- hole **location** dims — `render_locations`
- **slot** size/position dims — `render_slots`
- **GD&T** frames / datum symbols / finishes — `render_gdt`
- authored STEP/AP242 **PMI** dimensions — `render_pmi` (#393 fixed by #524)

**Bucket B — obstacle-aware, but still owned by a per-pass solve/carve rather than the
shared corridor batch:**

- **envelope** dims — `render_envelope`
- **turned-diameter / rotational** leaders — `render_diameters` / `render_rotational`
- **height / step ladder** — `render_height_ladder`
- **hole callouts / balloons** — `annotations/holes.py`; front-view callouts are now on
  the strip solver (#513 fixed by #524), but callouts are not a cross-pass corridor batch
- **pitch dims** — `_place_pitch_dim`; axis-aligned placement uses strip carve, and the
  diagonal/rotated fallback is now bounded + obstacle-aware (#514 fixed by #524)
- **raw `Drawing.place_dim`** — remains the explicit unsolved escape hatch; user edit
  intents are tracked under ADR 0012 / #511

**Bucket C — truly ad-hoc fixed-step/count-stack placement:**

- No known remaining Bucket-C placers from the July 2026 audit. #374 is closed; #524 also
  closed the residual front-callout, pitch-fallback, repair, and step-sizing findings.

## Remaining work

[...]

PR #510, opened 2026-07-07
docs: clarify place_dim is an incremental edit, not a corridor re-solve (#396)

### LINKED ISSUE #396: place_dim() uses the single-carve path, not ADR-0009 collect-then-solve (#388 follow-up)
## Context

Split out from #388 (the `--script` human/AI post-edit surface). Non-blocking for #388's acceptance criteria.

`Drawing.place_dim()` lets a script add a dimension into the existing strip space after the automated pass. It currently uses the **single-position carve path** (`place_strip_candidates` / `carve_free_position` for one candidate), not the ADR-0009 **collect-then-solve** batch path that the automatic placers now use.

## Problem

A dimension added via `place_dim()` therefore does not get:
- **Priority selection** — it can't lose to a higher-priority occupant, nor bump a lower one, the way batch candidates do.
- **Corridor-routing / crossing-free ordering** — it's placed at the first free slot, not solved into the datum-ordered ladder (#345/#346).
- **Dedup** against a coincident automatic dimension.

This is acceptable for simple, deliberate edits, but a human/AI editor reasonably expects a `place_dim()` call to obey the same layout guarantees as the automatic pass. Today it can silently place a dim that the batch solve would have ordered or deduped differently.

## Options

1. **Document the limits (recommended first step).** State clearly in the `place_dim()` docstring + the `--script` docs that it is a single-slot placement, not a full corridor re-solve; suggest `finalize_drawing(dwg)` (#388 Phase 2) to recompose afterward.
2. **Route a single candidate through `solve_corridor`.** Wrap the added dim as one `CorridorCandidate` and re-drain the affected strip so it participates in select/order/dedup. This is the "correct" behaviour but means re-solving a strip that was already committed at build time — real work, and it interacts with `finalize_drawing`'s repack.

## Recommendation

Do option 1 now (cheap, honest). Consider option 2 only if post-edit `place_dim()` misplacement shows up as a real problem once #388 Phase 2 lands.

## Related
- #388 (parent), ADR 0009 (collect-then-solve placement).

pzfreo:
## What

Resolves #396 per **the issue's own recommendation (Option 1)**: document that `Drawing.place_dim()` is a single-slot *incremental* placement, not the ADR-0009 collect-then-solve batch — rather than force the re-solve (Option 2).

## Why Option 1 is the correct engineering (not a shortcut)

`place_dim()` runs **after** the build's corridor batch is already committed. The three properties #396 lists — priority selection, crossing-free re-ordering, dedup against a coincident auto dim — are **batch-build** properties. Recovering them post-build would mean **re-solving the strip and moving already-placed dims**, which is:
- surprising for a deliberate incremental edit, and
- properly the job of a **recompose** (`finalize_drawing` / rebuild), not a single placement.

The issue author flagged exactly this ("re-solves a strip already committed at build time — real work, interacts with finalize_drawing's repack") and recommended Option 1. `place_dim` is already **obstacle-aware** (the carve clears every placed annotation), so it has no invisible-occupant collision class — it shares no strip with a *concurrent* build pass, so it is **not** an ADR 0009 co-solve gap (#477).

## Change

Docstring only: state the single-carve is **by design**, enumerate the three properties it lacks and where they come from (the recompose), and point at `finalize_drawing` (when available) like `drop()` does. README already says "auto-placed" (accurate) — unchanged.

Docs-only; existing `place_dim` tests unchanged.

Closes #396.

### DIFF
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -570,10 +570,17 @@ class Drawing:
                 :meth:`drop` / :meth:`annotations_of` can find it (#398).
             **kwargs: forwarded to ``Dimension`` (e.g. ``label=``, ``tolerance=``).
 
-        Uses the single-position strip carve, not the ADR-0009 collect-then-solve
-        path the automatic placers use — fine for adding a dimension into free
-        space, but it does not re-solve the strip or dedup against existing dims
-        (#396). Prefer :meth:`dimension` for a feature-referenced edit.
+        Uses the single-position strip carve (obstacle-aware: it clears every
+        placed annotation), **not** the ADR-0009 collect-then-solve path the
+        automatic placers use — **by design** (#396). ``place_dim`` is an
+        *incremental edit* run after the build's corridor is already committed, so
+        it deliberately does not re-solve the strip: a full solve would reorder or
+        dedup **already-placed** dims (surprising for a deliberate edit). It
+        therefore has no cross-pass priority selection, crossing-free re-ordering,
+        or dedup against a coincident auto dim — those are properties of the batch
+        build, recovered by a *recompose* (:func:`finalize_drawing`, when
+        available) or a rebuild, not of a single incremental placement. Prefer
+        :meth:`dimension` for a feature-referenced edit.
 
         Falls back to a fixed ``slot`` offset when the strip is full or when no
         layout analysis is available (e.g. when ``auto_dims=False`` was not used


pzfreo:
Parking: #396's correct resolution isn't to document the single-slot limit — it's to REMOVE it. Superseded by the unified edit-into-global-solve design (ADR + tracking issue incoming): user dimension edits become pinned, priority-ranked candidates in the one corridor solve.

PR #544, opened 2026-07-09
Finish ViewBlock layout authority

### LINKED ISSUE #112: layout: derive ViewBlock footprints from composed annotation boxes
## The architecture gap

Today the layout places the **bare orthographic views** (`_layout_geometry`) and then annotations are added afterward, with space reserved indirectly by widening the inter-view corridors (`StripDepths`, and `pv_halo` for balloons — #111). This is "place the views, then reserve margins."

The cleaner model is **compose-then-pack**: build each view as a *block* whose bounding box already includes its dimensions, callouts, and (when ballooned) its balloon ring, then lay out the complete blocks so they tessellate without overlap. "Build the views, then lay out the views."

## Why it matters

- A view's true footprint (geometry + halo + dims) becomes a first-class `Placeable`, so packing is exact rather than approximated by per-side corridor padding.
- Balloons stop relating awkwardly to a neighbouring view (e.g. the front view directly below the plan view) — the plan block carries its balloon ring and is positioned as a unit.
- It unifies with the constraint-based layout direction (ADR 0003): views as `Placeable` blocks with composite sizes, solved together with the annotations.

## Relationship to existing work

- #111 added a `pv_halo` reservation as a pragmatic approximation of this — it makes room for the balloon ring but does not compose the block. Good enough for normal parts; the block model is the proper version.
- Pairs with #94 (2D solve) and helpers#144 (once patterns are detected, far fewer parts balloon, shrinking how often a balloon block is even needed).

## Acceptance

- Each view exposes a composite footprint (geometry + its annotations/balloons).
- `_layout_geometry` packs those footprints (via the layout solver) rather than padding corridors by scalar bands.
- No regression on the existing geometry/golden tests.

pzfreo:
## Summary
- make FV/PV vertical centering use the same composed ViewBlock gap in estimated and measured layout paths
- remove the old estimator-only plan-view lift exception
- add a regression test that pins estimator centering to composed block footprints

Closes #112.

## Validation
- uv run pytest tests/test_make_drawing.py::TestComposeViewBlocks tests/test_make_drawing.py::TestDynamicCorridors tests/test_make_drawing.py::TestTwoPassLayout tests/test_make_drawing.py::TestComposeThenPackRepack -q
- uv run pytest tests/test_make_drawing.py -q
- uv run pytest -q
- uv run ruff check src/draftwright/sheet.py tests/test_make_drawing.py

### DIFF
--- a/src/draftwright/sheet.py
+++ b/src/draftwright/sheet.py
@@ -903,7 +903,6 @@ def _layout_geometry(
     pv_hh = y_size * scale / 2
     sv_hw = y_size * scale / 2
 
-    pv_below = _est_pv_below_depth()
     est_blocks = _compose_view_blocks(
         x_size, y_size, z_size, scale, strips, n_steps, section=section
     )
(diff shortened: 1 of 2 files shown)


pzfreo:
Adversarial review: logic checks out — the removal of the estimator-path halo lift-trick is algebraically byte-identical when `halo == 0` (confirmed by hand), and intentionally changes output only when the plan view balloons, unifying with the measured/repack path per #112.

One test-coverage gap: `test_estimator_vertical_stack_is_centred_from_composed_blocks` (tests/test_make_drawing.py:1123) calls both `_compose_view_blocks` and `_layout_geometry` with `strips=None`, which forces `halo=0` — exactly the case that was already correct before this PR. It doesn't pin the halo>0 behavior this PR actually changes. There's indirect coverage via `TestFinishedSheetLayout::test_finished_sheet_has_no_layout_collisions[_bolt_circle]` (lint-clean on a ballooned part end-to-end), but nothing asserts exact FV/PV/SV coordinates under ballooning, so a future regression in the halo path could slip through silently.

Suggest adding a halo>0 variant of the new test before merging (not a blocker otherwise — CI is green, no correctness bugs found).

Issue #568, opened 2026-07-11
Unify the recognizer contract across draftwright and vendored drafting-helpers recognizers


pzfreo:
## Summary

The feature recognizers in `recognition/` have grown organically and diverge in shape. As #559/#555/#560 added new ones, the inconsistency became visible. Establish a single uniform recognizer contract and retrofit the existing recognizers (including the ones vendored from `build123d-drafting-helpers`) to it.

## Current inconsistencies

| recognizer | module | returns | IR shape |
|---|---|---|---|
| `find_holes` / `find_bosses` / `analyse_cylinders` / `find_hole_patterns` | `_features.py` (vendored from helpers) | plain dicts / recogniser types | `HoleFeature` / `BossFeature` / `PatternFeature` |
| `find_turned_steps` | `turned.py` | `TurnedProfile \| None` (dataclass) | `StepFeature` |
| `find_slots` | `slots.py` | `list[Slot]` (dataclass) | `SlotFeature` |
| `find_plates` (#559) | `plates.py` | `list[Plate]` (dataclass) | `PlateFeature` (standalone) |
| `find_chamfers` (#560) | `chamfers.py` | `list[Chamfer]` (dataclass) | `ChamferFeature` (standalone) |
| `analyse_face_levels` | `levels.py` | `list[float]` | `StepLevelFeature` |
| `find_step_shoulders` (#555) | `levels.py` (folded in) | **`list[(axis, pos)]` raw tuples** | field on `StepLevelFeature` |

Divergences:
1. **Return type** — most return frozen dataclasses (`Slot`/`TurnedStep`/`Plate`/`Chamfer`), but `find_step_shoulders` returns **raw tuples** and the vendored `_features.py` recognizers return dicts/ad-hoc types. The tuple case is the clear wart.
2. **Module home** — dedicated module per recognizer (`plates.py`, `chamfers.py`, `slots.py`, `turned.py`) vs folded into a shared module (`find_step_shoulders` in `levels.py`).
3. **IR mapping** — standalone `Feature` type (plate/chamfer/slot) vs a **field** on another feature (step shoulders on `StepLevelFeature`).
4. **Vendored recognizers** (`_features.py`) still carry the upstream `build123d_drafting.features` shape (dicts, different naming), not the draftwright dataclass idiom — a straddle left over from the ADR 0007 vendor-and-own move.

## Proposed uniform contract

[...]

Issue #555, opened 2026-07-11
Prismatic step position not dimensioned (two-level block under-constrained)


pzfreo:
## Summary

A prismatic part with a step (rebate / two height levels) gets its overall envelope and its two heights dimensioned, but **not the position of the step along the part** — so where the shoulder falls is left implicit. The part is under-constrained: two different geometries produce the same sheet.

## Reproduce

```python
from build123d import Box, Pos, export_step
block = Box(80, 60, 30)                        # 80 x 60 x 30
block -= Pos(0, -15, 7.5) * Box(80, 30, 15)    # rebate: front half lowered to 15
export_step(block, "stepped.step")
```

```
draftwright stepped.step --title "Stepped Block" --format pdf
```

**Observed** (side/profile view shows the L clearly): overall `80` (width), `60` (depth), `30` and `15` (the two heights). But the depth at which the step occurs — here 30 of the 60 — is **not** dimensioned. The step could sit anywhere along the 60 and the sheet would look the same.

**Expected**: a location dimension for the step edge along its axis (e.g. `30` from the reference face to the shoulder), so the two levels are fully constrained.

## Impact

Found while building CADGenBench self-test fixtures. The stepped block is a first-year orthographic staple, yet its auto drawing can't be reconstructed unambiguously — an agent (or human) has to guess the shoulder position from the drawn proportions. Whole prismatic-step / rebate / shoulder class is affected.

Repro on `0.2.14.dev0`.

pzfreo:
## Acceptance test case

The step's **along-axis position** is not dimensioned. Uses an *asymmetric* step so the position can't hide behind another value:

```python
from draftwright import build_drawing
from build123d import Box, Pos

def _labels(d):
    return sorted(str(o.label) for o in d._named.values() if getattr(o, "label", None))

def test_step_position_dimensioned():
    # shelf 20 deep at the front, back 40 deep, lowered by 15
    part = Box(80, 60, 30) - Pos(0, -20, 7.5) * Box(80, 20, 15)
    dwg = build_drawing(part, number="X")
    lbl = _labels(dwg)
    assert {"80", "60", "30", "15"} <= set(lbl)     # overall + heights already present
    assert "20" in lbl or "40" in lbl               # step position / shelf depth
```

**Current (failing):** `_labels` = `['15', '30', '60', '80', ...]` and `lint_summary()["by_code"]` is empty — so the missing step position isn't even flagged. The `assert "20" in lbl or "40" in lbl` fails.

PR #571, opened 2026-07-12
refactor: establish the uniform recogniser contract (ADR 0013 Phase 1, #568 step 0)

pzfreo:
## What

**Step 0 of ADR 0013 Phase 1** — establish the recogniser contract as landed code (the foundation later recognisers are written to and existing ones migrate onto), proven on one exemplar, plus fix the callout crack. **No behaviour change** — `test_layout_snapshot` is byte-identical.

Part of #568 (uniform recogniser contract). Hard rename, **no back-compat aliases** — set decisively, no two-idiom limbo.

## Changes

- **Hard rename** every *feature* recogniser `find_`/`analyse_` → `recognise_` across all call sites (`detect`/`analysis`/`coverage`/`_core`/`orchestrator`/`drawing`/`__init__` + tests): `recognise_holes`/`bosses`/`hole_patterns`/`slots`/`plates`/`chamfers`/`turned_steps`/`step_shoulders`/`face_levels`.
  - Cylinder-analysis **substrate** (`analyse_cylinders`/`full_cylinders`/`feature_diameters`) **keeps its names** — a tuple-of-dicts / diameter query, not a `list[record]` recogniser. Forcing it into the contract would be the wrong abstraction.
- **Exemplar retype:** `recognise_step_shoulders` → `list[StepShoulder]` (frozen dataclass) with keyword-only injected `levels`, killing the raw `(axis, pos)` tuple wart (#555). `detect.py` adapts it to the IR's existing tuple, so **no recognition object crosses the IR boundary** (ADR 0008 Am6) — byte-identical.
- **Callout crack fixed:** `C{leg}`/`{leg} × {angle}°` formatting moved off `ChamferFeature` (dropped `.callout()`/`.equal_leg`) into `from_model._chamfer_label` — the IR feature is now pure data like every other (ADR 0013 §7).
- Tighten `recognise_holes`/`bosses` return annotations to `list[HoleFeature]`/`list[BossFeature]`.
- **Delete** the dead `draftwright.features` compat shim (its only reason to exist — the `find_slots` back-compat name — is hard-gone; nothing imports it).
- Document the contract in `recognition/__init__.py` + CLAUDE.md.

## Deferred (tracked by #568)

- Return-shape retypes of `recognise_turned_steps` (`TurnedProfile | None` — genuinely 0-or-1 profile, a real design call) and `recognise_face_levels` (`list[float]`).
- Normalising the `_features.py` record idiom.
- `codespell` wiring — separate follow-up (by default it flags typos, not British-vs-American, so a "British enforcement" banner needs its own config).

These are safe per-recogniser sequencing behind byte-identity tests, gated by #568 — **not shim-debt**.

## Verification

`ruff check` + `ruff format --check` + `mypy` clean; import sanity; and green: `test_layout_snapshot` (byte-identity), `TestChamferCallout`, `test_recognition`, `test_turned_steps`, `test_part_model`, `test_e2e_standards`, `-m smoke`.


pzfreo (commented):
Two blockers before merge:

1. **The documented uniform contract is not established by this patch.** `recognition/__init__.py` says every recogniser takes `part`, keyword-only dependencies/tuning, and returns a deterministic list of frozen records. Several renamed functions still violate that contract: `recognise_hole_patterns(holes)` has no `part` and positional inventory; holes/bosses accept positional `cyls`; plates/chamfers retain positional tuning; and turned steps/face levels violate the return shape. The docs only acknowledge the last two. Either complete these signatures now, or accurately frame this as the naming step and enumerate all remaining violations under #568.

2. **The hard rename removes published compatibility surfaces without a deprecation/release path.** `analyse_face_levels` was top-level, recognition names were publicly importable, and `draftwright.features` was explicitly a deprecated compatibility module. This patch deletes all old names and the shim without a changelog entry or migration guidance, despite ADR 0007 choosing deprecate-and-vendor/full backward compatibility. Keep warning aliases through an announced breaking release, or land this only in a version where API breaks are permitted.

Also, replace “no behaviour change” with “no drawing-output change”: downstream imports and callable signatures are observable behavior and now fail immediately.

Focused verification passed locally: 98 tests across recognition, part-model, and turned-step suites.

pzfreo:
Both blockers addressed (pushed); the wording fix too.

**1. Contract doc no longer over-claims — and most signatures now actually conform.** I made the tuning/deps **keyword-only** everywhere it was cheap (zero caller churn — every caller already passed keywords): `recognise_holes`/`recognise_bosses` (`cyls`), `recognise_chamfers` (`tol`/`max_leg_frac`), `recognise_plates` (tuning), `recognise_face_levels` (tuning). So **naming + keyword-only signatures now hold for every recogniser.**

`recognise_hole_patterns(holes)` I kept positional deliberately: it's a **derived** recogniser — a pure function of another recogniser's records, no `part` and no tuning, so its single inventory arg is unambiguous. Rather than force a fake `part` (or churn 23 call sites), the contract doc now defines **two shapes** — part-based `(part, *, tuning/deps)` and derived `(inventory)` — and enumerates honestly what does *not* yet hold, all under #568:
- return shapes: `recognise_turned_steps` (`TurnedProfile | None` — genuinely 0-or-1 profile; whether it's even a `list[record]` recogniser is an open call) and `recognise_face_levels` (`list[float]`);
- the `_features.py` record idiom (name-clashes with the IR `Feature` types).

**2. Release hygiene for the intentional break.** No aliases — the hard break is deliberate (the two-idiom debt is exactly what we're avoiding). But you're right it needs a paper trail: added a CHANGELOG **Unreleased / Breaking** entry with the full rename map + migration (incl. `draftwright.analyse_face_levels` → `draftwright.recognise_face_levels` and the `draftwright.features` removal), and flagged that **this must ship in a 0.3.0 minor, not a patch**.

**Wording:** adopted "no *drawing-output* change (byte-identical)"; imports/signatures intentionally break.

Re-verified: ruff + mypy clean; `test_recognition` + `test_turned_steps` + `test_layout_snapshot` (byte-identity) + `-m smoke` all green. CI was already 16/16 on the prior head.

PR #572, opened 2026-07-12
refactor: recogniser records + conform turned/face-level recognisers (#568)

pzfreo:
## What

**Step 1 of the recogniser contract** (#568): finish making the recogniser layer conform to ADR 0013 §2 (`recognise_<feature>(part, *, …) -> list[<record>]`), with **zero exceptions**.

## Changes

- **Name-clash fix.** Renamed the recognition records that collided with the IR `Feature` types → `HoleRecord`/`BossRecord` (IR types unchanged).
- **`recognise_turned_steps -> list[TurnedStep]`** (was `TurnedProfile | None`; `[]` for a non-turned part). Root fix was the **record shape**: `TurnedStep` gained an `axis` field, so a step is self-contained. `TurnedProfile` survives only as a **pipeline aggregate** (`TurnedProfile.from_steps`, which rejects mixed axes + sorts) — `detect`/`analysis`/`coverage` wrap the list via it, so the `a.prof` pipeline is byte-identical.
- **`recognise_face_levels -> list[FaceLevel]`** (was `list[float]`).

## Why conform (not relax the ADR)

An earlier revision blessed the two stragglers as scalar/singular exceptions. Review + maintainer call: **ADR §2 is right; the records were the wrong shape.** Fixed the records instead — ADR 0013 §2 stands unamended.

## Notes

- **No drawing-output change** (`test_layout_snapshot` byte-identical incl. turned fixtures; e2e/part_model green).
- Part of the **0.3.0 breaking bucket** — CHANGELOG updated.

## Verification

`ruff` + `mypy` clean; `test_recognition`, `test_turned_steps` (incl. new `from_steps` adversarial tests), `test_layout_snapshot`, `test_e2e_standards`, `test_part_model`, `-m smoke` all green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ unchanged (byte-identical); only import paths and callable signatures change.
   `feature_diameters`) keeps its names.
 - **`recognise_step_shoulders` now returns `list[StepShoulder]`** (a frozen dataclass)
   instead of raw `(axis, position)` tuples; `levels` is keyword-only.
+- **Recognition record classes renamed to avoid clashing with the IR `Feature` types:**
+  `draftwright.recognition.HoleFeature` → `HoleRecord`, `BossFeature` → `BossRecord`
+  (the IR `draftwright.model.ir.HoleFeature`/`BossFeature` are unchanged).
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ unchanged (byte-identical); only import paths and callable signatures change.
   `feature_diameters`) keeps its names.
 - **`recognise_step_shoulders` now returns `list[StepShoulder]`** (a frozen dataclass)
   instead of raw `(axis, position)` tuples; `levels` is keyword-only.
+- **`recognise_turned_steps` now returns `list[TurnedStep]`** (empty for a non-turned
+  part) instead of `TurnedProfile | None`. `TurnedStep` gained an `axis` field (each step
+  is now a self-contained record); `TurnedProfile` remains only as a pipeline aggregate
+  (`TurnedProfile.from_steps(steps)`), no longer a recogniser return.
+- **`recognise_face_levels` now returns `list[FaceLevel]`** (a frozen `FaceLevel(z)`
+  record) instead of `list[float]`.
 - **Recognition record classes renamed to avoid clashing with the IR `Feature` types:**
   `draftwright.recognition.HoleFeature` → `HoleRecord`, `BossFeature` → `BossRecord`
   (the IR `draftwright.model.ir.HoleFeature`/`BossFeature` are unchanged).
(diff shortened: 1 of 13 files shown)


pzfreo:
**Adversarial review (4 lenses, each finding triple-verified): 0 surviving blockers, 1 confirmed NIT** (a second finding was refuted 2-of-3 and dropped).

**Confirmed NIT (governance):** the refined `recognition/__init__` contract doc blesses `recognise_face_levels -> list[float]` and `recognise_turned_steps -> TurnedProfile | None`, which contradicts **ADR 0013 §2**'s absolute "always `list[record]`, no `Optional`-singular, no bare `list`" — a silent ADR divergence CLAUDE.md forbids.

**Resolved by amending the ADR, not diverging:** added **ADR 0013 Amendment 1** — §2's absolute rule was over-strict (and, honestly, was written knowing `recognise_turned_steps` was a counter-example). The amendment records the correction: scalar/singular returns are an accepted shape where a record is ceremony; the uniform *intent* stands. Also fixed the stale CLAUDE.md line.

Docs-only change; no code/behaviour impact. The rename + CHANGELOG were already correct.

pzfreo:
Independent review found one substantive issue and one documentation nit:

1. **Major — `TurnedProfile.from_steps()` does not enforce the aggregate invariants it claims.** It chooses `steps[0].axis` and accepts every remaining record unchanged. Mixed-axis, unsorted, overlapping, or discontinuous steps therefore produce a profile whose `axis` and `shoulders` silently misrepresent its contents. The old recogniser-built `TurnedProfile` guaranteed these invariants by construction; moving aggregation into a public classmethod creates a new validation boundary. Please reject mixed axes and invalid spans, and either sort steps or require and validate strict axial ordering/contiguity. Add adversarial tests for mixed axes, reversed input, overlap, and gaps.

2. **NIT — the PR description describes the abandoned scalar/singular implementation.** The current head introduces `FaceLevel` and self-contained `TurnedStep` records, so the summary, rationale, and verification text should be updated to describe the final three-commit result.

No other correctness regressions found. Focused recognition, turned-step, part-model, and layout-snapshot tests passed locally.

pzfreo:
Both addressed (pushed):

**Major — `from_steps` invariant.** You're right that moving aggregation to a public classmethod created a validation boundary the old invariant-by-construction didn't need. Fixed proportionately: `from_steps` now **rejects mixed axes** (raises — that's the one invariant whose violation silently corrupts `axis`/`shoulders`) and **sorts by `lo`** so `shoulders` is order-robust. The sort is a **no-op for real recogniser output** (already sorted), so snapshot byte-identity holds. Added adversarial tests: mixed-axis raises, unsorted input sorts, empty → None.

I deliberately did **not** add contiguity/overlap/gap validation — that's defensive code against a scenario no caller produces (the recogniser guarantees contiguous non-overlapping spans), so I documented it as a caller precondition rather than re-validating spans in the aggregate. Happy to add it if you'd prefer belt-and-suspenders.

**Nit — stale PR description.** Rewritten to describe the final result (record rename + conform via self-contained `TurnedStep`/`FaceLevel`), not the abandoned bless approach.

PR #677, opened 2026-07-17
feat(export): unified export(formats=…) API + PNG raster export

pzfreo:
Cleans up the ad-hoc export surface and adds **PNG** (key for AI-based interactions).

## API
- **`Drawing.export(out, *, formats=("pdf",), dpi=150) → {format: path}`** — one entry point over `svg`/`dxf`/`pdf`/`png`, returns a dict, and internally handles the **SVG → PDF → PNG** intermediate chain (writing/removing intermediates it wasn't asked to keep). `formats` accepts a single name or an iterable.
- **CLI `--format` gains `png`** (`all` now expands to all four); `_emit` collapses to a single `export(formats=…)` call — the old temp-SVG-write-then-delete dance is gone.

## PNG rasteriser — permissive, no cairo
SVG → PDF → PNG via **pypdfium2** (Apache-2.0, wrapping Google PDFium BSD-3) + **Pillow** (HPND) — both permissively licensed and pure-wheel, so PNG works cross-platform with **no native cairo** (ADR 0006) and keeps the render path **dual-license-clean**. (reportlab-native `renderPM` was ruled out — its wheels here lack the `_rl_renderPM` accelerator and fall back to demanding cairo.)

## Deprecation (staged, as discussed)
- `export_pdf()` now **warns** → use `export(formats=("pdf",))["pdf"]`.
- The legacy `export(svg=, dxf=)` boolean/tuple form is **kept working** and marked legacy in the docstring (no runtime warning yet). Hard-deprecating it + migrating first-party callers (`make_drawing`/`Sheet`/the script emitters) is a deliberate **follow-up** to keep this PR focused.

## Licence note
Audited: all new deps are permissive (BSD/Apache/HPND). Existing `svglib` is **LGPL-3.0** (weak-copyleft, already in the tree for PDF) — flagged in the CHANGELOG/CLAUDE.md; purging it is a separate job.

## Verification
- New `TestExportFormats`: dict return for all four formats, PNG validity (magic bytes) + intermediate cleanup, `dpi` scaling, unknown-format `ValueError`, `export_pdf` deprecation warning. `TestFormatSelector` updated to the new `_emit` + `png`.
- Four lint checks clean; e2e / pmi / sheet_emit sweep green; `uv.lock` synced; CLAUDE.md + CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- **PNG raster export** — `--format png` on the CLI, or `Drawing.export(formats=("png",))`.
+  Rendered SVG → PDF → PNG via **pypdfium2** (Apache-2.0, Google PDFium BSD-3) + **Pillow**
+  (HPND) — both permissively licensed and pure-wheel, so PNG works cross-platform with **no
+  native cairo** (ADR 0006) and keeps the render path dual-license-friendly. `dpi=` sets the
+  raster resolution.
+
+### Changed
+
+- **Unified export API** — `Drawing.export(out, *, formats=("pdf",)) → {format: path}` is now
+  the single entry point over `svg`/`dxf`/`pdf`/`png`; it returns a dict and internally handles
+  the SVG→PDF→PNG intermediate chain (writing/cleaning up intermediates it wasn't asked to keep).
+  The CLI `--format` gains `png`.
+
+### Deprecated
+
+- `Drawing.export_pdf()` — use `export(formats=("pdf",))["pdf"]`; the method now warns.
+- The legacy `export(svg=, dxf=)` boolean keywords and their `(svg_path, dxf_path)` tuple return
+  — kept for back-compat; prefer `formats=[...]` (the dict API).
+
 ## v0.3.0 — 2026-07-14
 
 **Breaking — feature-recogniser API renamed** (ADR 0013 / #568): *deliberate* breaks with
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -139,12 +139,15 @@ entry. Keep `_LAYERS` and this section in step.
   via the package surface.
 - **`fonts.py`** — vendored, path-pinned IBM Plex fonts for deterministic
   cross-platform layout (ADR 0006).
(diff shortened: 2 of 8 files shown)


pzfreo:
**Blocking: PNG/PDF export can delete existing outputs**

In `Drawing.export()` (`src/draftwright/drawing.py`, around lines 2360–2384), intermediates are written directly to the user-visible `<out>.svg` and `<out>.pdf` paths and then unconditionally removed when those formats were not requested.

For example:

```python
dwg.export(out, formats=("svg", "pdf"))
dwg.export(out, formats="png")
```

The second call overwrites and then deletes both previously requested files. PDF-only export similarly deletes an existing SVG with the same stem.

Please use temporary paths for intermediates, or otherwise track whether each file existed before this export and only clean up files created solely as intermediates by the current call. A regression test that performs the two calls above and verifies the SVG/PDF remain would cover this.

pzfreo:
Good catch — that was a genuine data-loss bug, thank you. Fixed in the latest commit.

**Root cause:** the new `formats=` path wrote the SVG/PDF **intermediates to the user's `<out>.svg` / `<out>.pdf`** and then `os.remove`'d them, so `export(out, formats="png")` after `export(out, formats=("svg","pdf"))` overwrote *and deleted* the earlier files.

**Fix:** intermediates now go to a `tempfile.TemporaryDirectory()` when not themselves requested — **never the user's stem** — and are auto-removed when the `ExitStack` closes. No `os.remove` of user paths at all. Added the exact regression test you described (`export((svg,pdf))` then `export("png")`; assert the SVG/PDF survive **byte-identical**).

While in there, I also folded in the other findings from the Codex review of this PR:
- `self.dxf_path` reset each call (was left stale when dxf wasn't requested);
- cleanup is now exception-safe (the temp dir is removed even if a render raises — the old `os.remove`-after-render skipped cleanup on failure);
- `dpi<=0` rejected before any write;
- single-string `formats="PNG"` normalises like `("PNG",)`.

One I left as-is (noted): the stem-suffix strip is now format-aware for all four extensions, so the *deprecated* legacy `export("revision.pdf")` yields `revision.svg` rather than `revision.pdf.svg` — a minor improvement on the deprecated path. Shout if you'd rather I keep the legacy strip byte-identical.

Issue #602, opened 2026-07-13
perf: separate annotation placement footprints from OCCT rendering


pzfreo:
## Summary

Automatic annotation placement constructs complete build123d/OCCT geometry while searching candidate positions. Rejected candidates therefore pay nearly the full rendering cost. On a modest 15-hole grid this makes automatic drawing generation take roughly one minute even though recognition and projection complete in under one second.

The placement architecture should search using lightweight analytical footprints, then render each accepted annotation once.

Tested on `main` at `57ffd896fa3f`. Timings below are local measurements and intended as relative profiling evidence, not stable CI thresholds.

## Measured evidence

Synthetic rectangular plate with 15 holes arranged as two regular grid patterns:

| Stage | Time |
|---|---:|
| Feature recognition | ~0.22 s |
| Drawing without automatic dimensions | ~0.78 s |
| Drawing with automatic dimensions | ~63-67 s |
| `_analyse` | ~0.14 s |
| `_assemble` | ~61.64 s |
| Repack | ~0.04 s |
| Repair | ~1.04 s |
| `_annotate_holes` | ~57.89 s |
| `_add_grid_pitch_dims` | ~50.54 s |

Four pitch dimensions resulted in 139 `_dim` calls and roughly 50 seconds inside dimension construction. Profiling also recorded about 1,640 `BRepBndLib.AddOptimal_s` calls in a representative dense run.

The primary hotspot is the fallback loop in `src/draftwright/annotations/holes.py::_place_pitch_dim`. Every candidate offset executes `_make(...)`, creates complete dimension geometry, and obtains an OCCT bounding box before collision rejection.

## Related instances of the same pattern

- `annotations/_common.py::place_strip_candidates` calls `build(lo)` to measure perpendicular bands and calls `build(pos)` again during candidate evaluation. Corridor draining accounted for about 2.3 seconds in the same grid benchmark.
- `annotations/_common.py::strip_obstacles` rescans placed annotations and recomputes OCCT bounding boxes for repeated occupancy queries.
- `Drawing.export()` traverses and converts every shape independently for SVG and DXF. On a scattered ten-hole drawing, shape addition took about 2.2 seconds for SVG and 2.7 seconds for DXF.
- Lint precomputes label boxes but later recomputes full annotation boxes for page bounds and other passes. Repeated lints on the same drawing remained around 0.7-0.9 seconds despite the view-edge cache.
[...]

PR #681, opened 2026-07-17
perf(#602): memoise annotation bounding boxes across lint checks and calls

pzfreo:
Part of #602 (queue item 5: lint box reuse — the last measured draftwright-owned hotspot). Follows #678/#679/#680.

## What

An optimal OCC `bounding_box()` on fused annotation geometry costs ~10 ms, and the structural lint checks each measured independently — the page-bounds loop per item **per call**, `_lint_dim` again, and the centreline-pair / view-overlap checks once **per pair** for label-less items. On the scattered-plate benchmark that's 57 box computations per lint — 0.77 s of the 0.88 s pass — re-paid on **every `export()`** and every repair-loop iteration.

All full-box lookups now go through one memo, `_ann_box(item, cache)`: id-keyed, identity-checked (the entry stores the item, which also prevents `id()` reuse outright), persisted on the `Drawing` as `_ann_box_cache` beside the #143 `_view_edge_cache` and cleared at the same finalize-rollback site. A replaced annotation re-measures automatically. `lint_drawing` gains an optional `ann_box_cache=` with the same persistence contract as `view_edge_cache` — omitted, a fresh per-call cache still de-duplicates within the call.

## Numbers

| | before | after |
|---|---|---|
| `Drawing.lint()` (scattered plate) | 0.85–0.94 s | **0.06–0.08 s** |

The build's own lint→repair loop pre-warms the cache, so the export-time lint is ~free (the sole remaining OCC bbox on re-lint is `lint_location_coverage`'s single `part.bounding_box()` — the 3D solid, not annotation geometry).

## ADR fit

- **ADR 0002** — the lint→repair loop is the safety net that re-lints repeatedly; this is precisely the caller the #143 view-edge cache was built for, extended to the annotation boxes that dominated once view edges were cached.
- **ADR 0007** — draftwright owns linting; the change is contained in `linting/structural.py` + the `Drawing.lint` wiring. No layering change (drawing → linting is already downward).
- Semantics preserved: the memo returns exactly what each site computed before; failure paths (`None`) skip checks the same way the old per-site `except` blocks did.

## Gates

- Golden corpus **14/14 unchanged**
- New `tests/test_lint_box_cache.py`: memo + identity-check unit tests, persistent-cache-across-lints guard, and an integration guard that a re-lint recomputes ≤1 bounding box
- Lint suites (structural/linting/reconciliation + new) 65 passed; smoke 47 passed; ruff check / format / mypy (51 files) / codespell clean

### DIFF
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -448,6 +448,8 @@ class Drawing:
         # Same persistence idea for annotation bounding boxes (#602): an optimal
         # bbox on fused dimension geometry costs ~10 ms and lint needs one per
         # item; entries are identity-checked so replaced annotations re-measure.
+        # Contract: items are replaced, never mutated in place, once on the
+        # sheet (see _ann_box); lint() prunes entries for departed objects.
         self._ann_box_cache: dict = {}
         self.svg_path: str | None = None
         self.dxf_path: str | None = None
@@ -2144,6 +2146,12 @@ class Drawing:
         # longer relies on the helpers set_page module-global (ADR 0007).
         page_bbox = (_MARGIN, _MARGIN, self.page_w - _MARGIN, self.page_h - _MARGIN)
         view_shapes = [vis for vis, _ in self.views.values()]
+        # Prune box-cache entries for objects no longer on the sheet, so replaced
+        # annotations (repair's _replace_dim swaps in a fresh object) don't keep
+        # their OCC geometry strongly referenced for the drawing's lifetime.
+        live = {id(i) for i in self.items} | {id(v) for v in view_shapes}
+        for stale in [k for k in self._ann_box_cache if k not in live]:
+            del self._ann_box_cache[stale]
         # Most annotations are at sheet scale, but a non-sheet-scale view (the
         # enlarged detail view, #42) tags its dims with `_dw_scale`. Lint each
         # scale group with its own drawing_scale so label-vs-measured is correct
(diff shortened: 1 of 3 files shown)


pzfreo:
**Codex adversarial review — 3 rounds, disposition:**

- **R1** (1 High, 2 Medium): memory retention **accepted** → `lint()` now prunes departed objects (endorsed by R2 as correct incl. the repair-rollback interaction); negative caching **rejected** (deterministic failure for unchanged geometry — R2 endorsed the rejection); in-place mutation staleness initially rejected on the replace-never-mutate discipline.
- **R2** held the mutation High: `Drawing.items` exposes live shapes and `Shape.locate()` mutates in place → **fixed** with a location token on each cache entry (identity AND token must match; ~14 µs/read, lint stays ~0.1 s). R3 confirms this closes it.
- **R3** remaining Medium — geometry mutated while `.location` is preserved, or mutable location-less stand-ins persisted across lints — **accepted as a documented limit** (`f935db3`): no engine path does it (the whole editing surface replaces), stand-ins are a per-call test affordance never persisted on a `Drawing`, and detection would require topology hashing that costs as much as the measure the memo avoids. Same exposure has existed in the #143 view-edge cache throughout its life. R3 endorsed everything else: no missed call sites, sound token comparison, negligible overhead.

Merging on green per the #602 loop authorization (judgment gate).

PR #683, opened 2026-07-17
perf(#602): edge→faces adjacency map for the fillet neighbour search

pzfreo:
Part of #602. Follows #678/#679/#680/#681/#682 — found by the CTC-02 split profile.

## What

`recognise_fillets` found each cylindrical candidate's bridged neighbour faces by testing **every face** with `any(a.IsSame(b) …)` over both faces' full edge lists — O(faces² × edges²). On NIST CTC-02: **3.7 million `TopoDS.IsSame` calls**, ~6 s of the detector's 10.8 s (the most expensive detector in the build, run once per build since #682).

Now an edge→faces adjacency map is built once per call, keyed by build123d `Edge` — whose `__hash__`/`__eq__` are exactly `IsSame` semantics (same TShape + Location, orientation-insensitive; verified empirically on a shared reversed edge and per build123d's `__eq__` docstring) — i.e. the *identical* predicate the sweep used. Each candidate then walks only faces actually sharing one of its edges, deduped.

## Correctness

- The per-axis closest-coordinate preference is a min — order-independent — so visiting neighbours in edge order instead of face order cannot change the result.
- **Byte-identical output verified**: all 21 CTC-02 fillet records (axis, radius, anchor point) JSON-diffed old vs new — no difference.
- Golden corpus **14/14 unchanged**; recognition + recogniser-contract + detect-once suites (79 tests) pass.

## Numbers

CTC-02 `recognise_fillets`: **10.8 s → 4.6 s** (3 reps each). Remaining detector cost is the per-candidate OCC work (surface adaptors, solid classification) — linear and not pathological.

## Guard

`tests/test_fillets_adjacency.py`: runs the detector under cProfile and bounds the `IsSame` count (map form ≈78 vs sweep ≈752 on a 13-face part; the bound sits between with margin, includes a >0 sanity check so the metric can't silently vanish, and **fails on the old implementation** — verified).

## ADR fit

ADR 0013 recogniser contract untouched (same signature, same deterministic `list[Fillet]`); `recognition/` stays at the DAG bottom (no new imports). This is an internal algorithm change in one recogniser.

### DIFF
--- a/src/draftwright/recognition/fillets.py
+++ b/src/draftwright/recognition/fillets.py
@@ -82,7 +82,15 @@ def recognise_fillets(
     bb = part.bounding_box()
     max_ext = max(bb.max.X - bb.min.X, bb.max.Y - bb.min.Y, bb.max.Z - bb.min.Z)
     all_faces = list(part.faces())
-    face_edges = [(g, [e.wrapped for e in g.edges()]) for g in all_faces]
+    # Edge→faces adjacency, built once. build123d shape equality/hash is IsSame
+    # (same TShape + Location, orientation-insensitive) — the identical predicate
+    # the old per-pair scan used — so one dict pass replaces the
+    # O(faces² × edges²) IsSame sweep (#602: 3.7M IsSame calls, ~6 s of CTC-02's
+    # 10.8 s fillet detection).
+    edge_faces: dict = {}
+    for g in all_faces:
+        for ge in g.edges():
+            edge_faces.setdefault(ge, []).append(g)
 
     out: list[Fillet] = []
     for f in all_faces:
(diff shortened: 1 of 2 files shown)


pzfreo:
**Codex adversarial review — 2 rounds, disposition:**

- **R1**: adjacency mechanics fully endorsed (hash/eq = IsSame contract, collision handling, instanced-TShape behaviour, wrapper stability, id-dedup equivalence). Medium (traversal-order-dependent tie on exact-distance neighbour ties) **fixed** in `c01d65a` with the order-independent `(distance, coord)` tie-break; CTC-02 output re-verified byte-identical. Low (cProfile IsSame tripwire fragility) **accepted as documented** — the >0 sanity check fails loudly if the metric vanishes, and instrumenting pybind comparison internals isn't feasible.
- **R2**: the tie-break delta endorsed as clean ("deterministic total ordering… removes the traversal-order dependency"). Its sole High — an unterminated docstring in `_axis_aligned_axis` — was a transcription artifact of the review prompt embedding, per its own stated caveat: verified `ast.parse` + import of the real file succeed, and the closing `"""` is present on line 62. No remaining findings.

Merging on green per the #602 loop authorization.

PR #684, opened 2026-07-17
deps: build123d-drafting-helpers 0.14 — 9× faster dimensions, honest tight-span rendering

pzfreo:
Upgrades helpers to v0.14.0 (the #602-driven boolean-free ink release, helpers#177) and adapts draftwright to its honest rendering of tight-span dimensions.

## Numbers (4-core dev box)

| metric | 0.13.1 | 0.14.0 |
|---|---|---|
| per-`Dimension` construction | 234 ms | **25 ms** |
| grid plate build | 7.4 s | **2.8 s** |
| scattered plate build | 10.3 s | **3.6 s** |
| full fast tier | ~14 min | **~3.5–5 min** |

Combined with the merged #602 loop: grid plate 40 s → 2.8 s in one day.

## Why this is more than a version bump

v0.14 renders tight-span dims **correctly** (ISO outside arrows) where 0.13 silently degraded to a bare-line fallback (the "empty wire" band, fixed upstream). Honest, larger ink hulls + honest `label_bbox` metadata exposed four draftwright assumptions, each fixed at the root:

1. **`dim_footprint` models the flip** — outside-arrow tails (2·al past each end) and the witness keep-clear cut, mirroring `_dim_line_ink` exactly. 11/11 calibration cases ±0.15 mm (3 new tight-span cases).
2. **Diameter-row label spacing measures real text width** (`_text_size`) instead of a per-char estimate the honest label boxes outgrew (hypothesis tier caught ø-label overlaps).
3. **Leader shelf-flip collision repair**: helpers' Leader hangs its label leftward when the tip sits right of the elbow; a crowd-shifted row elbow could land a flipped label on its neighbour (seed-dependent hypothesis finding). Direction-aware interval check + a left-to-right no-flip cascade, engaging only on real collisions — lone edge flips (boss tolerance leaders) untouched.
4. **Plate-thickness corner conflict**: a tight-span thickness dim's witness hull overlaps the below strip's at the view corner at *every* position (AABB artifact — inks disjoint). Fix: opposite-strip → side-view fallthrough (the allowlisted GD&T on_drop pattern; carve allowlist + ADR 0009 note updated). Also adds corner reserves in `drain_corridors` (sibling force-candidate footprints as carve obstacles) — sound but insufficient alone for witness-anchored hulls; kept as groundwork for L-shaped occupancy (follow-up issue to be filed).

## Output changes (deliberate)

- 2 of 14 goldens re-blessed: one annotation each, ink-extent only (the outside-arrows fix).
[...]

Issue #636, opened 2026-07-15
Consolidation P1: finish ADR 0009 — migrate remaining carve_free_position render passes into the corridor solve


pzfreo:
Part of #635 (Phase 1).

## Problem

ADR 0009's guarantee — the invisible-occupant collision class removed *by construction* — only holds when every strip occupant is a candidate in the solve. Amendment 8 (#524) consolidated PMI, front-view callouts, and the pitch fallback, but the auto-pass still commits geometry via the single-position carve (`carve_free_position`, `annotations/_common.py:737` — the exact "clears already-placed annotations but is not a candidate in any solve" poke ADR 0012 flags). Remaining production callers:

| Pass | Sites |
|---|---|
| `render_plates` (#559, **new in 0.3.0**) | `from_model.py:1269` |
| `render_height_ladder` | `from_model.py:1632, 1674, 1708` |
| `render_step_positions` (#555, **new in 0.3.0**) | `from_model.py:1768` |
| PMI alternate-strip drop fallback | `from_model.py:2632` |
| holes `_rim_tip` | `holes.py:122, 139` |
| scattered-hole location `_place` | `holes.py:248` |
| pitch-dim bounded-vector fallback | `holes.py:1062` — *accepted by Amendment 8 as a probe, may stay; decide explicitly* |

`Drawing.place_dim` (`drawing.py:776`) is the documented ADR 0012 escape hatch and stays.

That two **new 0.3.0 features** landed on the carve path shows the migration isn't just unfinished — new work still defaults onto the legacy path.

## Task

- Convert each pass above to register `CorridorCandidate`s / strip candidates drained by the shared solve (the Amendment 8 pattern: authored-intent priority, drop callback for fallback sides).
- For any site that legitimately cannot be a solve candidate (e.g. the pitch diagonal probe), record the exemption in ADR 0009 explicitly rather than leaving it implicit.
- Add a guard test: no `carve_free_position` callers in `annotations/` outside an explicit allowlist (same fail-closed style as `test_import_boundaries.py`).

## Acceptance

- `grep carve_free_position src/draftwright/annotations/` → definition + allowlisted exemptions only, and the allowlist is asserted in a test.
- Existing placement gates stay green (`test_strip_layout`, `test_layout_property`, `test_layout_snapshot`, `test_e2e_standards`).
- ADR 0009's Current-decision header updated to drop the "remaining migration" note.

## Related

[...]

PR #688, opened 2026-07-18
feat(#685): L-shaped occupancy — decomposed obstacle boxes via segments

### LINKED ISSUE #685: layout: L-shaped occupancy — represent a dim as witness-corridor + dim-line boxes, not one hull
Follow-up from #602 / PR #684 (helpers 0.14). ADR 0009's occupancy model represents every annotation as one AABB. A dimension's ink is L-shaped (witness corridor + dim-line band), so its hull includes large empty corner regions — and helpers 0.14's honest tight-span rendering (outside arrows) made those hulls big enough to collide **structurally**:

- Two perpendicular strips' dims overlap hulls at the view corner at *every* position (the below dim's hull always reaches the view bottom via its witnesses; the left dim's always reaches the view edge) — even when their inks are provably disjoint. Verified on the declared L-bracket (PR #684 investigation) and the dshape ratchet pair (`dim_height`/`dim_loc_front_z400`, documented BENIGN).
- The corner-reserve mechanism added in #684 (`drain_corridors` passes sibling force-candidate footprints as carve obstacles) is sound but cannot help witness-anchored hulls — landed as groundwork.
- The plate-thickness opposite-strip/side-view fallthrough (#684) sidesteps the corner for that pass; joint-solving multiple deferred retries was explicitly deferred here (Codex #684 r2 Medium 2).

## Proposal

Represent a dim's occupancy as **two boxes** — the witness corridor (feature → dim line) and the dim-line band (line + arrows/tails + label) — in:
- `strip_obstacles` / the carve's obstacle projection,
- `dim_footprint` (return both boxes; single-hull callers can union),
- the corridor solve's predicted-box evaluation + accept-time validation,
- the cleanliness ratchet's overlap definition (burn down the dshape BENIGN entry when this lands).

By construction this eliminates the corner-contention class everywhere: witness corridors may cross other witness corridors (ISO 129-1), dim-line bands may not overlap anything.

## Acceptance

- The dshape ratchet entry and the L-bracket fallthrough become unnecessary (plate dim places on its natural left strip again).
- Goldens: expect deliberate placement improvements (tighter packing); review-bless with reasoning.
- Count-gated: no new OCC builds (the second box is analytical).

pzfreo:
Closes #685. The architectural fix for the witness-hull corner-collision class that helpers 0.14's honest rendering exposed (#684).

## What

`occupancy_boxes(o)` decomposes any annotation exposing helpers ≥0.14 `.segments` into **per-stroke boxes** — each inflated by a preset-aware pad, `max(1.2 mm, draft.arrow_length/2)`, which bounds the arrowhead's half-width and its protrusion past an inside-arrow shaft trim — plus its `label_bbox`; anything else keeps its hull. `strip_obstacles` emits the decomposition, and the carve's existing perpendicular-band filter then frees sibling-strip corners **with no solver changes**.

The cleanliness ratchet mirrors the decomposition with an honest exemption: only **transverse** stroke crossings (≥30° between stroke directions — an outer dim's witness passing through inner tiers, ISO 129-1) are benign; near-parallel stroke overlap and any label/hull involvement flag.

## Results (honest accounting, post-review)

- **The declared L-bracket's plate-thickness dim places on its natural left strip again** — the #684 opposite-strip fallthrough no longer triggers (kept as a dormant safety net).
- **The corner-artifact class is gone for real** (pairs whose hulls collided at view corners while their inks stayed disjoint).
- The **parallel witness-sharing class remains and is now explicitly documented**: datum-chain location dims and stacked step/slot ladders share their extension-line corridor — standard ISO running-dimension presentation, allowlisted per fixture with rationale. Distinguishing shared witness corridors from genuine collinear overprint needs semantic stroke roles helpers doesn't expose (possible follow-up). Two ladder-class pairs (slotted, turned_shaft) are newly *visible* and documented rather than hull-hidden.
- dshape's `dim_height` pair reclassified **PENDING #636** (real stroke-through-label; the height-ladder carve is invisible to the solve).
- 4 goldens deliberately re-blessed twice (decomposition, then the wider preset-aware pad).

## Review trail (4 rounds)

R1: both Highs fixed (preset-scaled pad; transverse-only exemption). R2: production fixes endorsed; three test-side findings applied verbatim. R3: caught that the transverse predicate had silently no-opped (an unasserted replace) — wired for real, allowlist re-triaged honestly. The oracle is now **stricter than pre-#685** while the occupancy model is more precise.

## Tests

New `test_occupancy_boxes.py`; the three `strip_obstacles` unit tests updated to intent-preserving decomposed assertions (segment-endpoint sampled). Full fast tier green; ruff/format/mypy/codespell clean.

## ADR fit

The inner ADR 0009 solve and its band filter are untouched — the change is confined to the occupancy *representation* (`_common.py`) and the ratchet's measure. ADR 0004 outer blocks unaffected. No new OCC builds (segments + label_bbox are metadata).


pzfreo:
**Codex adversarial review — 4 rounds, disposition:**

- **R1** (2 High, 2 Medium, 1 Low): fixed pad → preset-aware `max(1.2, arrow_length/2)` (bounds head half-width + inside-arrow protrusion); blanket stroke exemption → transverse-only; leader test → endpoint-sampled. Ink-coverage Mediums answered by the pad fix + helpers 0.14's documented segments/label_bbox coverage.
- **R2**: production fixes endorsed; three test-side findings applied verbatim (oracle pad mirrors production, endpoint negative assertion, inclusive boundary).
- **R3**: the round of the PR — caught that `_benign_crossing` was defined but never called (an unasserted text replace had silently no-opped), leaving the blanket exemption live. Wired for real; the parallel witness-sharing class resurfaced and was restored as documented BENIGN entries (ISO running dimensions); the PR narrative corrected to the honest accounting.
- **R4**: "No findings. The delta is clean." — predicate verified wired, exemption narrow, resurfaced cases fixture-accounted, narrative matches behaviour.

Full tier 1411 green ×3 through the review. Merging on green per the standing loop authorization.

Issue #635, opened 2026-07-15
Epic: consolidation pass — pay down leaf-level debt before the next feature wave


pzfreo:
## Why

A full project review (2026-07-15) found the architecture sound — one engine under all surfaces, a machine-enforced IR waist, disciplined ADRs — but **accreting faster at the leaves than it consolidates**. The dominant failure mode is regression-driven patching: fixes land as new branches/copies citing the issue they fixed, without the global model being re-derived. Evidence of the trend:

- The identical label-clear guard is copy-pasted **five times** across `render_chamfers`/`render_fillets`/`render_flats`/`render_pockets`/`render_grooves` (`from_model.py:912/978/1042/1146/1204`) — each 0.3.0 feature kind landed as a fresh copy.
- Two 0.3.0 features (**plates #559, step-positions #555**) landed on the legacy `carve_free_position` path rather than the ADR 0009 corridor solve.
- Magic constants are justified by bug numbers, not invariants (`0.62` text-width fudge ×3 while `_text_width()` exists; two dedup bases that a comment admits can disagree and resurrect #345).
- Comments cross-reference line numbers that have drifted (`holes.py` "line 638" → now ~1279).

None of this is runaway; all of it steepens the cost curve of the *next* feature. This epic is the consolidation release: **no new feature kinds until Phases 1–2 land**, then the remainder interleaved.

## Principle going forward

A change is not done if it (a) places geometry the corridor solve cannot see, or (b) adds a copy of an existing pass instead of extending a shared one. Same spirit as the #574 round-trip rule.

## Phases

- [ ] **Phase 1** — finish ADR 0009: migrate the remaining `carve_free_position` render passes into the corridor solve
- [ ] **Phase 2** — deduplicate the five leader-callout renderers; name the magic constants
- [ ] **Phase 3** — split the three complexity hotspots (`_annotate_holes`, `render_pmi`, `finalize`/`_classify_intents`)
- [ ] **Phase 4** — ADR 0005 §2 at last: an explicit `PlacementContext`; retire `Drawing` as the build-state bus
- [ ] **Phase 5** — DAG enforcement: extend `test_import_boundaries.py` beyond `model/`; keep CLAUDE.md/ADRs matching the territory
- [ ] **Phase 6** — test-suite hardening: specify intended behaviour, not current behaviour

## Ordering

Phases 1 and 2 first (they stop the accretion); 3 is independent; 4 before 6 (the declared seam is what the white-box tests re-target); 5 any time.

[...]

Issue #237, opened 2026-06-29
converge: prismatic-step ladder + envelope height + OD (right-strip group)


pzfreo:
Design epic for ADR 0008 convergence (see [roadmap](../blob/main/docs/plans/0008-convergence-roadmap.md)). These engine passes are **coupled** through the shared `fv_zones.right` ladder / `_right_ladder` cursor, so they migrate **together**, not one at a time:

- prismatic **step-height ladder** — `dim_step_*`, `_detect_step_repeat`, `_legible_steps` (orchestrator inline)
- envelope **height** — `dim_height` (shares the right-strip cursor with the ladder)
- **OD** — `dim_od` (rotational; above front)

**Prereq (new IR modelling):**
- a **prismatic-step `Feature`** — detection half-exists in `analyse_face_levels` → `step_zs`; emit Features from it in `build_part_model` (the IR currently models only *turned* steps).
- rotational-OD modelling (which `BossFeature`/diameter is the OD; needs the rotational classification the IR doesn't yet carry).

**Approach:** zone-aware renderer (like `render_envelope`) allocating from `fv_zones.right`, preserving the leapfrog cursor; folds in #230 (`N× rise` staircase collapse) and #222 (OD on the profile view). Migrate-and-delete the inline blocks. Judge by correctness + visual parity.

pzfreo:
Note: `PartModel.orientation` (turning axis / None) **already carries the rotational classification**, so the OD prereq is small — no new is_rotational modelling needed. Per Amendment 4, the section/zone machinery is shared infra the renderer feeds.

pzfreo:
Done (PR #280). Inline OD/centreline/bore-leader + prismatic step-ladder + overall-height (~210 lines, coupled via the fv_zones.right cursor) → two IR feature types (`StepLevelFeature`, `RotationalFeature`) + two zone-aware renderers (`render_height_ladder`, `render_rotational`), inline deleted. Behaviour-equivalent; full suite green. Deferred enhancements: #230 (turned N×-rise), #222 (OD on profile). Surfaced + filed #279 (phantom ø0 step diameter, pre-existing).

PR #689, opened 2026-07-18
feat(#636): height ladder joins the corridor solve — the last solver-invisible auto-pass

### LINKED ISSUE #636: Consolidation P1: finish ADR 0009 — migrate remaining carve_free_position render passes into the corridor solve

pzfreo:
Closes #636 (consolidation epic #635, P1). With this, ADR 0009's by-construction guarantee covers **every auto-pass strip occupant**.

## The migration

`render_height_ladder`'s step dims + overall height register as force `CorridorCandidate`s in the shared `(front, right)` corridor:
- The **leapfrog witness cursor** (#237) survives as a *build-time chain*: candidates share a solved-position map and each dim's witness anchors on its nearest already-built predecessor's line (builds follow candidate order — deterministic). Prediction uses an analytical footprint with the witness modelled from the view edge (conservative; accept-time validation trims drift).
- `dim_height` rides the OVERALL subchain → **outermost by construction** (the envelope pattern), replacing the carve's `outermost=True`. It packs 6.4 mm tighter on both prismatic goldens (sole golden change, re-blessed).
- `render_gdt`'s alt-strip fallback now defers via `ctx.post_drain` (the #684 plate pattern) — no carve can preempt an undrained sibling's corner.
- Allowlist shrinks by one; the manual verbs (`add_feature_callout`/`add_feature_location`) are recorded as an explicit ADR 0009 exemption (the #426 half — one annotation onto a finished sheet has no drain to join); `_place_pitch_dim` stays permanent (Amendment 8). ADR 0009's migration note updated to the close-out state.

## Three downstream consequences, each resolved on documented intent (not carve artifacts)

1. **Finalize drain gate**: `height_ladder_ids` joins the B2 drain gate + post-drain intent drop (the #646 step-position pattern) — A0 now only registers, so a ladder-only deferred edit still drains, and a pre-drain raise leaves intents recorded for a clean retry.
2. **Detail views**: the solve now *places* `dim_height` where the carve silently dropped it ("strip full") — but on crowded parts that room was the explicitly-requested detail view's. `_resolve_details` now demotes `dim_height` deliberately (logged) and retries once: same output as before, as an explicit decision honouring the user's `detail_view=True`.
3. **CTC-01 emit routing**: the Y-drilled hole's Z location lands on the **front** right ladder — `_locate_along_z`'s documented primary for a Y-axis hole (its circle shows in the front view) — co-solved with step/height dims as one running ladder. The old side placement was a crowding artifact of the pre-placed carve ladder; the test is updated deliberately with the reasoning inline.

Known residual (documented in ADR 0009 + the dshape ratchet entry): a witness crossing a fixed-height label is a perpendicular-axis conflict tier co-solving cannot fix — remedy is label-offset shifting for witness crossings (follow-up issue to be filed at merge).

## Gates

Full fast tier **1410+ passed / encapsulation ratchet respected** (the demotion reads public `annotations()`); carve-caller guard green with the shrunk allowlist; 2 goldens re-blessed (`dim_height` 6.4 mm tighter, sole field change); ruff/format/mypy/codespell clean.


pzfreo:
**Codex adversarial review — 2 rounds, disposition:**

- **R1**: endorsed the solved-chain lifecycle, the finalize transactional boundary, the gdt deferral, and the emit-test reasoning ("the previous side fallback was not itself a presentation guarantee"). Two findings, both applied:
  - High (a failed detail retry permanently deleted `dim_height` — net loss of both) → the demotion is now **transactional**: capture object+view, remove, retry, restore on failure.
  - Medium (the edge-anchored footprint could falsely exhaust the strip when an inner obstacle sat in the traversed witness region) → `_foot` now reads the **same solved-position map** as the build chain (first evaluation unchanged; later-segment retries tight; degenerate-offset guard).
- **R2**: "PR #689 is clean at `cea7b1c`. No actionable findings."

Full fast tier 1411 green ×2 through the review. Merging on green per the standing loop authorization; #636 closes with this.

Issue #692, opened 2026-07-18
test: layout-hypothesis strategy consumes ambient randomness — falsifying examples don't replay


pzfreo:
tests/test_layout_hypothesis.py::test_generated_part_has_no_layout_collisions fails on a seed-dependent minority of runs (observed repeatedly during #684/#691 work; reproduces identically on main, so unrelated to those changes) with an annotation_overlap lint hit. The deeper defect: the failure does NOT replay stably from Hypothesis's example database — the part-generation strategy appears to mix pytest-randomly's per-session seed into the geometry, so a cached falsifying example produces a different part on the next run. That breaks Hypothesis's core shrink-and-replay contract and makes every such failure a fresh hunt (during the 0.14 upgrade, pinning one required ~20 instrumented runs; it exposed the real diameter-row shelf-flip bug, so the tier has value — when it can be replayed).

Fix: make the strategies purely Hypothesis-driven (draw every dimension from st.floats/st.integers; no random module / ambient seed), then let the DB replay carry failures across runs. Then re-hunt the currently-known flake (an annotation_overlap between two ø labels) with a stable repro.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #693, opened 2026-07-18
feat(#690): shift a label clear of a foreign crossing witness

### LINKED ISSUE #690: layout: shift a label along its dim line when a witness corridor must cross it
Follow-up from #636 / PR #689 (and the #685 honest-oracle work). The last dshape ratchet pair (`dim_height` × `dim_loc_front_z400`) is a **perpendicular-axis conflict tier co-solving cannot fix**: a location dim's witness must cross the entire right strip to reach its tier, and the height dim's label sits at a fixed height on that path. Any label in an inner tier at that height gets crossed, wherever the tiers land.

## Proposal

Generalise the #129 centreline machinery (`clear_label_of_centerlines` → `label_offset_x`): when the corridor solve accepts a candidate whose **witness path crosses an already-placed (or co-solved) label**, shift that label along its own dim line to clear the crossing — the same remedy a human draughtsman applies, and the same mechanism lint's `label_centerline_overlap` suggestion already recommends.

Sketch:
- At drain time (post-solve, pre-build — positions known), compute witness segments per accepted candidate analytically; test against co-solved siblings' label boxes (`dim_footprint` gives both).
- On a crossing, compute the minimal `label_offset_x` that clears (the #129 interval-carve logic), pass it into the build.
- helpers ≥0.14 renders shifted labels correctly (uniform keep-clear, witness gap cut at the real label position — its v0.14 fix list).

## Acceptance

- The dshape `_KNOWN_OVERLAPS` entry burns down (the ratchet forces this — its comment already points here).
- No label ends up crossed by a witness on the cleanliness corpus; goldens re-blessed with reasoning where labels shift.
- Count-gated: offsets computed analytically, no extra OCC builds.

pzfreo:
Closes #690. The perpendicular-axis conflict class tier co-solving cannot fix: a location dim's witness must cross the whole strip to reach its tier, so any inner-tier label at that height is crossed wherever the tiers land (the dshape `dim_height` case, PENDING since #685).

## What

`reconcile_witness_labels` (repair.py — the ADR 0002 machinery as a **deterministic late pipeline pass**, called by both build paths after every corridor drains and defers): for each spec-carrying dim whose label a **foreign transverse stroke** crosses — a zero-thickness-line test (constant coord strictly inside the label band + >0.5 mm overlap on the other axis; parallel strokes are the stacked-shaft pattern, exempt) — compute the minimal clearing `label_offset_x` along the dim's own line (interval carve over its span, clamped to it) and rebuild once via `_replace_dim`. Unshiftable labels stay put (unchanged output; lint reports as before).

Design note: an in-solve seam was prototyped and **dropped** — tracing showed the conflict pair is never co-solved (the dims arrive via different placement calls), so only a post-drain pass sees both; one mechanism at the right altitude.

## Results

- The dshape witness-through-label defect is **fixed** (label shifts ~2.4 mm clear; verified by the new regression test, which asserts the no-crossing property across the fixture and reconciliation idempotence).
- The dshape ratchet pair reclassifies honestly: what remains is the **shared-datum witness** (both dims anchor the same base edge — the BENIGN running-dimension corridor, same as plate_holes).
- Goldens **unchanged** (no golden-corpus part had a crossing).

## Gates

Full fast tier 1411 passed / 1 failed — the failure is the pre-existing #692 hypothesis-replay flake (reproduces identically on main). Ruff/format/mypy/codespell clean; DAG intact (repair imports only `_core`; the reconciler inlines its ten-line interval carve rather than importing the corridor's).

### DIFF
--- a/src/draftwright/annotations/orchestrator.py
+++ b/src/draftwright/annotations/orchestrator.py
@@ -72,6 +72,7 @@ from draftwright.model import (
     plan_dimensions,
     plan_sections,
 )
+from draftwright.repair import reconcile_witness_labels
 
 
 def _wrap_rows(header, data, ncols):
@@ -356,6 +357,11 @@ def _auto_annotate(dwg, a: Analysis, *, detail_view: bool = False):
     # (ADR 0009 end state, #345/#346/#393) — dedup coincident spans, order the ladder —
     # BEFORE the section/detail views so they see the placed ladder as an obstacle.
     drain_corridors(ctx, dwg)
+    # (#690) Witness-crossing label reconciliation — after every corridor (and its
+    # deferred fallthroughs) has placed: shift any dim label a FOREIGN transverse
+    # stroke crosses, along its own line, via the repair machinery. Deterministic,
+    # runs in both build paths.
+    reconcile_witness_labels(dwg)
 
     # Turned/circlip-groove callouts (#148c): {width} WIDE × ø{dia} via a leader off each
     # groove. A groove is a secondary leader-callout on a turned shaft — exactly where the
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -1560,6 +1560,7 @@ class Drawing:
             feature_hole_keys,
         )
         from draftwright.model import PartModel, plan_dimensions
+        from draftwright.repair import reconcile_witness_labels
 
         _section = r.section
         corridor_ids, callout_ids, dia_ids = r.corridor_ids, r.callout_ids, r.dia_ids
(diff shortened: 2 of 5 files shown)


pzfreo:
**Codex adversarial review — 3 rounds (all open rein), disposition:** R1: three findings, all applied — whole-span destination blocking (a shift can no longer land on another witness), diagonal strokes skipped as out-of-class, pinned dims exempt via the `_find_dim` registry contract. R2: R1 fixes certified; one Medium (diagonal *target* dims under the axis-aligned solve) — fixed with its exact remedy (skip at the stroke rule's tolerance; the diagonal pitch fallback has its own clearance search). R3: **'PR #693 is clean at `9d94f5c`… No new findings.'** Full tier green ×3 modulo the pre-existing #692 hypothesis flake (reproduces on main). Merging on green per the standing loop; closes #690.

PR #718, opened 2026-07-18
fix: scope the label_bbox warning memo to one lint run

pzfreo:
The Codex post-merge sweep of today's eight audit PRs (ddbca39..9dfcfee) returned one finding, fixed here: `_warned_label_bbox` (introduced by the #711 review fix for the O(n²) warning flood) was a process-global set of `id(item)` values that never cleared — a freed id reused by an unrelated later object would have its warning silently suppressed, and the set grew without bound in long-lived processes.

Fix: clear the memo at `lint_drawing` entry. Within one lint run the ids are stable (the run holds the items alive), so the flood dedup is unchanged; across runs nothing stale survives.

Verification: ruff/format/mypy/codespell clean; `test_lint_structural.py` + `test_lint_box_cache.py` + `test_lint_reconciliation.py` + `test_linting.py` — 71 passed.

### DIFF
--- a/src/draftwright/linting/structural.py
+++ b/src/draftwright/linting/structural.py
@@ -115,7 +115,10 @@ def _item_label(item) -> str:
 
 # Items whose raising label_bbox was already warned about — several checks read the
 # same item's label_bbox (per centreline pair, per view), so an unmemoised warning
-# would flood the log O(n²) on one bad item (#711 review).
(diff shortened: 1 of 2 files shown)


pzfreo:
Codex review addressed: the memo is now a genuinely run-local set threaded through the check helpers (no module global at all), plus a test pinning once-per-run (not once-per-process) semantics.

Issue #731, opened 2026-07-18
#698 child: derive the step group view instead of hardcoding it


pzfreo:
Child of epic #698. `planner._group_view` hardcodes `if feature.kind == "step": return "front"` under a docstring claiming views are derived, never hardcoded. Derive it from the feature frame like every other kind (or document why front is genuinely invariant for turned steps).

Issue #710, opened 2026-07-18
Fuzz-found: packed view overflows A3 drawable area by 0.8 mm (view_out_of_bounds on a generated part)


pzfreo:
`test_layout_hypothesis.py::test_generated_part_has_no_layout_collisions` found a generated part whose finished drawing carries `view_out_of_bounds`: a right-column view (bbox x=272.90–410.78, y=88.03–183.97 page-mm, i.e. ~137.9×95.9) extends **0.8 mm past the right drawable edge** of an A3 landscape page (drawable (10, 10, 410, 287)). Two distinct view items report the identical bbox (duplicated message).

This is a compose-then-pack near-miss (ADR 2 (was 0004)): the (scale, page) fitness accepted the layout but the real built geometry overhangs by under a millimetre — the box-layout footprint the packer used was slightly narrower than the final view + annotation ink, and `_repack_to_fixed_point` did not (or could not) pull it back.

**Repro:** stochastic — found 2026-07-18 during a local full-tier run (the failing example lives in the local `.hypothesis` DB and replays deterministically there; CI draws fresh examples, so PR CI is unlikely to hit it). Debug capture came from temporarily printing in the `view_out_of_bounds` check while replaying.

Note: unrelated to the #701 lint changes — the same example fails on `main`.

Worth checking:
- which footprint under-measures by ~1 mm at pack time (iso right-strip? side-view annotation boxes?),
- why two view items share one bbox (double-registered view shape → duplicated lint messages),
- whether the packer needs a small safety pad at the drawable edge.

Issue #707, opened 2026-07-18
Emitted Sheet does not faithfully reproduce direct CLI drawing


pzfreo:
## Bug

A drawing produced by the normal Draftwright CLI path is materially different from the drawing produced by executing the declarative `Sheet` script emitted for the same STEP file.

This was observed while integrating Draftwright `0.3.3` into Maquetto. Maquetto previously generated the editable Sheet with `generate_sheet_script(...)` and executed that script for its first PDF. The result did not appear equivalent to using the CLI directly to create the PDF.

## Expected invariant

For the same STEP input and the same title-block/layout options, these paths should be visually and semantically equivalent:

1. Direct CLI/default engine path:

```bash
draftwright part.step --format pdf
```

Equivalent API path: `draftwright.builder.build_drawing(...)` followed by `Drawing.export(formats=("pdf",))`.

2. Declarative emitter path:

```bash
draftwright part.step --script --style sheet
python part.py
```

Equivalent API path: `generate_sheet_script(...)`, then execute the emitted `Sheet` script and build/export it.

The emitted script is documented as reproducing the detected drawing faithfully; it should retain the direct build’s detected features, view selection/orientation, annotations, dimensions, title-block settings, page/scale choice, and materially equivalent layout.

## Impact

Consumers that expose an editable Sheet cannot use the emitted script as their initial drawing artifact without diverging from Draftwright CLI output. Maquetto has to use `build_drawing(...)` for the first PDF/PNG and retain the emitted Sheet only for later edits, creating two possible rendering paths.

## Suggested regression

Add an emitter-fidelity fixture that:

- runs a representative STEP through `build_drawing(...)`;
- emits and executes its declarative Sheet;
- compares the resulting detected/declared feature census, views, annotation/dimension inventory, page and scale, lint result, and rendered output;
- fails on missing/repositioned drawing content beyond documented deterministic tolerances.

## Environment

- Draftwright: `0.3.3`
- build123d-drafting-helpers: `0.14.0`
- Python: `3.12`

## Fixture status

The affected STEP file has not yet been attached to this issue. It should be reduced or added as a private/public regression fixture before attributing the mismatch to a specific recogniser or layout pass.

pzfreo:
Draft PR #743 adds a temporarily disabled direct/generated-script parity characterization suite.

[...]

Issue #647, opened 2026-07-15
finalize() retry after a mid-drain exception is not transactional


pzfreo:
Split out of PR #646 (#636 review, Codex `gpt-5.6-sol` xhigh). **Pre-existing** — not introduced by the plate/step corridor migration; the same applies to `render_locations`/`render_slots`.

## Problem

`Drawing.finalize()` is documented as resilient (a raised replay leaves the rest recorded, so a corrected retry can re-run). But the corridor-drain path is not transactional across a retry:

1. **Intents replay and duplicate committed corridors.** B2 removes the location/slot/user-dim intents only *after* `drain_corridors()` returns. If a later corridor's `solve_corridor` raises, those intents remain; a retry re-runs `render_locations`/`render_slots`, recreating already-committed corridors and appending duplicate candidates. Location retries allocate fresh first-free names (`m_locx1`), so the duplicate survives rather than replacing the original.
2. **A single `solve_corridor` can partially commit.** It mutates the drawing candidate-by-candidate then runs `on_place`/`on_drop`; a raise part-way leaves annotations/coverage/warnings from the failed attempt, and a retry operates against them (relocations, double callbacks).

## Failure sketch

A deferred `locate()` with non-zero X and Y commits the plan corridor as `m_locx0`; the side-view solve then raises. On retry the retained locate intent queues the same plan dimension as `m_locx1` — both remain, silently duplicating one measurement.

## Why deferred

A correct fix is finalize-retry transactionality (drop each corridor's source intents as it commits, and/or make `solve_corridor` rollback-safe). That's a cross-cutting change to the recompose model (ADR 0009/0012), well outside the two-pass carve→corridor migration in #636. Tracking under the #635 consolidation epic.

Part of #635.

PR #757, opened 2026-07-19
fix(#661): finalize resolves queued detail requests; Y-turned step emission fixed

### LINKED ISSUE #661: finalize() reconstruction never resolves queued DetailRequests (no detail views on the edit path)
Spun out of the Codex review of PR #660 (#647/#639). **Pre-existing** — `main`'s `finalize()` has never called `_resolve_details`, so this is not a regression from that PR; filing so it isn't lost.

## Problem

The auto-pass (`orchestrator._auto_annotate`) ends by calling `_resolve_details(dwg, a, ctx=ctx)`, which turns every queued `DetailRequest` into a lettered DETAIL view. `Drawing.finalize()` (the deferred-edit reconstruction path) threads the same `PlacementContext` through `render_step_lengths` / `render_height_ladder` — which can append a `DetailRequest` (e.g. an X-turned crowded head-block, or a prismatic crowded step band via `_request_prismatic_detail`) — but **never calls `_resolve_details`**. The request is silently discarded when the per-run ctx is dropped at return.

## Repro sketch

Reconstruct (via `with dwg.deferred(): …` + step-length intents) an X-turned part with ≥2 adjacent head steps narrower on the page than `2 * draft.arrow_length`. The auto-pass produces `detail_a` + individual detail step dims; the finalize reconstruction of the same part produces neither.

## Fix direction

Call `_resolve_details(self, a, ctx=ctx)` in `finalize()` after leg D (`_maybe_tabulate_holes`), mirroring the auto-pass order (details resolve after the section + tabulation so the detailer sees them as obstacles). Needs a test on the turned-head + prismatic-step reconstruction paths.

Part of the #635 consolidation epic (editable-surface parity between the auto-pass and finalize).

pzfreo:
Loop item 2 of the script/CLI-parity cluster. Closes #661. Two commits.

## Commit 1 — the missing finalize stages (#661)

`_drain_intents` gains the `detail_request` and `details` thunks the canonical `_PASS_SEQUENCE` already ordered — previously prismatic detail requests were never queued on the edit path and turned-head requests died with the per-run ctx. Design decisions (each verified against the auto path):

- **`detail_view` persists on `BuildState`** (filled at the one `builder._assemble` site; writer-guard allowlist extended) so finalize gates identically to the auto path — flag AND escalation, never escalation alone. The parity test's existing `detail_view=True` patching works unchanged; no new emitter surface.
- **Iso ordering mirrored**: the auto path resolves details *before* the iso grows; finalize inherits the grown iso, which starved a knife-edge placement. The details stage re-projects the iso at sheet scale for the resolve and refits after — both mutations inside the #647 snapshot set (deterministic; bug-for-bug parity including the auto path's known ~1.6 mm iso/detail-dim overlap).
- **Height-demotion generalised**: `_resolve_details`' retry was name-coupled to `dim_height`; a scripted envelope-height verb replays as `dim_lengthN` — new `_overall_height_name` identifies it by envelope attribution + footprint.
- New rollback test: a raise after `detail_a` placed → view/coords/annotations/intent all restored, clean retry places it exactly once.

## Commit 2 — the Y-turned crash

The characterization xfail framed it as script-only; it also crashed **direct** builds (`render_step_lengths` → degenerate front projection → helpers raises). Root fix per the #731 axis contract: the renderer skips non-x/z step groups (info log), the emitter emits the flagged gap-comment pattern instead of an unreplayable verb, and `_classify_intents.len_ids` gets the same x/z guard so a hand-written Y step-length intent live-replays — and genuinely places (side view).

## Acceptance

`test_script_detail_parity.py`: **1 passed / 6 xfailed → 5 passed / 2 xfailed** — four strict xfails flipped honestly (markers removed): both prismatic detail tests, the turned-head detail test, and the Y-axis policy test. Remaining: side-drilled locations, rotational furniture (the #707 re-scope items).

## Verification

- Guards (incl. the extended writer allowlist) 81 · targeted detail/finalize/intent 76 passed · goldens 14, zero shift
- Full fast tier: **1478 passed, 3 skipped, 2 xfailed — zero failures** (even the #710 hypothesis test passed)
- lint ×4 clean; CHANGELOG (two Fixed entries); ADR 0012/index rows updated honestly


src/draftwright/annotations/sections.py
```diff
@@ -563,28 +564,66 @@ def _resolve_details(dwg, a: Analysis, *, ctx) -> None:
...
+    equal to the part's overall height.
+
+    The generalised path is demotion-safe, not best-effort (Codex review): a
+    **pinned** name is never a candidate — a pin is the user's "this stays put"
+    (ADR 0012), which outranks the demotion heuristic — and the match must be
+    **unambiguous**: zero or several surviving candidates (e.g. a hand-authored
+    twin of the height dim, or a square part whose depth label equals its height)
+    return ``None``, and the caller simply proceeds without a demotion retry."""
+    if "dim_height" in dwg.annotations():
+        return "dim_height"
```
pzfreo: The pin-safety check is bypassed for the canonical auto-pass name. `dwg.pin("dim_height")` is a public operation, but this early return hands that pinned annotation to `_resolve_details`, which removes it before retrying. It also accepts any user replacement registered under that name without validating feature/label/shape. Please run `dim_height` through the same pinned/type/identity checks as generalized candidates (or at minimum reject it when pinned), and add the canonical-name regression—the current test only exercises `dim_length0`.

src/draftwright/annotations/sections.py
```diff
@@ -563,28 +564,66 @@ def _resolve_details(dwg, a: Analysis, *, ctx) -> None:
...
                 _log.warning(
-                    "dim_height demoted: the requested crowded-step detail view takes its room"
+                    "%s demoted: the requested crowded-step detail view takes its room", hname
                 )
             else:
                 # Transactional (#689 review): the detail may fail for reasons other
-                # than dim_height's room — restore it rather than losing BOTH.
-                dwg.add(hobj, "dim_height", view=hview)
+                # than the height dim's room — restore it rather than losing BOTH.
+                dwg.add(hobj, hname, view=hview)
```
pzfreo: The failed-retry rollback restores only object/name/view, but `remove()` also deletes feature provenance (and pin state). For a generalized `dim_lengthN`, the restored dimension therefore disappears from `annotations_of(envelope)`, so `_overall_height_name` cannot find it on a later finalize and semantic `drop(envelope)` no longer removes it. Snapshot `feature_of(hname)` (and pin defensively) before removal and restore all registry metadata, or make this temporary removal transactional. A regression should force the second `_render_detail` to fail and assert provenance is unchanged after restoration.

pzfreo:
Adversarial review: two correctness issues remain in the detail-height demotion path.

1. The canonical `dim_height` fast path bypasses the new pin/identity checks, so a publicly pinned—or user-replaced—annotation can still be removed during the detail retry.
2. If the retry also fails, the remove/re-add rollback restores object/name/view but loses feature provenance (and any pin metadata), breaking subsequent semantic discovery and `drop(feature)` behavior.

I left inline comments with suggested fixes and regression coverage. CI is green, but I recommend holding merge until both are addressed.

pzfreo:
Both findings addressed in 859ecf1 — thanks, these were real.

1. **Canonical `dim_height` pin bypass** — `_overall_height_name` no longer returns `dim_height` unconditionally. It now runs the canonical name through the same guards as the generalised `dim_length{n}` path: never a candidate when **pinned** (a pin is the user's 'stays put', ADR 0012), and rejected when its label no longer equals the part height (identity — guards a user replacement). A guard failure falls through and returns `None` → no demotion.

2. **Provenance loss on failed retry** — the restore now snapshots `feature_of` + pin state *before* `remove()` and reinstates both, so the restored height dim keeps its envelope attribution (`annotations_of`/`drop`/later re-discovery). (`remove()` had dropped feature + pin per #89.)

Regressions added: the canonical `dim_height` pin/identity guard (the prior test only covered `dim_length0`), and a forced failed-retry asserting the restored dim keeps its envelope provenance. Lint + encapsulation guard clean; CI re-running.

PR #760, opened 2026-07-19
fix(#632): model and reconcile boss heights

### LINKED ISSUE #632: 0.3.0: declarative-mode coverage lint false-negative — passes 1.0 with an undimensioned axis
### Summary

In declarative `Sheet` mode, `Drawing.lint_summary()` returns `passed=True, score=1.0` for a part that is genuinely **under-dimensioned** — e.g. a boss with no height/Z dimension anywhere on the sheet. Declarative mode appears to credit coverage from *declared* features without checking that each produced a rendered dimension, so a real omission passes clean. This is the mirror image of the false-*positive* discussed in #628.

### Repro

```python
from draftwright import Sheet

WALL=3.0; L,W,H=90,64,38
cover = (Box(L,W,H) - Pos(0,0,-WALL)*Box(L-2*WALL,W-2*WALL,H)
         + Pos(0,0,H/2+5)*Cylinder(14,10) - Pos(0,0,15)*Cylinder(6,40))
boss = Pos(0,0,H/2+5)*Cylinder(14,10); bore = Pos(0,0,15)*Cylinder(6,40)

sh = Sheet(cover, title="C")
sh.envelope(); sh.boss(boss); sh.hole(bore).through()
sh.pocket(Pos(0,0,-WALL)*Box(L-2*WALL,W-2*WALL,H))
# NB: no height/Z dimension declared at all
dr = sh.build()
print(dr.lint_summary())   # {'passed': True, 'score': 1.0, 'warnings': 0, ...}
```

The rendered sheet has **no** vertical dimension (neither overall height nor boss height), yet lint is perfectly clean. A consumer trusting the lint would ship an ambiguous drawing.

### Expected

Declarative-mode coverage should verify that declared features actually produced rendered dimensions along each principal axis (or otherwise flag axes with no dimension), rather than crediting coverage from the declaration alone. Pairs with #628 (false positive on the recognition side) — the coverage check needs to reconcile *rendered* annotations, not intent, in both directions.

Environment: draftwright 0.3.0.

pzfreo:
## Summary

- carry detected and object-declared boss axial height/span through `BossFeature` and the dimension planner
- render a prismatic boss-height linear dimension in a profile-view corridor alongside the existing end-on diameter leader
- reconcile each modeled prismatic boss height against live feature-owned dimensions and report `boss_height_missing` when absent
- preserve the existing turned OD/step policy and legacy explicit diameter-only boss declarations

## Root cause

The existing completeness checks reconciled boss diameters, turned axial lengths, and hole locations, but `BossFeature` did not contain the boss axial extent. As a result, the reported boss could receive `ø28` and the part could receive an overall `48` envelope dimension while the boss-specific `10` height was neither rendered nor checkable.

## User impact

Declarative and detected prismatic bosses now document their independent height. Removing or dropping that height no longer leaves a falsely perfect lint score; assembly drawings receive the same informational severity policy as the other coverage checks.

Closes #632.

## Validation

- exact issue repro: `m_bossdia_z0 ø28` plus `m_bossheight_z0 10`; lint clean
- focused boss/model suite: 27 passed
- script-detail/model suite: 21 passed, 6 expected xfails
- Ruff format/check and `git diff --check`: clean
- full fast suite reached 58% with no failures before the local run was interrupted; GitHub CI will run the complete matrix

### DIFF
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -1564,6 +1564,67 @@ def render_boss_diameters(dwg, groups, a, *, ctx) -> int:
     )
 
 
+def render_boss_heights(dwg, groups, a, *, ctx) -> int:
+    """Queue the axial height of each prismatic boss in a profile-view corridor (#632)."""
+    if a.is_rotational or a.prof is not None:
+        return 0
+    tier = dwg.draft.font_size + 2 * dwg.draft.pad_around_text
+    specs = {
+        "z": ("front", "right", a.fv_zones.right, "x"),
+        "x": ("front", "above", a.fv_zones.above, "y"),
+        "y": ("side", "above", a.sv_zones.above, "y"),
+    }
+    n = 0
+    for bi, g in enumerate(
+        sorted(
+            (g for g in groups if g.feature_kind == "boss"),
+            key=lambda g: (g.feature.frame.axis, g.feature.frame.origin),
+        )
+    ):
+        b = g.feature
+        pd = next(
+            (d for d in g.dims if (d.param.role, d.param.kind) == ("boss_height", "length")),
+            None,
+        )
+        if pd is None or pd.suppressed or pd.param.span is None:
+            continue
+        view, side, strip, stack = specs[b.frame.axis]
+        p1 = dwg.at(view, *pd.param.span[0])
+        p2 = dwg.at(view, *pd.param.span[1])
+        edge = max(p1[0], p2[0]) if side == "right" else max(p1[1], p2[1])
+        label = _fmt(pd.param.value) + _tol_suffix(pd.param.tolerance, dwg.draft)
+        name = f"m_bossheight_{b.frame.axis}{bi}"
+
+        def build(pos, p1=p1, p2=p2, side=side, edge=edge, label=label):
+            return _dim(p1, p2, side, abs(pos - edge), dwg.draft, label=label)
+
+        def footprint(pos, p1=p1, p2=p2, side=side, edge=edge, label=label):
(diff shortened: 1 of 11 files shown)


src/draftwright/annotations/from_model.py
```diff
@@ -1564,6 +1564,67 @@ def render_boss_diameters(dwg, groups, a, *, ctx) -> int:
...
+            tier,
+            CorridorCandidate(
+                name=name,
+                build=build,
+                order=(_SIZE_SUBCHAIN, bi, name),
+                on_place=lambda _nm: None,
+                on_drop=lambda _nm, value=pd.param.value: ctx.record_issue(
+                    "warning", "boss_height_dropped", f"boss height {_fmt(value)} not dimensioned"
+                ),
+                force=True,
```
pzfreo, line 1620: This creates two warnings for the same omission. When the forced corridor cannot place the height, `on_drop` records `boss_height_dropped`; `Drawing.lint()` then sees no live `Dimension` for the boss and adds `boss_height_missing`. The existing coverage contract deliberately excludes build-time drops from reconciliation (e.g. `lint_feature_coverage(exclude=...)`), and #632’s scoped plan explicitly called out no double-reporting. Please make one code authoritative or thread a per-feature exclusion into the reconciliation, and add a constrained-strip regression asserting exactly one issue.

pzfreo:
Addressed the fresh adversarial finding in `47039ec`: live boss-height reconciliation is now the sole authoritative omission signal, so a corridor drop no longer adds a duplicate build-time warning. Added a constrained no-strip regression proving the result is exactly one `boss_height_missing` issue. Ruff, format, mypy, and the focused original-repro + forced-drop tests pass.

Issue #639, opened 2026-07-15
Consolidation P4: ADR 0005 §2 — explicit PlacementContext; retire Drawing as the build-state bus


pzfreo:
Part of #635 (Phase 4). Executes the ADR 0005 §2 follow-up that has been "deferred" since the module split was declared complete.

## Problem

`Drawing` is still the implicit state bus ADR 0005 set out to kill. The annotation passes reach into it through **72 distinct leading-underscore accesses** (`dwg._record_build_issue` ×21, `dwg._escalations` appended directly, `dwg._named`, `dwg._part_model`, `dwg._corridor_batch`, `dwg._detail_requests`, …). Worst symptoms:

- The orchestrator **creates** private scratch attributes on `dwg` (`orchestrator.py:~172`: `dwg._detail_requests = []`, `dwg._escalations = []`, `dwg._corridor_batch = {}`) and `finalize` re-creates them defensively (`drawing.py:~1437`).
- Passes probe for build context with `getattr(dwg, "_analysis", None)` (`holes.py:91, 214, 345`; `sections.py:100`) — the exact reach-through antipattern §2 names.
- The "renderers import down, no cycle" property currently rests on duck-typing `dwg` rather than a declared interface.

## Task

- Define a `PlacementContext` (name TBD) owning build context (`Analysis`, view-edge cache, part model, hole-feature index) and per-run placement state (corridor batch, escalations, detail requests, intent records), constructed by `builder`/orchestrator and passed to every pass explicitly.
- Route build-issue recording and escalation through declared methods on that context (or the existing registry), not `dwg._*` writes.
- `Drawing` keeps user-facing result state only (items, registry delegation, export); the ADR 0005 §2 note and CLAUDE.md updated when done.

## Why now

Beyond hygiene: a declared context is a **testable seam**. The current suite validates `from_model.py` largely through its private helpers (see Phase 6, and `test_render_seam.py` / `test_strip_layout.py` imports of `_bore_half_span`, `_record_slot_drop`, …) because there is no public boundary to assert against. This phase creates one.

## Acceptance

- No `dwg._*` attribute **writes** from `annotations/`; reads limited to a documented compat allowlist that shrinks to zero.
- No `getattr(dwg, "_analysis", ...)` probing — context is a required parameter.
- One initialisation site for placement scratch state.
- Suites green; `test_import_boundaries.py`-style fail-closed check for the allowlist.

[...]

Issue #133, opened 2026-06-26
Locate side-drilled (X/Y-axis) holes — _add_location_dims is plan-view only


pzfreo:
## Gap

`_add_location_dims` only dimensions the position of **Z-axis (plan-view)** holes
(`z_holes = [h … if _axis_letter(h) == "z"]`, annotate.py:1148). Holes drilled
into the **side faces** (X- or Y-axis) get a diameter callout but **no location
dimensions** — a machinist can't position them.

## Quantified (drawing-service honest scoreboard)

`honest = lint_score × location_coverage`. The lint score rates these 1.00; the
location-coverage axis exposes the gap on real `pzfreo/gramel` parts:

| part | lint | loc | honest |
|------|------|-----|--------|
| gramel_shaft (X-axis taps + grub bore) | 1.00 | **0.00** | **0.00** |
| gramel_drive_plate (2 side holes) | 1.00 | **0.00** | **0.00** |
| gramel_shank | 1.00 | 0.33 | 0.33 |

Scoreboard: `maquetto:drawing-service/eval/score.py`.

## Fix

Extend `_add_location_dims` beyond the plan view: place location dims for
- **X-axis holes** in the side view (Y- and Z-offsets), and
- **Y-axis holes** in the front view (X- and Z-offsets),

mirroring the plan-view X/Y tiering onto the corresponding view strips. Datum =
the same min-corner convention. Pattern-located side holes (a side bolt circle /
array) are covered by their pattern callout and skipped, as in the plan path.

## Acceptance

`location_coverage` for gramel_shaft / drive_plate / shank rises toward 1.0;
their `honest` score rises from ~0 to match the lint score. No regression on the
plan-view path (the synthetic corpus stays honest 0.99).

PR #763, opened 2026-07-19
fix(#426): reconstruct side-drilled hole locations in the generated script

### LINKED ISSUE #426: Design: record/finalize remaining manual intent verbs through the corridor solve

pzfreo:
## What

The **last** direct-vs-generated-script parity gap of #426. A side-drilled (X/Y-axis) bore gets `dim_loc_*` position dims in the direct build, but the imperative `--script` emitter emitted a gap comment instead of `dwg.locate(f)` for non-Z holes — because `locate()`/`render_locations` are Z-plan only (#133). Strict xfail `test_generated_script_matches_direct_side_drilled_locations`.

## How

Same shape as the rotational fix (#762): the machinery already existed. `_locate_off_axis_holes` (`annotations/holes.py`) is a **whole-model** pass that already draws the X/Y/Z side-drilled witness/datum geometry, and `_PASS_SEQUENCE` already lists the `off_axis_across`/`off_axis_along` slots. Wired into finalize:

- emitter emits **`dwg.locate(f)` for holes on any axis** (drop the gap comment).
- `_classify_intents` restricts `corridor_ids` to **Z-axis** locates, adds **`off_axis_loc_ids`** for X/Y; two **`_s_off_axis_{across,along}`** drain stages call `_locate_off_axis_holes` (register-only → placed at the shared drain); **routed out of `_s_live_replay`** so `add_feature_location`'s Z-only `ValueError` is never hit (`locate()`'s *deferred* branch records without the axis check); gated + dropped in `_s_drain` with the #639 retry-safety pattern.

The Z-only `locate()` **live** contract (#133) is untouched — side-drilled locates route to the off-axis stages before they can reach the live verb.

## Verification

- Strict xfail now **passes**; the parity suite is **0 xfailed** (both #426 gaps closed).
- 145-test regression sweep green; the one #427-F1 gap-comment test updated to assert the new `locate()` emission.
- lint ×4 clean; import-boundary / encapsulation / private-import guards pass; auto-pass path untouched.

With #762, this **closes #426** (the two remaining reconstruction gaps). Refs #133.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,18 @@
 
 ### Fixed
 
+- **Generated imperative script reconstructs side-drilled hole locations**
+  (#426/#133): an X/Y-axis (side-drilled) bore gets `dim_loc_*` position dims in
+  the direct build, but the `--script` emitter emitted a gap comment instead of a
+  `locate()` for non-Z holes (because `locate()`/`render_locations` are Z-plan
+  only, #133). The emitter now emits `dwg.locate(f)` for holes on any axis;
+  `finalize()` routes by the feature's axis — Z-plan holes drain through
+  `render_locations` as before, side-drilled bores through the whole-model
+  `_locate_off_axis_holes` pass (registered at the auto-pass's own
+  `off_axis_across`/`off_axis_along` stages, placed at the shared drain). The
+  Z-only `locate()` live contract is untouched (side-drilled locates never reach
+  it — they route to the off-axis stages). Closes the last direct-vs-generated
+  script parity gap tracked by #426.
 - **Generated imperative script reconstructs rotational furniture** (#426/#424):
   a turned/cylindrical part's `dwg.model()` carries a `RotationalFeature`, but the
   `--script` emitter parked it in the gap-comment list, so an executed
--- a/src/draftwright/builder.py
+++ b/src/draftwright/builder.py
@@ -851,15 +851,11 @@ def _feature_listing(a: Analysis) -> str:
             continue
         if kind in ("hole", "pattern"):
             body.append("dwg.callout(f)")
-            if feat.frame.axis == "z":
-                body.append("dwg.locate(f)")
-            else:
-                # locate() is Z-axis only (it rejects side-drilled bores by contract, #133);
-                # an off-axis bore's position is auto-pass-only. Flag it like a gap kind (#424).
-                body.append(
-                    f"#     locate() is Z-axis only — this {feat.frame.axis}-drilled bore's "
-                    "position is auto-pass-only (#133). auto_dims=True to keep it."
-                )
+            # locate() records the position intent for a hole on ANY axis (#426/#133): a
+            # Z-plan hole drains through render_locations, a side-drilled (X/Y) bore through
+            # the whole-model _locate_off_axis_holes pass — finalize routes by the feature's
+            # axis, so the same verb line reconstructs both.
+            body.append("dwg.locate(f)")
             body.append("dwg.furniture(f)")
         elif kind in ("step", "boss"):
             if feat.frame.axis in ("x", "z"):
--- a/src/draftwright/builder.py
+++ b/src/draftwright/builder.py
@@ -827,8 +827,11 @@ def _feature_listing(a: Analysis) -> str:
     concentric-bore leaders, #424/#426). A section A–A is recorded when a
     counterbored/spotfaced/blind Z-hole warrants one (finalize renders it last). Feature
     kinds with no verb yet (pmi) are emitted as flagged comments naming the gap (#424) —
-    never silently dropped. Commenting any line drops exactly that intent (nothing is
-    auto-drawn, so there is no double-dimension risk). Pure function of *a*.
+    never silently dropped. Commenting a per-feature verb line drops exactly that intent
+    (nothing is auto-drawn, so there is no double-dimension risk); the WHOLE-MODEL passes —
+    ``rotational``, the ``dwg.locate(f)`` of side-drilled bores, and the ``role="step_height"``
+    height ladder — reconstruct as a set, so commenting *some* of their lines still redraws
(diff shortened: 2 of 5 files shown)


pzfreo:
Codex review addressed in 83e5f38:

**Concern 2 (axes-restricted side-drilled locate could still hit the Z-only `ValueError`)** — fixed. `off_axis_loc_ids` no longer requires `axes is None`; every side-drilled (X/Y) locate now routes to the off-axis stages regardless of the `axes` selector (the off-axis pass ignores it anyway), so `add_feature_location`'s Z-only raise is unreachable for a side-drilled feature. The emitter only ever produces `axes=None`, so this closes a latent *hand-edit* crash. Verified: a deferred `locate(side_drilled, axes=("x",))` now finalizes cleanly and places `dim_loc_*`.

**Concern 4 (whole-model pass breaks the per-line-drop promise)** — accepted + documented. This is the same established semantics as the `rotational` and height-ladder whole-model passes (my inline stage comment already noted it). Made the `_feature_listing` docstring honest: per-feature verbs drop per line; the whole-model passes reconstruct as a set (comment *all* their lines to drop the group). No behaviour change — it matches the direct build, which places all side-drilled locations.

Concerns 1 and 3 were confirmed SOUND. Parity suite 0-xfailed; lint + guards clean.

PR #808, opened 2026-07-22
refactor(#523): break the builder→cli→sheet_emit import cycle

### LINKED ISSUE #523: tech-debt: break the builder / cli / sheet_emit import cycle
## Problem

The package currently has a small import cycle across the library build path, CLI entrypoint, and Sheet script emitter:

- `src/draftwright/cli.py:165` imports `build_drawing` / `generate_script` from `draftwright.builder`.
- `src/draftwright/cli.py:169` imports `generate_sheet_script` / `resolve_object_spec` from `draftwright.sheet_emit`.
- `src/draftwright/sheet_emit.py:27` imports `detect_part_model` from `draftwright.builder`.
- `src/draftwright/builder.py:926-928` lazily imports `app` from `draftwright.cli`.

That forms `builder -> cli -> sheet_emit -> builder`. The lazy import in `builder.py` avoids an immediate runtime failure, but the layering is still muddier than it needs to be: CLI concerns point into builder, builder points back into CLI, and the emitter shares builder internals.

## Why it matters

This is low urgency, but it is a maintainability smell:

- it makes package import boundaries harder to reason about;
- it increases the chance of accidental import-time side effects as the CLI/emitter surface grows;
- it obscures the intended direction of dependencies between public library APIs, CLI wiring, and script generation.

## Likely direction

Keep the dependency direction one-way:

- `cli.py` should depend on library/build/emitter services;
- `builder.py` should not import the Typer app or CLI module;
- shared script-generation helpers should live in a library-level module that both CLI and builder can call without cycling.

One concrete option: move the console-app accessor/entry wiring out of `builder.py`, or make the legacy `builder.app` export come from a thin compatibility module that does not force `builder -> cli`.

## Acceptance

- Import graph has no strongly connected component involving `draftwright.builder`, `draftwright.cli`, and `draftwright.sheet_emit`.
- `import draftwright.builder`, `import draftwright.cli`, and `import draftwright.sheet_emit` still work.
- Existing CLI behavior and script emission are unchanged.
- Ruff, mypy, and relevant CLI/script-emitter tests stay green.

pzfreo:
Closes #523.

## What

`builder._cli` was a compat shim whose body lazily ran `from draftwright.cli import app`. That single upward edge was the only thing holding the `builder`/`cli`/`sheet_emit` **strongly-connected component** together — the other three edges (`cli→builder`, `cli→sheet_emit`, `sheet_emit→builder`) are all downward.

This moves the `_cli` shim from `builder.py` to `cli.py`, right beside the Typer `app` it launches, so it needs no import at all. `make_drawing` re-exports `_cli` from `cli` instead of `builder`. `builder` no longer references `cli`.

The graph is now a plain DAG:

```
cli → {builder, sheet_emit}
sheet_emit → builder
builder → ∅
```

The last `_LAZY_UPWARD_EXEMPT` entry (`("builder","cli")`) is removed — the dict is now empty, kept closed as a ratchet.

## Acceptance (from the issue)

- ✅ No SCC involving `builder`/`cli`/`sheet_emit` (verified by AST scan + `test_import_boundaries`).
- ✅ `import draftwright.builder` / `.cli` / `.sheet_emit` all still work.
- ✅ CLI behaviour + script emission unchanged — the console-script entry point was always `draftwright.cli:app`, and `from draftwright.make_drawing import _cli` still resolves.
- ✅ Ruff, ruff-format, mypy, codespell clean; `test_import_boundaries` (11) + CLI/script-emitter suite (126) green.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,7 +46,8 @@ entry. Keep `_LAYERS` and this section in step.
     measure-and-repack → `Drawing`), `make_drawing` (+ export), and the
     editable-script generator (`generate_script`). Imports `drawing`/`analysis`/
     the annotation orchestrator/the stage modules — never `make_drawing` (a DAG).
-    *(The CLI moved out to `cli.py`; `_cli` is now a thin shim.)*
+    *(The CLI moved out to `cli.py`; the `_cli` compat shim lives there too (#523),
+    so `builder` no longer imports `cli`.)*
   - **`cli.py`** — the Typer command-line interface (#289): argument parsing,
     `--version`, shell completion, `--format`, rich help. The engine (build123d)
     is imported **lazily inside the command body** so completion/`--help`/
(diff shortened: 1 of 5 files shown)


pzfreo:
### Adversarial review (Codex) — resolved

No newly-introduced cycle; import-boundary + facade/CLI smoke tests pass. Two points raised:

1. **`from draftwright.builder import _cli` now raises, and `python -m draftwright.builder` is a no-op** — *accepted, by design.* Preserving `builder._cli` would require `builder → cli`, which is exactly the edge this PR removes; it cannot be kept without defeating #523. Verified nothing in the repo, tests, docs, or `[project.scripts]` depends on either: `builder._cli`'s only consumer was `make_drawing` (re-pointed to `cli`), and the console-script entry point is `draftwright.cli:app`. `builder._cli` was a private leftover of the #289 split, never a supported surface. The canonical shim path (`from draftwright.make_drawing import _cli`) and `python -m draftwright.make_drawing` both still work.

2. **Stale docstring in `make_drawing.py`** — fixed in 6e8e3e8.

PR #813, opened 2026-07-22
fix(#812): reconstruct plate thickness dimensions in the emitted --script

### LINKED ISSUE #812: Script parity: plate thickness dimensions are dropped in the emitted --script reconstruction
## Summary

The imperative `--script` emitter (`generate_script`) does not reconstruct **plate thickness** dimensions: the executed script's drawing has no `dim_plate_*` annotations even though the direct `build_drawing` path draws them. This is a script↔CLI parity divergence, distinct from the machined-**leader-callout** kinds (pocket/fillet/flat/chamfer/groove) fixed in #811.

## Reproduction

An L-bracket (base plate + upright wall) exports to STEP; direct build vs. generated script:

```
DIRECT plate dims: ['dim_plate_y0', 'dim_plate_z0']
EMIT: dwg.dimension(f, "length", role="thickness")   # 8   (×2)
SCRIPT plate dims: []
```

The emitter correctly emits `dwg.dimension(f, "length", role="thickness")` for each `PlateFeature` (its `parameters()` returns a spanned `DimParameter`), but that recorded intent is never routed to `render_plates` in `finalize()`, so nothing is placed.

## Root cause

Unlike the five leader-callout kinds, a plate is a **spanned linear dimension** placed by `render_plates`, which only *registers corridor candidates* — the placement happens in the shared `drain_and_reconcile`. The finalize path has no route that (a) buckets a `role="thickness"` plate-dimension intent, (b) calls `render_plates(..., only=<plate features>)` at the `plates` `_PASS_SEQUENCE` slot, and (c) includes plates in the `_s_drain` guard so the corridor is actually drained. This bug predates #811 (verified on `main`).

## Proposed fix

Route the plate thickness dimension intent through a dedicated finalize stage:
- Classify `dwg.dimension(f, "length", role="thickness")` intents on `PlateFeature`s into a `plate_feats` bucket in `_classify_intents`.
- Add a `_s_plates` stage that calls `render_plates(self, plan_dimensions(model), a, ctx=ctx, only=plate_feats)` at the existing `plates` slot (before the drain), threading `only=` through `render_plates` (skip-in-place to keep `dim_plate_*` indices stable — the `render_slots` idiom).
- Add plates to the `_s_drain` guard + its intent-removal set so the registered corridor candidates are drained.
- Canary: an L-bracket whose script reproduces `dim_plate_*` exactly (fails on current `main`/#811 head).

Discovered during the #811 Codex review (round 2, finding 1). #811 deliberately scoped plate out and kept the five genuine leader-callout kinds.

pzfreo:
## Summary

Closes #812. The imperative `--script` emitter did not reconstruct **plate thickness** dimensions: the executed script's drawing had no `dim_plate_*` annotations even though the direct `build_drawing` path draws them.

A `PlateFeature` is a **spanned dimension** — its `parameters()` returns a `("length", "thickness")` `DimParameter` with a span — so the emitter correctly emits `dwg.dimension(f, "length", role="thickness")`. But that intent matched **no route** in `_classify_intents` and was silently dropped in `finalize()`, so nothing was placed. (Distinct from the machined leader-callout kinds fixed in #811; this predates #811 — verified on `main`.)

## Fix

Route plate thickness like a slot — a corridor-registered dimension drained by the shared solve:
- **`_classify_intents`**: classify `role="thickness"` plate-dimension intents into `plate_ids`/`plate_feats`.
- **`_s_plates`**: registers `render_plates(..., only=plate_feats)` at the existing `plates` `_PASS_SEQUENCE` slot (before the drain); `_s_drain` now runs when plates are present and drops the intents only after a successful solve (clean retry on a raise, per the #647 transaction).
- **`render_plates`** regains its `only=` subset (skip-in-place *after* consuming the per-axis index, so `dim_plate_*` names stay stable); `only=None` is the unchanged auto-pass.
- **`plate_ids`** joins the live-replay routed set so a plate dim doesn't double-place.

## Testing

- New canary `test_generated_script_reproduces_plate_thickness_dims`: an L-bracket (base plate + upright wall) whose script reproduces `dim_plate_y0`/`dim_plate_z0` exactly — **verified to fail on `main`** (`[] != [...]`) via the worktree ritual.
- Full local suites green: parity module (17), make_drawing (767), e2e_standards + carve/declare/tolerances/encapsulation/fillets (241). All four lint checks clean.

## Context

This is the last discrete **emit gap** in the script↔CLI parity sweep. After #810 (NTS note), #811 (machined leader callouts), and this, the residual corpus divergences are the finalize-vs-auto **placement-quality** differences (crowded-sheet layout-driven drops) — explicitly out of byte-identity scope (ADR 0004/0012) and tracked by the #426/#661/#707 convergence work, not discrete emit bugs.

### DIFF
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -1811,7 +1811,7 @@ def render_boss_heights(dwg, groups, a, *, ctx) -> int:
     return n
 
 
-def render_plates(dwg, groups, a, *, ctx) -> int:
+def render_plates(dwg, groups, a, *, ctx, only=None) -> int:
     """Plate/wall thicknesses (#559): the thin extent of each recognised slab
     (`PlateFeature`), placed in the view where its thin axis is characteristic — a Z
     plate (horizontal slab) as a vertical dim left of the front elevation, a Y plate
@@ -1846,6 +1846,8 @@ def render_plates(dwg, groups, a, *, ctx) -> int:
         lbl = _fmt(val) + _tol_suffix(pd.param.tolerance, draft)
         i = counts[pl.axis]
         counts[pl.axis] += 1
+        if only is not None and pl not in only:
+            continue  # #812 finalize subset: skip AFTER consuming i so dim_plate names stay stable
         if pl.axis == "z":
             # Horizontal slab (base plate): vertical dim on the front-elevation left strip.
             # For a Z plate the in-plane centroids are (u=X, v=Y); the front view discards
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -198,6 +198,8 @@ class _IntentRouting:
     only_len: set
     slot_feats: set
     machined_ids_by_kind: dict
+    plate_ids: set
+    plate_feats: set
 
 
 @dataclass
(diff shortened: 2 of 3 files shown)


pzfreo:
Closing: the plate-thickness 'gap' was a name-keyed false positive. Content-keyed comparison shows both plate thicknesses are already reconstructed on main's script (as generically-named `dim_length*` dims via live-replay through `dimension()`) — identical content to the direct build, only differently named. Per the no-byte-identity decision (ADR 0004/0012), generic-naming differences are out of scope, and this change additionally introduced two regressions (double-dimensioning with pin/priority; ignoring name/side/view). The genuine script↔CLI emit gaps were resolved by #810 (NTS note) and #811 (machined leader callouts).

Issue #841, opened 2026-07-24
Make repeated slot/pocket callouts reliable and expose section/detail control


pzfreo:
## Context

Follow-up to #837. Once the five tuner-jig pockets are declared manually, Draftwright 0.3.8 can render a useful drawing, but several annotation and API behaviours prevent a robust AI-authored production drawing.

The part is a 26 × 161 × 21 mm bar with five identical blind obround pockets, approximately 13.6 × 7.88 × 19 mm deep, on a common centreline at 27.2 mm pitch.

### 1. Repeated slot dimensions are partially dropped

Five explicit `sheet.slot(...)` declarations render all five widths, but only three of five length dimensions:

```text
slot1 length dim not placed (no room beside the plan)
slot2 length dim not placed (no room beside the plan)
code=slot_dim_dropped, severity=info
```

The output is visually noisy even where the repeated dimensions fit. For an identical pattern, the desired drawing is one representative size callout plus `5X`/pattern location dimensions.

### 2. Anchored notes are side-sensitive and may disappear

```python
sheet.note("5X OBROUND SLOT 13.60 X 7.88", slot_handle,
           view="plan", side="right")
```

exports successfully but drops the note:

```text
m_gdt0 not placed (no room in the plan right strip or its opposite)
code=gdt_dropped, severity=warning
```

Reduced matrix on this A3 layout:

- one slot: auto/above/below render; left/right drop
- five slots: auto/above render; below/left/right drop

Omitting `view`/`side` lets the solver place the note successfully.

### 3. `_Params.note()` chaining is surprising/broken

`sheet.slot(...)` and `sheet.pocket(...)` return `_Params`. The intuitive generated-code form:

```python
sheet.slot(...).note("...")
```

forwards to `Sheet.note` without supplying the handle as `ref` and raises:

```text
TypeError: Sheet.note() missing 1 required positional argument: 'ref'
```

The working form is:

```python
slot = sheet.slot(...)
sheet.note("...", slot)
```

This differs from `_Hole.note()` and `_Dim.note()` and is easy for an AI to get wrong.

### 4. No public `Sheet.section()` / `Sheet.detail()` declaration

The best-practice drawing needs a section through one representative blind pocket to communicate the floor/depth cleanly. Draftwright contains planner-triggered section/detail machinery, but 0.3.8 exposes no public Sheet verb to request one. The current approximation is a pocket leader containing `7.9 × 13.6 × 19 DEEP`.

### 5. Successful best-practice approximation

[...]

Issue #499, opened 2026-07-07
Reconciliation lint: extend declared→exists to patterns and slots


pzfreo:
## Follow-up from #487 (PR #498)

`lint_declaration_reconciliation` (#487) closes the *declared → exists* gap for the **cylindrical singletons** — `hole`, `boss`, `step` — matching each against recognised cylinders on axis + ⌀ + in-plane position. Two declared feature kinds were left out of scope and remain blind to a stale declaration:

- **`pattern`** — a `PatternFeature` has `.members` (multiple locations), each effectively a bore. Reconciling means confirming a matching cylinder at *each* member location (and deciding how many missing members warrant a warning — all vs any).
- **`slot`** — a `SlotFeature` is prismatic (no single defining cylinder), so it needs a different geometry check (e.g. matching against `find_slots` / the milled-slot recogniser by width/length/axis).

Aspects (`control_frame`/`datum_ref`/`finish`/`note`) are intentionally excluded — they annotate, they don't assert geometry.

### Acceptance
- A declared pattern whose member holes were removed from the part emits `declared_feature_absent` (or a pattern-specific code).
- A declared slot absent from the part emits a warning.
- No false positives on valid declared patterns/slots.

Machinery to reuse: `recognition.find_hole_patterns` / `find_slots`, and the `_RECON_DIA_TOL`/`_RECON_POS_TOL` tolerances already in `linting/coverage.py`.

PR #848, opened 2026-07-24
feat(#841): recognised pocket-pattern kind — declared path (PR 1/2)

pzfreo:
**#841 outcome 1**, PR 1 of 2 (approved scope A — full recognised kind). N identical blind pockets render as ONE grouped `count× W × L × D DEEP` callout + `(n-1)× pitch` dim(s), instead of N competing per-pocket size dims (some of which drop for lack of room — the tuner-jig symptom).

This PR is the **declared path end-to-end**; STEP recognition + the emitter follow in **PR 2**.

## What it does

For the tuner-jig bar (5 pockets, 27.2 pitch):
```python
member = pocket(width=7.88, length=13.6, depth=19, long_axis="x", width_axis="y", depth_axis="z", ...)
sheet.pocket_pattern(member, kind="linear", count=5, pitch=27.2, direction=(0, 1, 0))
```
→ `5× 7.9 × 13.6 × 19 DEEP` + `4× 27.2`, lint clean. Grid works too (`6× 8 × 8 × 4 DEEP` + `1× 30` / `2× 40`).

## Surfaces (mirroring the hole `PatternFeature` chain)

- **IR** — `PocketPatternFeature` (`model/ir.py`): composes a representative `member` PocketFeature + linear/grid arrangement; `parameters()` = member W/L/D + pitch/grid. `frame.axis` = the opening-normal (depth) axis, so the callout view derives correctly.
- **declare** — `pocket_pattern(member, *, kind, count, pitch|grid, ...)` (`model/declare.py`), mirroring `pattern()`; reuses `_pattern_members` to lay the array out in the pocket's opening plane. `Sheet.pocket_pattern` verb.
- **render** — `render_pocket_patterns` (`annotations/holes.py`, beside the pattern pitch furniture it reuses; pulls the pocket-leader helpers from `from_model` — the allowed import direction). One representative leader on the array centre + the pitch dim(s). Member pockets are composed into the pattern (not in `model.features`), so `render_pockets` never double-renders them.
- **orchestrator** — a `pocket_patterns` stage placed **pre-drain** (like a hole pattern — the pitch dim needs strip room the post-drain decoration slots lack).

## Design notes

- **Two pattern types, not one unified.** The codebase already keeps `SlotFeature`/`PocketFeature` separate (separate render passes, declare verbs), so a `PocketPatternFeature` now + a `SlotPatternFeature` sibling later is *more* consistent than a unified `RecessPatternFeature` — and lets this PR focus on the tuner-jig driver (pockets, leader-based) without dragging in slot linear-dim placement.
- **Coverage:** no-op — slots/pockets are deliberately out of coverage scope (#499).

## Tests

`tests/test_pocket_pattern.py` (5): declare composition, linear render (one grouped callout + pitch, members not double-rendered), grid render (both pitch dims), input guards, and `model()` inspection. Regression: 215 (boundaries/registry/declare/part-model/render-seam) + 91 (pocket/pattern/slot rendering) + smoke tier, all 4 lint checks green.

Part of #841. PR 2: `recognise_pocket_patterns` + detect converter + emitter + tuner-jig e2e.


pzfreo:
Addressed the round-1 Codex review in f4f9964:

**Correctness (fixed):**
- `pocket_pattern()` now enforces the opening-plane constraint — rejects a linear `direction=` with a depth-axis component and non-coplanar explicit `members=`.
- `render_pocket_patterns` dimensions the true extrema **along the array direction**, so an unordered `members=` override can't produce a pitch dim contradicting its `(n-1)×` label.

**Editable/emit surface — made honest, full wiring deferred:** the manual `callout()`/finalize + reconstruction-emit path for pocket patterns needs the same **pre-drain** furniture placement the auto pass uses (unlike the post-drain lone machined callouts), so it lands with the #841 outcome-3 follow-up. Here: `dwg.callout(pocket_pattern)` raises a clear error instead of crashing, and reconstruction emit flags the pattern with a comment instead of silently dropping it.

pzfreo:
Addressed the round-2 Codex review in a01e9c0:

**High — pitch-dim name collision (fixed):** pocket-pattern pitch dims now carry a distinct `dim_pocketpat_pitch` prefix (grid via a new `name_prefix=` kwarg, default unchanged for hole patterns), so a plan-view hole pattern and pocket pattern no longer overwrite each other on `dim_pitch_plan0`. New regression test builds both on one sheet and asserts both survive.

**Medium ×2 — linear `members=` truthfulness (fixed):** a linear array is collinear + constant-pitch by definition, so `pocket_pattern(kind='linear')` now **rejects** explicit `members=` — the computed `pitch=`/`direction=` layout is the only path that guarantees the `(n-1)× pitch` label matches the geometry. (Also corrected a self-inconsistent round-1 test.) Grid keeps `members=`; its pitch labels are geometry-derived.

9 pocket-pattern + 266 pattern/emit/declare regression + all 4 lint checks green.

pzfreo:
Addressed the round-3 Codex review in 132e245:

**High — grid `members=` unvalidated (fixed):** rather than add lattice/centroid validation, `pocket_pattern()` now **rejects explicit `members=` outright for both kinds** (a declared pattern is always computed from count + pitch/grid + layout; the detector builds the IR dataclass directly with real geometry, bypassing this constructor). Dead coplanarity/length checks removed.

**Medium — `rows=1`/`cols=1` grid (fixed):** now requires `rows>=2 and cols>=2` — a single-row/column array is linear (`kind='linear'`), so it no longer silently loses its pitch dim.

**Medium — orphan pitch furniture (fixed):** pitch dims are now gated on the grouped size/depth callout actually landing, so a dropped callout on a crowded sheet no longer leaves orphan pitch dims.

10 pocket-pattern + 334 pattern/declare/emit/holes regression + all 4 lint checks green.

pzfreo:
Codex round-7 re-review is **clean** — no substantive defects; confirmed the `math.hypot` fix is correct and that all other norm sites (`_add_grid_pitch_dims`, `_place_pitch_dim`, `_radial_candidates`) already use it. All 16 required CI checks green. Merging.

PR #854, opened 2026-07-24
feat(#841): slot-pattern editable surface — callout/finalize + re-emit (PR C)

pzfreo:
Completes the through-slot pattern sibling (declared #851, recognition+emit #852): the manual `dwg.callout()` edit verb and reconstruction re-emit for slot patterns are now wired, **mirroring the pocket-pattern editable surface (#850) exactly**.

- **`drawing.py` `callout()`** — a `slot_pattern` live branch renders the grouped `SLOT W × L` callout + pitch via `render_slot_patterns(only={feature})`, returning the `m_slotpat*` name; `view=`/`name=` rejected (auto-placed).
- **`drawing.py` finalize** — `_classify_intents` buckets slot-pattern callout intents (`slot_pattern_ids` on `_IntentRouting`); a new `_s_slot_patterns` stage drains them, keyed `"slot_patterns"` so `run_stages` fires it at its pre-drain slot; the ids join `_s_live_replay`'s `routed` set.
- **`builder._feature_listing`** — a `slot_pattern` branch emits `dwg.callout(f)` (no `locate`/`furniture` — a slot pattern has neither).

## Tests
- `test_manual_callout_verb_places_grouped_callout_and_pitch` + `test_deferred_callout_reconstructs_the_pattern` (live + deferred/finalize).
- The `_reconstruct` helper (mirrors `_feature_listing`) gains a `slot_pattern` branch.

This closes the #841 `SlotPatternFeature` sibling epic (declared → recognise+emit → editable, matching the pocket epic #848/#849/#850).

## Verification
18 slot-pattern + 16 reconstruction + 4 lint checks green locally; full fast tier running.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #481, opened 2026-07-06
render_gdt: fall through to another side before dropping a GD&T frame (P2c placement quality)


pzfreo:
## Problem

`render_gdt` (P2b) places each declared GD&T item (control frame / datum / finish) into a
**single, fixed** `(view, side)` strip. When that strip is congested it **drops the item**
with a `gdt_dropped` warning — it never tries the opposite side first. This bites the P2c
flagship use case: `sheet.control(bore).position(0.1, to="A").perpendicularity(0.05, to="A")`
stacks two wide frames, and on a congested strip one drops.

P2c.1/P2c.2 mitigated this with a **view-aware default side** (`_FEATURE_SIDE` in
`declare.py`: plan→above, front/side→below — the empirically roomiest per view) plus a
`side=` override. That makes the common case place, but it's a heuristic, not a cure: a part
whose default side *is* congested (e.g. a plan view with off-datum location dims above) still
drops, and the caller must hand-pick a side.

## What good looks like

Mirror the slot pass's `_below_or_drop` fallthrough (`from_model.py`): a GD&T candidate whose
`on_drop` fires should try the **opposite side** (and/or the left/right strips) before
recording the drop — so a frame lands *somewhere* legible rather than vanishing. The corridor
already exposes the per-candidate `on_drop` hook this needs.

## Scope

- `render_gdt`: on drop, attempt the alternate side(s) via the same strip machinery (a
  `carve_free_position` / `place_strip_candidates` retry, as `render_slots` does), then warn
  only if all sides are full.
- Keep it deterministic (ADR 0001) and lint-clean.
- Consider whether stacked frames on one feature should share a leader (a true
  `CompositeFeatureControlFrame`) rather than N independent same-site frames — a related
  legibility improvement (separate follow-up if larger).

## Context

- Renderer: #478 (P2b). Verbs: #479 (P2c). The `_FEATURE_SIDE` heuristic + `side=` override
  are the current stopgap.
- Related placement work: the ADR 0009 unification umbrella #477 (GD&T is a first-class
  corridor candidate; its selection/priority is #357).

PR #855, opened 2026-07-24
feat(#841): auto-relax anchored-note side instead of silently dropping (behaviour C)

pzfreo:
Closes the last substantive **#841** item (confirmed-behaviour #2/#3): an anchored note (or GD&T frame / surface finish) declared with an explicit `view=`/`side=` **silently dropped** when that strip had no room (`gdt_dropped`, warning) even while the export "succeeded" — a requested annotation vanishing without recourse.

Per the chosen design (option **C — auto-relax side, keep the note**):

## Change
Generalise the #481 opposite-side fallthrough in `render_gdt`'s on-drop `_retry`: when the requested strip is full, try the **opposite** side, then the two **perpendicular** sides (flipping the leader orientation), placing on the first strip with room — via the same `carve_free_position` / title-block-clear checks, so it **never overshoots** the drawable area. Only when **no** strip fits does it drop.

A relaxed placement records an **INFO `gdt_side_relaxed`** issue naming requested-vs-actual side, so the override is visible, not silent — the #841 goal is that a requested annotation is never *silently* lost. An explicit `side=` is now a **preference honoured when the strip has room**, not a hard constraint that forfeits the annotation.

## Tests
- `test_note_relaxes_side_when_requested_strip_full` — a wide note on the narrow left strip relaxes to above/below, places in-bounds, records `gdt_side_relaxed`, no `gdt_dropped`.
- `test_note_honors_requested_side_when_it_fits` — a fitting note keeps its requested side with **no** spurious relax notice.
- `test_wide_frame_in_narrow_strip_relaxes_not_overshoots` (was `..._drops_not_overshoots`) — the wide multi-datum FCF now relaxes + places **in bounds** rather than dropping; the real no-overshoot invariant (`annotation_out_of_bounds`) is preserved.

## Verification
14 GD&T-placement + 4 lint checks green locally; full fast tier running.

Note: the pocket + slot pattern epics (#848–#854) already closed confirmed-behaviour #1 and outcomes 2–4; this closes the remaining note-side-sensitivity item.

### DIFF
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -3448,28 +3448,50 @@ def render_gdt(dwg, model, a, *, ctx) -> int:
                 _zones=_zones,
                 _px=_px,
                 _py=_py,
-                _hz=_hz,
                 _sz=_sz,
                 _bld=_bld,
                 _feat=_feat,
                 _tb=_tb,
             ):
-                alt = {"above": "below", "below": "above", "left": "right", "right": "left"}[_s]
-                alt_strip = getattr(_zones, alt, None)
-                if alt_strip is not None:
-                    axis2 = "y" if alt in ("above", "below") else "x"
+                # Auto-relax the requested side (#841 outcome C): the requested strip is full, so
+                # try the OPPOSITE side, then the two PERPENDICULAR sides, placing on the first
+                # with room. A note the caller asked to see should appear somewhere legible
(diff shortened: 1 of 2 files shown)


pzfreo:
Codex review is **clean** — no substantive defects; verified the orientation flip, stacking-axis extent, perpendicular span, closure captures, post-drain timing, title-block rejection, and INFO-severity convention. Local full fast tier green (1654 passed); all 16 required CI checks green. Merging — this closes out #841.

Issue #754, opened 2026-07-19
refactor: route rotational diameter dimensions through planner output


pzfreo:
## Context

The #698 convergence close-out migrated the dimension-bearing feature passes that were explicitly in scope, but `RotationalFeature` remains a real planner bypass:

- `RotationalFeature.parameters()` emits OD and bore `DimParameter`s.
- `plan_dimensions()` therefore builds `DimensionGroup`s and folds authored decorations/tolerances onto them.
- `annotations/from_model.py::render_rotational` reads raw `rot.od` / `rot.bores` instead of the planned dimensions.

The computed groups are discarded, leaving authored tolerances and planner-level convention/suppression unreachable for this pass. This is the same class of latent risk that #698 removed from the other dimension-bearing renderers.

This is separate from #714, which owns deduplication of the mirrored per-axis rendering blocks.

## Proposed scope

- Make the rotational OD and concentric-bore dimensions consume their planned dimensions.
- Keep axis centrelines and other non-dimensional rotational furniture model-routed.
- Preserve the current X/Y/Z view and placement behavior.
- Add binding tests proving rendered values and authored tolerances come from planner output rather than raw feature fields.
- Update ADR 0015's planner coverage table once complete.

## Acceptance criteria

- `render_rotational` does not format OD/bore dimension values directly from raw `RotationalFeature` fields.
- Authored tolerance/fit decoration reaches rotational OD/bore labels.
- Axis-specific output and existing rotational coverage remain unchanged.
- ADR 0015 records rotational dimensions as planner-fed while distinguishing the remaining model-routed furniture.

Found while addressing the architecture accuracy review on #753.

Issue #631, opened 2026-07-14
0.3.0: Sheet.step(boss) emits no height dimension and drops an annotation (declarative API)


pzfreo:
### Summary

On the declarative `Sheet` API, `Sheet.step(boss)` (where `boss` is an external cylinder sitting on a face) does **not** emit a height dimension — and in fact **reduces** the annotation count. Adding the call drops an annotation rather than adding one.

### Repro

```python
from build123d import Box, Cylinder, Pos
from draftwright import Sheet

WALL=3.0; L,W,H=90,64,38
cover = (Box(L,W,H) - Pos(0,0,-WALL)*Box(L-2*WALL,W-2*WALL,H)
         + Pos(0,0,H/2+5)*Cylinder(14,10)          # Ø28 x 10 boss
         - Pos(0,0,15)*Cylinder(6,40))             # Ø12 bore
boss = Pos(0,0,H/2+5)*Cylinder(14,10)

def n(with_step):
    sh = Sheet(cover, title="C")
    sh.envelope(); sh.boss(boss); sh.hole(boss and Pos(0,0,15)*Cylinder(6,40)).through(); sh.pocket(Pos(0,0,-WALL)*Box(L-2*WALL,W-2*WALL,H))
    if with_step: sh.step(boss)
    return len(list(sh.build().iter_annotations()))

print("without step(boss):", n(False))   # 11
print("with    step(boss):", n(True))     # 10  <-- fewer!
```

`Sheet.step(boss)` returns a `_Dim` handle (so it *looks* like it declared something), but no height dimension appears in any view and the total annotation count goes **down**. The only way I found to dimension the boss height is the manual escape hatch:

```python
sh.dimension(kind="linear", value=10, label="10", dominant_axis="z",
             ref_pts=[(14,0,19),(14,0,29)])
```

### Expected

`Sheet.step(obj)` on a boss/raised feature should emit its height (Z extent) dimension, or raise if it doesn't apply — not silently drop an existing annotation.

Environment: draftwright 0.3.0.

Issue #630, opened 2026-07-14
0.3.0: detail_view=True is a no-op — no detail view added, crowded callouts still dropped


pzfreo:
### Summary

`build_drawing(part, detail_view=True)` (and `make_drawing(..., detail_view=True)`) is a **no-op** on a part whose lint explicitly recommends a detail view: the output is identical to `detail_view=False` — same view count, same annotations, no detail view added, and the crowded callouts it should rescue are still missing.

### Repro

Using the shelled-cover part from the sibling issue (Ø28 boss concentric with a Ø12 bore, which trips `step_dim_dropped: ... (use a detail view)`):

```python
from draftwright import build_drawing
# part = shelled cover (see sibling issue)
for dv in (False, True):
    dr = build_drawing(part, title="Cover", detail_view=dv)
    print(dv, len(dr.views), len(list(dr.iter_annotations())))
```

Output — **identical** for both:

```
False 4 15
True  4 15
```

Neither a 5th (detail) view is added nor is the dropped Ø28 diameter / boss-height dimension recovered. The `step_dim_dropped` lint's advice ("use a detail view") therefore has no corresponding working switch.

### Expected

`detail_view=True` should either add a scaled detail view that carries the callouts too crowded for the main views (resolving `step_dim_dropped` / `feature_not_dimensioned`), or — if it can't apply to a part — say so rather than silently producing the same drawing. Sibling: the boss under-dimensioning itself. Related: #56 (lint should suggest detail_view when shoulders crowd), #628 (coverage-lint reconciliation).

Environment: draftwright 0.3.0.

PR #866, opened 2026-07-26
docs(adr): propose ADR 0016 — declared dimensioning intent

pzfreo:
## Summary

Proposes **ADR 0016 — Declared dimensioning intent: capture what to measure, let the engine place it**, a direction-setting record (Status: **Proposed**, decider: pzfreo). No code changes.

Today a generated `Sheet` script mirrors the part's *features* but not the drawing's *dimensioning* — the ⌀ callouts, location ladders, and overall spans are derived by the planner and placed by the corridor solver, invisible in the code and unadjustable from it. This ADR captures how to close that gap.

## The decision — a referential dimension line

One verb: **`sheet.dim(<feature>, <role>)`** — names a feature (or the envelope) and the measurement to show, and **carries no number**. The value is read from the referenced geometry, so a size lives in exactly one place (the feature line for a STEP part; the build123d object for a live part) and a dimension **cannot drift** from it. That gives a *complete* per-dimension mirror of the sheet **and** single source of truth at once. The corridor solve (ADR 0014) still decides ordering, dedup, and position — **a dimension line references; the engine places.** The same verb covers three behaviours:

- **Surface & drop** — the emitter emits one `dim` line per dimension the drawing carries; commenting it out suppresses that dimension (keyed by `(feature, DimParameter role)`, never a coordinate).
- **Add** — a `dim` line for a measurement the planner wouldn't auto-produce (span, pitch, thickness) enters the shared solve as a new candidate.
- **Emphasise** — `.pin()` / `.priority()` chain onto a `dim` line (already ADR 0012).

Only *dimensions* become lines; low-level furniture (centre marks, section arrows, hatching) stays engine-automatic so the mirror doesn't explode.

## Why it fits (not fights) the ADRs

The **unfinished half of ADR 0001 §2**, which already says editing should be domain-semantic intent (*"dimension this bore's depth"*), not strip/zone machinery. ADR 0001 rejected only the hardcoded, place-everything editable DSL — not this. Extends ADR 0011 (declare features → declare dimensions) and ADR 0012 (edit one dim → the dim set); consumes ADR 0015 (planner) and ADR 0014 (solve). The `.thread` / `.finish` aspect is the archetype.

## Worked example

Includes the 80×50×8 mounting plate: the referential `sheet.dim(...)` lines read against the rendered sheet, showing `⌀20` appearing once (on the feature; the dim line references it) and that **dropping a dimension removes the callout, not the hole** — geometry and annotations are separate layers.

## Honest edges recorded

- **Suppression / full mirror is gated on the still-open global recompose (#426/#707)** and on stable `(feature, role)` identity for auto dimensions (ADR 0010 / ADR 0015). **Additive intent is reachable on today's machinery.**
- The emitter only mirrors what it can round-trip — same fidelity floor the feature emitter holds.

5-phase plan (additive → identity → suppression-by-omission → recompose → emitter dimension-mirror). Index row + reading path added.

## Testing

Docs-only. Guards (an emitted-`dim`-re-solves fidelity test and a reference-not-restate / scale-independence audit) are named in the ADR as deferred-until-landed.

https://claude.ai/code/session_01Dxfx6MBFZReLkZc8hRxfzK

### DIFF
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -59,23 +59,33 @@ do.** The generated script becomes a mirror of the drawing's *intent*, not its g
 every line is a droppable / editable declaration; the engine still owns derivation and
 placement; re-running re-solves.
 
-Three kinds of dimensioning intent, all scale- and placement-independent (anchored in
-feature / world terms, per the ADR 0012 constraint):
+The one surface is a **referential dimension line** — `sheet.dim(<feature>, <role>)` —
+that names a feature (or the envelope) and the measurement to show, and **carries no
+number**. The nominal value is always read from the referenced geometry, so a size lives
+in exactly one place — the feature declaration (for a STEP part, its detected snapshot;
+for a live part, the build123d object it reads from) — and a dimension can never drift
+from it. This is what dissolves the dual-source-of-truth objection to a complete
+per-dimension mirror: the mirror is complete *and* single-source because the lines
+reference rather than restate. The same verb covers three behaviours, all scale- and
+placement-independent (anchored in feature / world terms, per the ADR 0012 constraint):
 
-- **Additive** — declare a measurement the planner would not auto-produce, referencing
-  features or faces, not coordinates: a span between two features, a pattern pitch, a
-  wall thickness, an angle, an overall extent. It becomes a `CorridorCandidate` in the
-  shared solve like any planned dimension.
-- **Suppressive** — declare that an auto dimension should *not* appear, by its **semantic
-  identity** (feature + `DimParameter` role), not by name-string or position. "Drop this
-  hole's location dim" is intent; the solver simply never receives that candidate.
-- **Emphasis** — rank / anchor, already ADR 0012 (`pin`, `priority`). Folded in here as
-  the third face of the same layer so the three read as one vocabulary.
+- **Surface & drop** — the emitter emits one `dim(...)` line per dimension the drawing
+  carries, so the script mirrors the sheet; commenting a line out **suppresses** that
+  dimension, keyed by its semantic `feature + DimParameter role` identity, never a page
+  name or coordinate.
+- **Add** — a `dim(...)` line for a measurement the planner would not auto-produce (a
+  span between two features, a pattern pitch, a wall thickness, an angle) enters the
+  shared solve as a new `CorridorCandidate`.
+- **Emphasise** — `.pin()` / `.priority()` chained onto a `dim(...)` line ranks or
+  anchors it in the solve (already ADR 0012), without fixing a coordinate.
 
-The `.thread` / `.finish` pattern is the template and the invariant: a dimensioning
-intent enriches or edits the dimension *set* the drawing carries, and the corridor solve
-still decides ordering, dedup, priority-drop, and position. **Intent selects; the engine
-places.**
+The `.thread` / `.finish` aspect is the template and the invariant: a `dim` line edits
+the dimension *set* the drawing carries; the corridor solve still decides ordering,
+dedup, priority-drop, and position. **A dimension line references; the engine places.**
+
+Only *dimensions* are surfaced as lines. Low-level furniture the engine derives — centre
+marks, section arrows, hatching, the NTS caption — stays automatic; surfacing each as a
+line would explode the mirror without adding any editable intent.
 
 ### Constraints this forces (the honest edges)
 
@@ -102,22 +112,76 @@ places.**
   intents the engine can faithfully re-solve, never decorative lines that re-run cannot
   reproduce (the same fidelity contract `emit_sheet_script` holds for features).
 
+## Worked example — the mounting plate
+
+An 80 × 50 × 8 plate: a central ⌀20 bore with a ⌀32 × 1.5 spotface (which auto-triggers
+section A–A) and four ⌀5 corner holes. Today `emit_sheet_script` writes the four
+*features* and leaves the dimensions, views, and section as comments. Under this ADR the
+emitter mirrors the drawing as **referential dimension lines** — every callout on the
+sheet is one line, and no line restates a number:
+
+```python
+from draftwright import Sheet
+from build123d import import_step
+
+part = import_step("plate.step")
+sheet = Sheet(part, title="PLATE", number="DWG-001")
+
+# ── Features — geometry; each size lives here, once ────────────────────────────
+corners = sheet.hole(diameter=5,  at=(-32, -18, 4), axis="z", count=4, members=[...])
+bore    = sheet.hole(diameter=20, at=(0, 0, 4), axis="z").spotface(diameter=32, depth=1.5)
+env     = sheet.envelope()                     # the overall bounding geometry
+
+# ── Dimensions — each REFERENCES a feature; no numbers restated ────────────────
+sheet.dim(bore,    "diameter")    # ⌀20 THRU ⌴ ⌀32 ↓ 1.5   (read off `bore`)
+sheet.dim(corners, "diameter")    # 4× ⌀5 THRU              (read off `corners`)
+sheet.dim(corners, "location")    # the hole-location ladder      ← comment out to drop
+sheet.dim(bore,    "location")    # bore on centre
+sheet.dim(env,     "width")       # 80    (read off the bbox)
+sheet.dim(env,     "depth")       # 50
+sheet.dim(env,     "height")      # 8     (thickness)
+# sheet.dim(corners, "pitch")     # ← uncomment to ADD a 64 × 36 grid-pitch dim
+
+# ── Views & section — editable lines; the engine still packs them ──────────────
+sheet.view("front"); sheet.view("plan"); sheet.view("side"); sheet.view("iso")
+sheet.section("A-A", through=bore)    # cut the spotfaced bore (else auto-triggered)
+
+sheet.export("plate")
+```
+
+Reading it against the rendered sheet:
+
+- **`⌀20` appears once.** The number lives on the `bore` feature line; `sheet.dim(bore,
+  "diameter")` only says *show `bore`'s diameter callout*. Change `diameter=20` → `25`
+  (or edit the build123d object, for a live part) and the callout follows — there is no
+  second copy to keep in sync.
+- **Dropping a dimension is not dropping the hole.** Comment out `sheet.dim(corners,
+  "location")` and the location ladder vanishes; the four ⌀5 *circles* stay, because they
+  are geometry projected from the part, not annotations. Geometry and dimensions are
+  separate layers — only editing the part removes a hole.
+- **The commented `pitch` line is the additive case** — a measurement the planner did not
+  place, added by reference and still ordered / placed by the corridor solve.
+
+The A/B "features imply dimensions" vs "every dimension is a line" fork explored during
+design collapses here: the referential form gives the completeness of the second with the
+single-source-of-truth of the first.
+
 ## Consequences
 
-- A new intent vocabulary on `Sheet` / `Drawing` (verb names deferred to the roadmap):
-  additive measurement declarations (feature/face-referenced), and a suppression verb /
-  per-feature `.no_location()`-style aspect. `pin` / `priority` are re-documented as the
-  emphasis face of the same layer.
+- One referential `sheet.dim(<feature>, <role>)` verb on `Sheet` / `Drawing` (exact role
+  names deferred to the roadmap), reading its value from the referenced geometry and
+  carrying none itself; `.pin()` / `.priority()` chain onto it as the emphasis face.
+  Suppression needs no separate verb — a surfaced `dim` line commented out is the drop.
 - `plan_dimensions` (ADR 0015) grows an intent input: declared additive measurements join
   the planned `DimensionGroup`s; declared suppressions filter them. The corridor solve
   (ADR 0014) is unchanged — it still receives one candidate population per strip.
 - `intents.py` (ADR 0012) is the recording home; `Drawing.finalize()` /
   `_PASS_SEQUENCE` the drain. Additive intent lands there first; suppression follows the
   #426/#707 recompose.
-- `sheet_emit` gains an intent-mirroring pass: after the feature basis, emit the
-  dimensioning intents (and views / sections, ADR-0016-adjacent config work) the drawing
-  carries — each a commentable, editable *intent* line, with the auto-derived remainder
-  still produced by the engine on re-run.
+- `sheet_emit` gains a dimension-mirroring pass: after the feature basis, emit one
+  referential `dim(...)` line per dimension the drawing carries (plus views / sections,
+  ADR-0016-adjacent config work) — each commentable and editable, none restating a number,
+  with low-level furniture still produced automatically by the engine on re-run.
 - Extends ADR 0011 (declare features) to declare *dimensioning intent*; extends ADR 0012
   (edit one dimension) to declaring the dimension *set*; consumes ADR 0015 (planner) and
   ADR 0014 (solve); fulfils ADR 0001 §2. Does **not** reintroduce the ADR 0001 hardcoded
@@ -132,21 +196,25 @@ places.**
 2. **Semantic identity for auto dimensions.** Expose a stable `(feature, role)` handle for
    planned dimensions (ADR 0010 provenance + ADR 0015 roles) so intent can *reference* an
    auto dimension.
-3. **Suppressive intent.** `undim` / `.no_location()` filtering `plan_dimensions` output
-   by that handle.
+3. **Suppression by omission.** A surfaced referential `dim` line, when commented out,
+   filters that `(feature, role)` from `plan_dimensions` output — no separate verb.
 4. **Full recompose (#426/#707).** Reconstruct the automatic population at finalize and
-   co-solve with declared intents, making suppression / emphasis honest and script/direct
-   output convergent.
-5. **Emitter intent-mirror.** Emit the round-trippable dimensioning intents (with views /
-   sections) as editable lines; keep the self-describing comment as the floor for the rest.
+   co-solve with declared dimensions, making suppression / emphasis honest and
+   script/direct output convergent.
+5. **Emitter dimension-mirror.** Emit one round-trippable referential `dim(...)` line per
+   placed dimension (with views / sections); keep the self-describing comment as the floor
+   for anything not yet mirrorable.
 
 ## Open questions
 
-- The additive-intent verb surface: which semantic measurements to support first (span,
-  pitch, thickness, angle, radius, overall), and how a two-feature reference reads
-  fluently (`sheet.measure(a, b)` vs a feature-handle method).
-- Whether suppression is a standalone verb (`sheet.undim(hole)`) or a per-feature aspect
-  (`sheet.hole(...).no_location()`), and how it composes with `of(...)` on detected models.
-- Guards to add when this lands: a fidelity test that an emitted intent re-solves to the
-  same dimension (mirroring `test_sheet_emit` parity), and an audit that intent carries no
-  page geometry (the scale-independence invariant).
+- The `role` vocabulary for `sheet.dim(feature, role)`: which measurements to support
+  first (`"diameter"`, `"location"`, `"pitch"`, `"width"`/`"depth"`/`"height"`, `"angle"`,
+  `"radius"`), and how a *two-feature* span reads fluently — `sheet.dim(a, b)` /
+  `sheet.dim((a, b), "span")` vs a feature-handle method.
+- How `sheet.dim(...)` composes with `of(...)` on a detected model (referencing an
+  auto-detected feature by object/index to drop or emphasise its dimension), and whether a
+  bare `sheet.dim(feature)` (no role) means "all of that feature's dimensions".
+- Guards to add when this lands: a fidelity test that an emitted `dim` line re-solves to
+  the same dimension (mirroring `test_sheet_emit` parity), and an audit that a `dim` line
+  carries no number or page geometry (the reference-not-restate / scale-independence
+  invariant).
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -4,8 +4,8 @@
 - **Date:** 2026-07-26
 - **Deciders:** Paul Fremantle (pzfreo)
 
-> **Scheduling reality.** Only phase 1 (additive intent) is reachable on today's
-> machinery. Phases 3–5 — suppression, honest reconciliation, and the emitter's
+> **Scheduling reality.** Only phase 1 (augmenting intent, `add_dimension`) is reachable on
+> today's machinery. Phases 3–5 — suppression, honest reconciliation, and the emitter's
 > dimension mirror — sit behind the global recompose (#426/#707), the longest-open
 > item on the roadmap. Accepting this ADR is therefore also a decision to schedule
 > that recompose; it is the prerequisite, not an adjacent nicety.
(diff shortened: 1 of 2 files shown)


pzfreo:
## Architectural review

The core direction is strong and fits draftwright’s architecture well. The central rule — **a dimension line references; the engine places** — preserves the right boundaries: geometry remains the source of values, users declare manufacturing intent rather than page coordinates, the planner owns dimension policy, and ADR 0014 still owns placement. Emitting planner intent rather than placed annotations is especially important: solver pressure must not silently rewrite version-controlled source.

I would support accepting this direction, but the ADR needs three load-bearing clarifications first.

### 1. Explicit dimension-set mode is required for backward compatibility

The ADR says absence of a `dimension(...)` line unambiguously means “suppressed.” Existing generated and hand-authored `Sheet` scripts contain no referential dimension lines and rely on automatic planning. Without an explicit mode boundary, the engine cannot distinguish:

- a legacy script requesting normal automatic dimensions; and
- a new explicit script intentionally suppressing every dimension.

The end state needs a sentinel such as `Sheet(..., dimensioning="explicit")`, a dimension-set declaration, or an equivalent emitted marker. Existing scripts should remain automatic; generated intent-mirror scripts can opt into explicit selection, where omission becomes meaningful.

### 2. Phase 5 contradicts the emitter invariant

“The script records intent, not what got placed” correctly requires emission from the planner’s intent set so solver-dropped dimensions keep their source lines. The phase-5 wording currently says:

> Emit one ... line per placed dimension

That should be **per planned dimension intent** (or equivalent wording). “Placed” reintroduces the exact layout-dependent source churn the ADR rejects.

### 3. Define the dimension-intent handle contract

The decision promises `.pin()` / `.priority()` chained from `sheet.dimension(...)`, but the transitional signature sketch returns `Sheet`. The ADR should state that the referential verb returns a dedicated dimension-intent handle (or explicitly choose keyword arguments instead) and define how that handle participates in semantic identity, suppression, and the planner input.

### Scope recommendation

The referential dimension layer, explicit dimension-set semantics, planner/emitter mirror, and `dimension` → `measured_dimension` migration belong together. The proposed `sheet.view(...)` surface and redesigned named `sheet.section(...)` form are adjacent but separate decisions. I recommend removing them from the normative ADR 0016 end state, or clearly deferring them to a follow-up ADR, so accepting 0016 does not implicitly accept an unresolved view/section API.

Overall: **excellent architectural direction, not yet a merge-ready decision record**. Once explicit-mode compatibility, planned-vs-placed wording, and the intent-handle contract are settled, I would support acceptance.

pzfreo:
## Follow-up: a cleaner mode boundary

After discussing the explicit-set ambiguity, I think there is a cleaner final API than either a constructor keyword or a dimension-set context manager:

```python
# Planner-selected dimension set
dwg = Sheet(part).auto_dimensions().build()

# Authored dimension set
sheet = Sheet(part)
sheet.dimension(bore, "diameter")
sheet.dimension(bore, "location")
dwg = sheet.build()
```

The semantic rule would be:

- `auto_dimensions()` explicitly requests the planner-selected dimension set.
- `dimension(...)` declarations form the complete authored set.
- Neither one produces an intentionally undimensioned drawing.
- Mixing automatic and authored-set declarations fails clearly unless a separate additive operation is deliberately introduced.

```python
sheet.auto_dimensions()
sheet.dimension(bore, "diameter")
# ValueError: cannot mix automatic and authored dimension sets
```

If augmenting the automatic plan is needed, it should read differently—e.g. `add_dimension(...)`—so a reader can distinguish a complete authored set from an automatic plan plus one extra request.

This is a breaking change from today’s implicit automatic behavior:

```python
Sheet(part).build()                   # geometry/views, no dimensions
Sheet(part).auto_dimensions().build() # planner-selected dimensions
Sheet(part).dimension(...).build()    # authored dimension set
```

Given the package is alpha and has few users, I think that break is justified by the resulting API clarity. It avoids a hidden `dimensioning="explicit"` mode, avoids a parallel surface, handles the empty set naturally, and makes omission unambiguous: dimensions absent from the authored path were never requested because automatic planning was never enabled.

Under this design the ADR can describe two explicit sources of dimension intent—planner-generated and author-declared—followed by the same shared planner/corridor solve. I prefer this over my earlier constructor-keyword suggestion.

pzfreo:
All four points addressed in 0445556, with one amendment to the mode boundary.

**1. Explicit dimension-set mode** — took the follow-up design over the constructor keyword. `auto_dimensions()` requests the planner set; `dimension(...)` declarations *are* the complete authored set; mixing raises; `add_dimension(...)` reserved for augmenting the automatic plan. New section "Two explicit sources of dimension intent".

**Amendment: a build that requested neither also raises**, rather than returning an undimensioned drawing.

```python
Sheet(part).build()
# ValueError: no dimension set requested — call auto_dimensions() for the
# planner-selected set, or declare dimensions with dimension(...)
```

A silently undimensioned sheet is plausible-looking and unbuildable — the "clean but incomplete drawing" the roadmap operating model ranks as worse than a visible failure, and precisely the case the #632 completeness lint exists to stop passing. Raising is consistent with #631 (boss + coincident step raises rather than dropping a height) and #630 (unplaceable detail reports rather than no-ops). It costs nothing in practice: the emitter always writes one of the two forms, so a generated script never hits it. It also keeps the rest of your design intact — no hidden mode, no parallel surface, empty set handled naturally, omission unambiguous.

**2. Phase 5 wording** — confirmed as a live self-contradiction (phase 5 said "per placed dimension" while the invariant section said the opposite). Now "per **planned dimension intent**", with the reason inline.

**3. Intent-handle contract** — the referential verb returns a `DimensionIntent` handle, not `Sheet`. It carries the `(feature, role)` identity, anchors ADR 0010 provenance, and is what `.pin()` / `.priority()` chain from. Grounded in the existing façade pattern: `hole()` → `_Hole`, `diameter()` → `_Dim`, while `chamfer()` / `fillet()` return `Sheet`. The transitional overload's dual return type is recorded as a wart that ends at 0.4.0.

**4. Scope** — the view / section surface now has its own "Out of scope" section and an illustrative-only marker in the worked example, so accepting 0016 commits to the dimension layer alone. Also noted that `sheet.section` is a *reshape* of the verb shipped in v0.3.9 (#847), not a blank sheet, so it needs its own overload-versus-rename call in the follow-up ADR.

**Phasing** — phase 3 is re-scoped to land the set boundary *before* suppression, since omission has no meaning until a set is declared complete. The breaking change ships with the `measured_dimension` rename so callers migrate once.

pzfreo:
## Second-round architectural review

The first review’s four concerns are now addressed convincingly: the explicit source boundary, planned-vs-placed wording, `DimensionIntent` return contract, and view/section scope are all much clearer. Two new load-bearing issues emerge from the revised model.

### 1. `(feature, role)` is not a unique dimension identity

The existing IR permits multiple `DimParameter`s with the same role on one feature. For example:

- a blind hole can have both diameter and depth with role `"bore"`;
- a counterbore can have both diameter and depth with role `"counterbore"`;
- a grid pattern can have two pitch parameters with the same kind and role.

That means even `(feature, kind, role)` is insufficient for the two grid pitches. The ADR needs a stable parameter identity—such as `(feature_id, parameter_id)`—with a semantic discriminator like axis/slot or a planner-assigned parameter key. This is foundational because the same identity drives handles, suppression, de-duplication, emitter mirroring, and provenance.

### 2. “Add” conflicts with complete authored-set semantics and the phased rollout

The ADR says `dimension(...)` can add a measurement the planner would not otherwise produce, but the revised rule also says `dimension(...)` declarations form the complete authored set and cannot be mixed with `auto_dimensions()`.

Under complete-set semantics, `sheet.dimension(pattern, "pitch")` means “emit only pitch,” not “run the automatic plan and add pitch.” Yet `add_dimension(...)` is only reserved/unspecified, while phase 1 is still described as additive intent.

Please either:

- specify and land `add_dimension(...)` in phase 1, or
- remove additive behavior from `dimension(...)` and rename/reorder phase 1 accordingly.

The same `dimension(...)` call should not change meaning from additive before phase 3 to full-set after phase 3.

### 3. Narrow the complete-mirror claim

The phrase “complete per-dimension mirror” still overstates the contract. The ADR now acknowledges that grouped passes may suppress dimensions set-wise, some inter-feature dimensions are not nameable, and unsupported dimensions fall to the comment floor.

I suggest promising a **mirror of round-trippable, semantically identified planner intent**, followed by an explicit list of gaps. That is precise without constraining future planner internals unnecessarily.

Overall, the core direction remains excellent and substantially improved. I would resolve the parameter-identity model and the additive/full-set semantic conflict before accepting the ADR, because both affect the public API and every later implementation phase.

pzfreo:
## Recommended dimension identity model

I recommend a stable semantic parameter key scoped to a feature, rather than `(feature, role)`, `(feature, kind, role)`, list position, or a UUID.

```python
@dataclass(frozen=True)
class DimensionId:
    feature_id: FeatureId
    parameter_id: ParameterId


@dataclass(frozen=True)
class DimParameter:
    id: ParameterId
    kind: DimensionKind
    role: str
    value: float
```

Example parameter IDs:

```python
ParameterId("bore.diameter")
ParameterId("bore.depth")
ParameterId("counterbore.diameter")
ParameterId("counterbore.depth")
ParameterId("grid.pitch.x")
ParameterId("grid.pitch.y")
```

This permits parameters to retain the same grouping/presentation role while still having distinct identities:

```python
DimParameter(
    id=ParameterId("bore.diameter"),
    kind="diameter",
    role="bore",
    value=10.0,
)

DimParameter(
    id=ParameterId("bore.depth"),
    kind="linear",
    role="bore",
    value=18.0,
)
```

The public API can remain clean and need not expose these keys routinely:

```python
sheet.dimension(hole, "diameter")
sheet.dimension(hole, "depth")
sheet.dimension(pattern, "pitch", axis="x")
sheet.dimension(pattern, "pitch", axis="y")
```

Those selectors resolve internally to identities such as:

```python
DimensionId(hole.id, ParameterId("bore.diameter"))
DimensionId(hole.id, ParameterId("bore.depth"))
DimensionId(pattern.id, ParameterId("grid.pitch.x"))
DimensionId(pattern.id, ParameterId("grid.pitch.y"))
```

A returned intent handle should carry that canonical identity:

```python
@dataclass(frozen=True)
class DimensionIntent:
    id: DimensionId
    priority: int = 0
    pinned: bool = False
```

The planner, authored-intent layer, suppression, de-duplication, provenance, and emitter mirror can then all use the same key:

```python
planned: dict[DimensionId, PlannedDimension] = {
    dimension.id: dimension
    for dimension in measurement_planner.plan(model)
}
```

For inter-feature measurements, I would use a separate semantic identity rather than forcing them into a feature parameter:

```python
@dataclass(frozen=True)
class RelationDimensionId:
    relation: Literal["distance", "angle", "offset"]
    subjects: tuple[FeatureId, ...]
    axis: Literal["x", "y", "z"] | None = None
```

The governing principle is that identity should describe the engineering measurement, not the annotation or planning pass that happened to produce it. Roles remain useful metadata, but should not be identity. Semantic keys are stable across reordering and planner changes, readable in diagnostics and snapshots, and avoid the opacity of random UUIDs.

pzfreo:
All three second-round points addressed in the pushed commit, with one refinement to the identity model and one substantive divergence from the recommended shape.

## 1. Parameter identity — confirmed, and it sorts into three tiers, not one

Checked against `model/ir.py`. The collision is real and there are more instances than the review named, but they are **not the same problem** and lumping them would produce the wrong fix:

- **Tier 1 — same role, different kind.** Blind hole `("diameter","bore")` + `("depth","bore")`; counterbore, spotface and countersink each a diameter+depth (or angle) pair. `kind` resolves these.
- **Tier 2 — same role *and* kind, semantically distinct.** A grid emits two `("length","grid_pitch",…)` — row and column. Your case; nothing in `(kind, role)` separates them; forces a third component.
- **Tier 3 — same role and kind, deliberately *one* addressable thing.** `FaceLevels` emits one `step_height` per level and one `step_position` per shoulder; `Rotational` one `bore` per concentric bore. `ir.py` says so at the source: *"correlated SETS routed as a whole … a single `role=` intent rebuilds the whole ladder."* Here the role already **is** the identity, and minting per-member keys would contradict the ADR's own grouped-drop edge.

So identity is per **addressable dimension** — the set for a correlated pass, the parameter for an independent one. `DimensionId(feature, parameter)` adopted as recommended.

**One divergence: `ParameterId` is derived, not a hand-authored field.** Composing it as `role` + `kind` + optional discriminator, rather than an `id=` literal on each of the forty `DimParameter(...)` sites. An `id=` that can disagree with the `role=` beside it is a new class of silent bug, and only tier 2 carries genuinely new data (`axis="x"`/`"y"`, today's sole instance). It's the same single-source argument the ADR makes about numbers. Everything else lands as recommended: semantic strings not UUIDs, list position rejected, `axis=` selector, `RelationDimensionId` for inter-feature.

**`FeatureId` is explicitly not minted here.** It doesn't exist today — the plan carries `DimensionGroup.feature` by object, and within a script run the variable *is* the reference, so `sheet.dimension(bore, "diameter")` needs no key. A durable one is only needed where an intent must survive re-detection, which is the `of(...)` question already open. Recorded as deferred rather than silently assumed.

Rotational bores flagged **provisional** — they're the #754 residual coverage debt, so set-vs-member is settled by that migration, not by this ADR.

## 2. Additive vs complete-set — confirmed live, fixed by specifying `add_dimension`

Took your first option. `add_dimension(...)` is now **specified**: same referential contract, same `DimensionIntent` handle, augments the planner's set, and **requires `auto_dimensions()`** (there is nothing to augment in an authored set, where every line is already `dimension(...)`). Overlap with the plan is idempotent via the identity key, not an error.

Phase 1 renamed to lead with it, and the phasing section now carries the rule explicitly: *a verb never changes meaning between phases*. `dimension(...)` does not exist until phase 3 lands the set boundary that gives it its meaning. The "Add" bullet in the Decision now names the second verb rather than overloading the first.

## 3. Mirror claim — narrowed

Now "a mirror of **round-trippable, semantically identified planner intent**", with a table of the five gaps: correlated sets (per member), inter-feature spans pending the relation selector, materialized AP242 PMI, low-level furniture, and anything the emitter can't re-solve. Two shrink as the identity model grows; three are decisions and stay. The property actually promised is that *within the identified set, presence and absence each mean something exact* — which is all suppression-by-omission needs.

## Bonus the identity model paid for

The grouped-drop wart resolves rather than being documented around. Because a correlated set is one identity, the declarative mirror emits **one** line for a ladder and commenting it drops the set — so "comment some members and the group still redraws anyway" stops being a caveat the emitted comment has to carry. It does mean the `--style imperative` retirement gate is **coverage** parity, not line-for-line parity; that's now stated.

## Guards added to the deferred list

A `ParameterId` uniqueness audit across the IR (the check that would have caught the grid-pitch collision before this review did), an ambiguous-selector-raises test (`dimension(pattern, "pitch")` with no `axis=`), and a re-detection stability test. Plus a third source error to cover: `add_dimension` without `auto_dimensions()`.

pzfreo:
## Third-round architectural review

The latest revision resolves the earlier headline concerns convincingly: explicit set sources, additive semantics, planned-versus-placed emission, bounded mirroring, migration behavior, and view/section scope are now substantially clearer. I still see three architectural issues in the identity model.

### 1. Correlated-set identity contradicts the proposed uniqueness guard

The ADR says multiple `step_height`, `step_position`, or rotational-bore parameters deliberately share one `ParameterId` because the set is the addressable unit. It later requires an audit that no feature produces two parameters with the same `ParameterId`. That audit would reject the correlated sets the design intentionally permits.

The model needs a first-class addressable unit, for example:

```python
AddressableDimension(
    id=ParameterId("step_height"),
    parameters=(height_1, height_2, height_3),
)
```

`DimensionId` should identify this unit, while its member parameters can retain distinct internal identities if needed. Otherwise dictionaries, provenance, suppression, and uniqueness checks cannot consistently distinguish an accidental collision from an intentional group.

### 2. The compound-hole example does not match parameter identity

The ADR maps:

```python
sheet.dimension(bore, "diameter")
# -> DimensionId(bore, "bore.diameter")
```

but the worked example says that declaration emits the complete compound callout:

```text
⌀20 THRU ⌴ ⌀32 ↓ 1.5
```

That line addresses multiple measurements—not only `bore.diameter`. The current IR and planner explicitly group a hole's bore, depth, counterbore, spotface, and countersink parameters into one callout.

The ADR needs to choose whether the addressable unit is:

- an individual engineering measurement, permitting separate suppression; or
- a compound callout/group, with an identity such as `hole.callout`.

It currently claims the former identity while demonstrating the latter behavior. A first-class `AddressableDimension`/group model would resolve this case as well as the ladders.

### 3. Phase 1 depends on phase 2

`add_dimension()` is specified as returning a handle carrying `DimensionId` and being idempotent when it overlaps the automatic plan. But phase 1 ships `add_dimension()`, while semantic identity does not land until phase 2. Phase 1 therefore cannot provide the verb's final stated contract.

Please land identity first, or combine phases 1 and 2.

### Smaller corrections

- Phase 1 includes “span-between-features,” but the relation selector remains open. Restrict phase 1 to feature-addressable measurements unless the relation syntax is decided first.
- `RelationDimensionId` cannot canonically sort subjects for every relation. Distance is symmetric, but signed offsets and oriented angles may depend on subject order. Normalization should be relation-specific.
- The `add_dimension(pattern, "pitch")` example omits the `axis=` discriminator even though the identity section correctly says that exact call must raise for a grid.

Overall, the ADR is close, but I would resolve the addressable-unit/compound-callout model before acceptance. It is the central abstraction beneath identity, handles, suppression, deduplication, and emission; leaving it implicit would push the hardest decision into implementation.

pzfreo:
All six points taken. Two were live self-contradictions in the identity model I added last round, and both are fixed by the abstraction you named.

## 1. The uniqueness audit rejected the design — confirmed

Tier 3 said correlated sets deliberately share one `ParameterId`; the deferred guard said no feature may yield two parameters with the same `ParameterId`. Those cannot both hold. **`AddressableDimension(id, parameters)` adopted as a first-class model type** — usually one parameter, N for a ladder — with the grouping **declared by the planner, never inferred from a key collision**.

That also closes a failure mode of the derived key I had missed: a grid pitch missing its `axis=` discriminator would otherwise look like an intentional correlated set rather than an error. Deriving the key is only safe *because* grouping is declared separately. The guard is now phrased over addressable units rather than raw `DimParameter`s, and still catches the grid-pitch case — two ungrouped pitches are two units.

## 2. The compound-hole example — the sharper catch, and you are right

Verified in `from_model.hole_callout_spec`: a hole's bore ⌀, depth, counterbore, spotface and countersink parameters collapse into **one** `HoleCallout`. So `sheet.dimension(bore, "diameter")` cannot both key `bore.diameter` and emit `⌀20 THRU ⌴ ⌀32 ↓ 1.5`. The example was demonstrating group behaviour under a parameter identity.

I did not take either horn of the fork, because checking the renderer showed the two cases are genuinely different rather than two spellings of one thing:

- **Correlated set** — one identity, N parameters, no member addressable. Grouping at the **identity** layer.
- **Compound callout** — N identities sharing one leader, each separately suppressible. Grouping at the **render** layer. `callout_from_spec` already takes `None` per segment, so dropping the two spotface lines yields `⌀20 THRU` — that is mechanically available today, not a new capability.

Keying identity on the callout (`hole.callout`) would contradict this ADR's own governing principle — identity describes the engineering measurement, not the annotation that renders it. So the model carries both notions explicitly, which is a third argument for making the addressable unit first-class: without it there is nowhere to state the difference.

One real asymmetry recorded rather than glossed: the bore ⌀ is the callout's **head** (`hole_callout_spec` returns nothing without it), so suppressing it suppresses the whole callout rather than leaving a headless one. The worked example is now three lines with the aggregation stated.

## 3. Phase 1 depended on phase 2 — confirmed

`add_dimension` returns a handle carrying a `DimensionId` and is idempotent against the plan; both need identity. **Merged into one phase, identity first.** The phasing note now carries both rules rather than just the one: *a verb never changes meaning between phases*, and *no phase ships a verb whose stated contract it cannot yet honour*. Downstream phases renumbered.

## Smaller corrections — all three taken

- **Phase 1 restricted to feature-addressable measurements.** "Span-between-features" removed; it would have meant guessing the relation syntax that is still open.
- **`RelationDimensionId` normalization is relation-specific.** You are right that a blanket sort is wrong — symmetric distance canonicalizes by sort, but reordering the subjects of a signed offset or an oriented angle negates the measurement, so sorting would key two different dimensions as one. Each relation now declares its own normalization.
- **`add_dimension(pattern, "pitch")` gained the `axis=`** it needed; the example was violating the rule stated two sections above it.

pzfreo:
## Fourth-round architectural review

The latest commit addresses all six previous points well: addressable units are now first-class, uniqueness is correctly defined over those units, compound callouts are separated from correlated sets, relation normalization is relation-specific, identity and augmentation land together, and the examples/scope are corrected. Two implementation-level contradictions remain in the decision record.

### 1. Suppressing bore depth can falsely turn a blind hole into `THRU`

The ADR now treats bore diameter and bore depth as independently addressable dimensions. However, the current `from_model.hole_callout_spec()` derives:

```python
through = depth is None
```

If an authored set contains `bore.diameter` but omits `bore.depth`, filtering the planner parameters removes the depth and the renderer emits `THRU`. A blind hole has then become a through-hole on the drawing, which is materially incorrect.

Selection/visibility must therefore be separate from engineering data. The complete feature parameters should remain available to the compound-callout builder, with an explicit visibility mask controlling which segments are displayed. Alternatively, unsupported partial selections must raise; simply filtering `DimensionGroup.dims` is unsafe.

The same dependency needs a defined contract for the callout head. The ADR notes that `hole_callout_spec()` returns `None` without the bore diameter, but does not say what happens when spotface/counterbore intents remain present: do they raise, lint-and-drop, or implicitly restore the required head? Calling every segment “separately suppressible” overstates the behavior until this dependency is specified.

### 2. `AddressableDimension` is defined at the wrong/unclear pipeline boundary

The proposed type contains raw parameters:

```python
@dataclass(frozen=True)
class AddressableDimension:
    id: ParameterId
    parameters: tuple[DimParameter, ...]
```

But the ADR calls it a planner-level unit, while the existing planner enriches every `DimParameter` into a `PlannedDimension` carrying `convention`, `suppressed`, `reason`, `datum`, and provenance. Grouping only raw parameters either loses that metadata or requires another undocumented parallel grouping after planning.

Please define the actual pipeline contract, for example:

```python
# IR-side grouping, if needed
ParameterGroup(id, parameters)

# Planner output consumed by reconciliation/emission
PlannedAddressableDimension(
    id,
    members: tuple[PlannedDimension, ...],
)
```

—or one planner-output `AddressableDimension` whose members are `PlannedDimension`s. The ADR should state how the existing `DimensionGroup.dims` changes and which concrete type suppression, provenance, deduplication, and the emitter consume.

### 3. Stale wording

The constraints section still says:

> Augmenting intent needs no such key

Phase 1 now correctly establishes the opposite: `add_dimension()` needs identity for its handle and idempotence against the plan. The distinction is that suppression additionally needs identity stability across recomposition/re-detection.

Overall, this is now close. I would resolve the false-`THRU` path and make the addressable-unit insertion point explicit before accepting the ADR; both otherwise defer correctness-critical behavior to implementation.

pzfreo:
All three taken. Point 1 is the best catch in four rounds — I verified it in the source and it is a live defect today, not just a hazard this ADR would introduce.

## 1. The false `THRU` — confirmed, and it is already broken

`from_model.hole_callout_spec` line 129: `"through": depth is None`, and `ir.py` line 142 only emits the depth parameter `if not self.through`. So the presence of a parameter is the *sole* carrier of a manufacturing fact. Filtering it turns a blind hole into a through hole on a print someone quotes from.

I did not take either of your two options, because checking `ir.py` showed a simpler root cause: **`HoleFeature.through` is a first-class IR field** (line 121) and `DimensionGroup` already carries the feature. The renderer is re-deriving something it can just read. That makes it a bug independent of ADR 0016 — a hand-built `DimensionGroup` misreports today — and it generalises into the rule the ADR actually needs:

> No renderer may infer an engineering fact from the presence or absence of a dimension parameter. Parameters carry values *for display*; facts live on the feature.

Your "selection must be separate from engineering data" conclusion is right and now stated, with the mechanism identified rather than proposed: **`PlannedDimension` already carries `suppressed`/`reason`, and 13 render sites already honour it.** So suppression **marks, it does not filter** — the group keeps its full engineering data. The compound-callout path (`_first`/`hole_callout_spec`) is the *one* reader that ignores the flag; closing that gap is now explicitly part of the suppression phase rather than left to implementation discretion. No new visibility-mask type needed.

**Head dependency, specified.** Suppressing the bore ⌀ while spotface/counterbore intents remain now **raises**. Not lint-and-drop (silently discards authored intent) and not implicit restore (makes the script say something it does not say). "Every segment separately suppressible" is corrected to hold *except the head*, declared by the addressable unit rather than scattered in the renderer.

## 2. Wrong pipeline boundary — confirmed

`DimensionGroup.dims` is `tuple[PlannedDimension, ...]`, and `PlannedDimension` carries `convention`/`suppressed`/`reason`/`datum`/`feature`. Grouping raw `DimParameter`s loses exactly the metadata suppression and provenance need. `AddressableDimension.members` is now `tuple[PlannedDimension, ...]` and `DimensionGroup.dims` is retyped to `tuple[AddressableDimension, ...]`.

**One type, not your two.** I dropped the IR-side `ParameterGroup`: grouping is a *planner* decision by the rule we agreed last round, so an IR-side group has nothing to hold and `Feature.parameters()` should keep returning a flat list. I also re-checked the no-new-type alternative — a `unit_id` field on `PlannedDimension`, units being the dims that share it — and rejected it for the round-3 reason: it reintroduces a state where you cannot distinguish a declared group from an accidental key coincidence. A container makes arity structural.

Migration cost named honestly in the ADR rather than discovered later: retyping `dims` touches ~25 read sites across `from_model.py`/`holes.py`/`compose.py`, nearly all `next(pd for pd in g.dims if …)`. A flattening accessor keeps it mechanical.

## 3. Stale wording — fixed

Corrected to: both intents need the key (`add_dimension` for its handle and idempotence); suppressive intent *additionally* needs identity stable across re-detection and recomposition — which is the actual reason identity ships with the augmenting verb while suppression waits for the set boundary.

---

**Follow-up worth filing separately:** the `through = depth is None` derivation should read `feat.through` regardless of whether this ADR is accepted.

Issue #867, opened 2026-07-27
Epic: implement ADR 0016 — declared dimensioning intent


pzfreo:
Implement [ADR 0016 — Declared dimensioning intent](../blob/main/docs/adr/0016-declared-dimensioning-intent.md) (merged in #866): `sheet.dimension(<feature>, <role>)` as a *referential* verb that names a measurement and carries no number, with the engine still owning derivation and placement. **A dimension line references; the engine places.**

## Scope of this epic

PR 0 through PR 7 below. Deliberately **out of scope**, because each needs a decision this epic should not make:

- the emitter dimension-mirror (needs the view/section surface, which ADR 0016 explicitly leaves to a follow-up ADR)
- the redundancy lint
- retiring `--style imperative`

## Children

| # | Child | Gate / oracle |
|---|---|---|
| 0 | `hole_callout_spec` reads `feat.through` instead of deriving `depth is None` | canary must **fail** at the pre-fix commit |
| 1 | `ParameterId` derivation + `axis=` discriminator on grid pitch; uniqueness + stability audits | pure addition, zero behaviour change |
| 2 | `AddressableDimension` + retype `DimensionGroup.dims` + flattening accessor | `tests/test_refactor_golden.py` unchanged, **never re-blessed** |
| 3 | `DimensionId` + handle on planned dimensions, ADR 0010 provenance wiring | provenance audit stays green |
| 4 | `add_dimension(...)` + `DimensionIntent` + `.pin()`/`.priority()` + idempotence | overlap with the plan yields exactly one dimension |
| 5 | `measured_dimension` rename + transitional overload | generated AP242 scripts regenerate clean |
| 6 | Mandatory dimension source: the three errors + the breaking change | all three raise; emitter always writes a source line |
| 7 | Suppression **marks**, not filters; head-dependency raises | a blind hole never prints `THRU` under any suppression |
| 8 | Authored-set suppression by omission | **gated on the open question below** |

## Ordering correction to the ADR

ADR 0016 phase 1 ships `add_dimension(...)`, which "**requires** `auto_dimensions()`". So `auto_dimensions()` has to exist in phase 1 as an **optional** explicit request. Making it **mandatory** — the breaking change — is phase 2 (PR 6). The ADR reads as though both land together; they cannot.

## Open question to settle before PR 8

[...]

PR #877, opened 2026-07-27
fix(#868): read `through` off the hole feature, not off a missing depth param

### LINKED ISSUE #868: ADR 0016 PR0: hole callout must read feat.through, not infer it from a missing depth param
Child of #867.

`from_model.hole_callout_spec` derives `"through": depth is None`, and `HoleFeature.parameters()` only emits the depth parameter `if not self.through`. So **the presence of a dimension parameter is the sole carrier of a manufacturing fact.** Any consumer that filters the parameter list — which is exactly what ADR 0016's suppression does — turns a blind hole into a through hole on a print somebody quotes from.

`HoleFeature.through` is a first-class IR field and `DimensionGroup` already carries the feature (`feat.member` for a pattern), so the renderer is re-deriving what it can simply read.

This is a live defect today, independent of ADR 0016: a hand-built `DimensionGroup` already misreports.

**Also in this PR:** flip ADR 0016 `Proposed` → `Accepted`, and correct its stale blocker references (#426/#707/#754 are all closed — see #867's open question).

**Gate:** the canary test must **fail** on a worktree at the pre-fix commit, or it is theater.

pzfreo:
`hole_callout_spec` derived `"through": depth is None`, while
`HoleFeature.parameters()` emits a bore depth only for a blind hole. The
presence of a *dimension* was therefore the sole carrier of a *manufacturing
fact*: any consumer that filters the parameter list turns a blind hole into a
through hole on a print somebody quotes from.

`HoleFeature.through` is a first-class IR field and `DimensionGroup` already
carries the feature (`.member` for a pattern), so the renderer was re-deriving
what it can simply read.

This is live today — a hand-built `DimensionGroup` already misreports — and it
is a prerequisite for ADR 0016 suppression, which drops planned dimensions by
design. The general rule it establishes: a renderer may not infer an
engineering fact from the presence or absence of a dimension parameter.
Parameters carry values for display; facts live on the feature.

Both new tests are canaries: verified failing on a worktree at the pre-fix
commit (06f2b2e) and passing here.

Also flips ADR 0016 to Accepted and corrects three stale references it merged
with: #426/#661/#707 all closed 19-21 Jul and #754 on 22 Jul, so the
"phases 2-4 sit behind the global recompose" premise no longer holds as
stated. Since closed-as-umbrella is not capability-landed (#426's own closing
comment says the parity work remains observably incomplete, #743's suite is
skipped with xfails, and `drain_and_reconcile` does not reconstruct the
automatic candidate population), the blocker is recorded as an open question
on #867 rather than resolved either way. Notes the cheaper question inside it:
a Sheet script's authored set is known before `build()`, so phase 2 may need
no recompose at all.

Closes #868.

---

**Epic:** #867 (ADR 0016 implementation), PR 0 of 8.

**Verification**
- Canary validated: both new tests **fail** on a worktree at `06f2b2e` (pre-fix) and pass here.
- Full fast tier: 1667 passed.
- All four lint checks clean (ruff check, ruff format, mypy, codespell).

### DIFF
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -1,14 +1,25 @@
 # ADR 0016 — Declared dimensioning intent: capture what to measure, let the engine place it
 
-- **Status:** Proposed
-- **Date:** 2026-07-26
+- **Status:** Accepted
+- **Date:** 2026-07-26 (accepted 2026-07-27; implementation epic **#867**)
 - **Deciders:** Paul Fremantle (pzfreo)
 
-> **Scheduling reality.** Only phase 1 (identity + `add_dimension`) is reachable on
-> today's machinery. Phases 2–4 — suppression, honest reconciliation, and the emitter's
-> dimension mirror — sit behind the global recompose (#426/#707), the longest-open
-> item on the roadmap. Accepting this ADR is therefore also a decision to schedule
-> that recompose; it is the prerequisite, not an adjacent nicety.
+> **Scheduling reality (corrected 2026-07-27).** Phase 1 (identity + `add_dimension`)
+> is reachable on today's machinery. Phases 2–4 were written as sitting behind "the
+> global recompose (#426/#707), the longest-open item on the roadmap" — **that premise
+> was stale when this ADR merged.** #426 and #661 closed 2026-07-19, #707 on 07-21, all
+> as completed; there is no open recompose issue.
+>
+> Closed-as-umbrella is not the same as capability-landed, so the blocker is now an
+> open question rather than a stated fact (#867): #426's closing comment says the parity
+> work "remains observably incomplete", the #743 parity suite is skipped with strict
+> xfails underneath, and `drain_and_reconcile` solves registered corridors and reconciles
+> witness labels without reconstructing the automatic candidate population. There is also
+> a cheaper question inside it — **suppression by omission in a `Sheet` script happens
+> before `build()`**, when the set is known at plan time, so it may need no recompose at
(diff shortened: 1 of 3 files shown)


pzfreo:
## Adversarial review (Codex, 2 rounds)

**Round 1** — one substantive finding, [P2] docs:

> ADR 0016 still contradicts itself about the suppression prerequisite. The new scheduling note says pre-build `Sheet` suppression may require no recompose […] But the constraints section still concludes that "suppressive / full-mirror intent depends on #426/#707" […] the accepted implementation plan remains ambiguous and points implementers at closed prerequisites.

Correct, and it was my own half-applied correction: I flagged the constraint bullet's *opening* as open-status but left its closing sentence asserting the dependency flatly, and left phase 3 titled after the closed issues. Fixed in 6914ba7 by splitting the two paths explicitly everywhere suppression is discussed — **pre-build** (a `Sheet` script's authored set, known before the solve, plausibly needs no reconstruction) vs **post-build** (`Drawing.finalize()` dropping an automatic dimension, which does), with #867 as the open question rather than #426/#707 as a prerequisite.

No correctness findings against the code in either round.

**Round 2** — *"No substantive findings"*; the finding verified as genuinely resolved at all three sites.

**Note on review scope:** Codex could not execute the suite (local sandbox failed to initialize a namespace) and reviewed the pushed SHAs via the GitHub API instead. Test execution is covered here: full fast tier 1667 passed locally, plus CI.

**Separately found during my own pass:** #878 — the emit side loses blindness for a `through=False, depth=None` hole. Same rule, different site; deliberately not folded into this PR.

Issue #872, opened 2026-07-27
ADR 0016 PR4: add_dimension() + the DimensionIntent handle


pzfreo:
Child of #867.

`sheet.add_dimension(<feature>, <role>)` — a scale-independent augmenting measurement recorded on the model and entered as a `CorridorCandidate` alongside the planner's set. Reachable on today's solve; it is simply a new candidate.

- returns a **`DimensionIntent`** (not `Sheet`) so `.pin()`/`.priority()` chain (ADR 0012);
- carries no number — the value is read from the referenced geometry;
- **idempotent** against the plan via `DimensionId`: overlap is not an error;
- **feature-addressable measurements only.** Inter-feature spans wait for the relation selector, which ADR 0016 leaves open — shipping them here would mean guessing that syntax.

**Ordering note (see #867):** `add_dimension` requires `auto_dimensions()`, so `auto_dimensions()` ships here as an **optional** explicit request. Making it mandatory is PR6.

**Possible fold-in:** #563 (explicit view/side placement intent) may fall out of this naturally.

**Gate:** an idempotence test that a declaration overlapping the plan yields exactly one dimension.

Issue #870, opened 2026-07-27
ADR 0016 PR2: AddressableDimension as a first-class planner type; retype DimensionGroup.dims


pzfreo:
Child of #867. **The risky one.**

`AddressableDimension(id, members: tuple[PlannedDimension, ...])` — usually one member, N for a correlated set (the `step_height`/`step_position` ladders, rotational bores), with the grouping **declared by the planner, never inferred from a key collision**.

Members are `PlannedDimension`s, not raw `DimParameter`s, so `convention`/`suppressed`/`reason`/`datum`/provenance survive the grouping. `DimensionGroup.dims` is retyped accordingly; a flattening accessor keeps the ~25 read sites in `from_model.py`/`holes.py`/`compose.py` mechanical — readers that do not care about grouping never learn about it.

One type at one boundary: an IR-side parameter group would have nothing to hold, since grouping is a planner decision and `Feature.parameters()` stays flat.

**Gate — the whole safety property of this PR:** `tests/test_refactor_golden.py` output must be byte-identical. If golden output moves during a pure refactor, the refactor is wrong. **Never re-bless.**

PR #880, opened 2026-07-27
feat(#869): derived ParameterId + the grid-pitch discriminator

### LINKED ISSUE #869: ADR 0016 PR1: derived ParameterId + the grid-pitch axis discriminator
Child of #867.

Land the semantic key as a pure derivation — `role` + `kind` + an optional discriminator — with no production behaviour change:

- the discriminator on the two `DimParameter("length", "grid_pitch", …)` grid pitches (`axis="x"`/`"y"`), today's sole tier-2 case;
- a **uniqueness audit** over every feature's parameters, phrased so it will later hold over addressable units;
- a **stability test** that re-detecting a part yields the same `ParameterId`s.

Derived, not hand-authored: an explicit `id=` on each of the ~40 `DimParameter(...)` sites would restate `(kind, role)` where it adds nothing and create a new silent-bug class when the two disagree (ADR 0016, 'Dimension identity').

**Gate:** pure addition. `tests/test_refactor_golden.py` must be untouched.

pzfreo:
The ADR 0016 identity key, as a pure addition — no production behaviour
changes, and the golden corpus is untouched.

`DimParameter.parameter_id` derives `role.kind`, plus a discriminator where
one is needed. Derived rather than hand-authored so the ~40 construction
sites cannot drift a literal away from the `role=` beside them.

Two design calls the ADR's illustrative strings did not settle, amended
there in this commit:

`kind` is included ALWAYS, not only where it disambiguates. The prettier
rule (drop it when a role happens to be unique, giving `step_height`) makes
an id depend on its SIBLING parameters, so adding a field to a feature would
silently repoint every intent aimed at an existing one. That destroys the
stability the id exists for, and stability outranks prettiness. Ids are
uniform: bore.diameter, bore.depth, step_height.length, grid_pitch.length.row.

The discriminator is `row`/`col`, not `x`/`y`. PatternFeature.angle may
rotate the lattice, so a row pitch is not an X pitch in general and the IR
must key on what it actually knows. Mapping a user-facing `axis=` onto
row/col belongs at the facade, where the angle can be consulted; how `axis=`
should read on a 30-degree grid is now an open question against #872.

The uniqueness audit is fail-closed: the feature universe is enumerated
mechanically from model.ir, so a new Feature type fails the suite until it
is sampled. Verified minimal — with every exemption removed, exactly the two
known tier-3 correlated sets collide (RotationalFeature's concentric bores
and StepLevelFeature's step_height ladder) and nothing else. Those
exemptions are scaffolding that #870 deletes when AddressableDimension makes
a correlated set one unit holding N members.

Closes #869.

---

**Epic:** #867 (ADR 0016 implementation), PR 1 of 8.

**Verification**
- Full fast tier: 1701 passed, 1 skipped (the documented rotational exemption).
- `tests/test_refactor_golden.py`: 14 passed, **unchanged** — this PR adds a field and a property, it does not alter output.
- All four lint checks clean.


pzfreo:
## Adversarial review (Codex, 4 rounds) — clean

All findings were on the **tests**, none on the production change. Each round tightened one theme — whether the guards actually guard — and each fix was real, so the rounds converged rather than thrashed.

| Round | Finding | Fix |
|---|---|---|
| 1 | Exemptions broad enough to hide collisions: whole-feature skip for `RotationalFeature`, bare-role filter for `step_height`/`step_position` | Keyed on exact `(feature kind, role)` pairs; two tests pin that the narrowing is real |
| 1 | Stability test compared only `(kind, parameter_id)`, so an id could repoint silently | Carries value + feature origin; row/col→grid-tuple mapping pinned explicitly |
| 2 | Two live detections still cannot catch a *code change* that repoints | Committed `_EXPECTED` map; verified non-vacuous against a simulated width/depth swap |
| 3 | That map covered only 5 ids — most of the identity surface was still determinism-only | `_SAMPLE_BINDINGS` pins all 22 feature types + a completeness check |

**Round 4: "No substantive findings … the code is ready."**

Two things worth recording that came out of doing the work rather than the review:

- Generating the full binding surfaced a genuine weakness in my own fixtures — three pattern samples had `pitch == row pitch == 30.0`, so a pitch/row-pitch swap would have hidden behind equal values. Sample pitches are now 25.0.
- Blessing the detected snapshot surfaced a wrong comment of mine: the fixture's four corner holes group into one count-group feature, **not** a grid pattern, so that part carries no grid pitches at all. Row/col is pinned at unit level, where it always was.

**Note on review scope:** Codex cannot execute here (the sandbox cannot create a namespace under Ubuntu's `apparmor_restrict_unprivileged_userns`), so it reviewed the pushed SHAs via the GitHub API. Execution is covered locally: full fast tier 1730 passed, golden corpus unchanged, all four lint checks clean.

PR #882, opened 2026-07-27
feat(#870): AddressableDimension — the identity layer, at the planner boundary

### LINKED ISSUE #870: ADR 0016 PR2: AddressableDimension as a first-class planner type; retype DimensionGroup.dims

pzfreo:
`AddressableDimension(id, members)` is the unit a `dimension(...)` line names and
the thing suppression, provenance and dedup key on. Members are
`PlannedDimension`s, not raw `DimParameter`s, so the planner's decisions —
convention, model-level suppression and its reason, the datum, provenance —
survive the grouping instead of needing a parallel structure beside it.

Grouping is DECLARED, never inferred from an id collision. `CORRELATED_SETS`
names the exact (feature kind, role) pairs that form one unit with N members:
the step_height/step_position ladders and a rotational body's concentric bores.
Without that table there is no way to tell an intentional set from an accidental
duplicate, and a grid pitch that lost its discriminator would silently become a
"correlated set" instead of failing the audit.

The names came out inverted from the ADR's sketch, and the migration went to
zero. The plan was to retype `dims` to hold units and add a flattening accessor,
at ~25 mechanical read sites. The same semantics fall out of adding `units` as
the field and making `dims` the flattening property — so all 22 source readers
and ~30 test readers are untouched, and only the construction sites changed (the
planner, plus five tests that build ad-hoc plans). On the riskiest child of this
epic a zero-churn diff is worth more than matching the sketch. ADR amended.

This also retires the exemption table PR 1 needed. Phrased over addressable
units rather than raw parameters, the tier-3 correlated sets stop being a
special case: a ladder is one unit holding three members so it cannot collide
with itself, while two ungrouped parameters sharing an id are two units and
still fail. Structure replaced the table — an exemption can hide a real
collision, arity cannot. The test now asserts no exemption list exists.

The #868 canaries improved in passing: they now suppress by identity
(`u.id != "bore.depth"`), which is the actual ADR 0016 operation rather than a
hand-rolled parameter filter.

Closes #870.

---

**Epic:** #867 (ADR 0016 implementation), PR 2 of 8 — the risky one.

**Verification**
- `tests/test_refactor_golden.py`: 14 passed, **unchanged**. This is the safety property for this PR: if golden output moves during a pure refactor, the refactor is wrong. Not re-blessed.
- Full fast tier: 1732 passed.
- All four lint checks clean.

### DIFF
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -565,8 +565,8 @@ constraining planner internals that are still moving. The gaps are known and fin
 | Low-level furniture | Centre marks, section arrows, hatching, the NTS caption carry no editable intent | Engine-automatic, by decision |
 | Anything the emitter cannot re-solve | The fidelity floor `emit_sheet_script` already holds for features | Self-describing comment |
 
-Two of those five are the identity model's boundary and shrink as it grows (relations,
-future correlated-set splits); the other three are decisions, and stay. **The property the
+Three of those six are the identity model's boundary and shrink as it grows (relations,
+locations, future correlated-set splits); the other three are decisions, and stay. **The property the
 mirror actually promises is that within the identified set, a line's presence and its
 absence both mean something exact** — which is all suppression-by-omission needs.
 
@@ -788,7 +788,8 @@ second with the single-source-of-truth of the first — over the identified set,
   suppression lands with the set boundary, while **post-build** suppression of an automatic
   dimension waits on the reconstruction question (#867) — not on the closed #426/#707.
 - `sheet_emit` gains a dimension-mirroring pass: after the feature basis, one referential
-  `dimension(...)` line per **planned** dimension, led by the explicit dimension-source call
+  `dimension(...)` line per **addressable dimension** — the unit, not the member, so a
+  correlated set gets one line — led by the explicit dimension-source call
   — each commentable and editable, none restating a number, with low-level furniture still
   produced automatically on re-run. The pass reads the planner's `DimensionGroup`s, **not**
   the drawing's placed annotations, so a solver-dropped dimension keeps its line. What it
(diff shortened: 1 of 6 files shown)


pzfreo:
## Adversarial review (Codex, 5 rounds) — **Merge**

First PR reviewed with execution enabled, and it paid for itself immediately: round 1 verified the golden file's **git blob hash** was identical to base rather than taking the claim on trust, ran both suites itself, and *reproduced* the constructor `TypeError` instead of theorising about it.

| Round | Finding | Resolution |
|---|---|---|
| 1 | Location dims bypass the identity layer while the docstring claimed `AddressableDimension` is *the* addressable unit | Scoped the claim; filed **#883** — the location path is structurally different, not merely unwired |
| 1 | "Migration cost went to zero" too strong — `DimensionGroup`'s constructor changed on a publicly exported type | Recorded as an intentional API break, with rationale |
| 3 | The location correction landed in one paragraph while four other passages still asserted locations are addressable | Propagated everywhere, plus two knock-ons the review had not flagged |
| 4 | Table grew to six rows but the text still said "two of those five" | "Three of those six", locations named as a boundary gap |
| 4 | Two passages promised one emitted line per `PlannedDimension` — contradicting the rule that a correlated set is N members under **one** unit | Both now say one line per `AddressableDimension` |

**Round 5: "Merge. No blocking issue found."**

**No code defect was found in any round.** Round 4 explicitly confirmed `_addressable()` groups only explicitly declared correlated sets, preserves ordering and planner metadata, and leaves accidental id collisions visible.

Two things worth recording beyond the findings:

- **The golden corpus cannot catch incorrect addressable grouping**, because flattening `units` yields the same `dims`. Expected — grouping is new semantic state with no rendering effect yet — but it means correctness here rests on `test_parameter_id.py`, not the golden. Better to know that than assume otherwise.
- Rounds 3–4 were consistency debris created by my *own* corrections to a 900-line ADR, not new discoveries. The lesson is to sweep every affected claim in one pass rather than fixing the passage in front of me.

**Verification:** golden corpus unchanged (blob-hash identical), full fast tier 1732 passed, four lint checks clean.

Issue #893, opened 2026-07-27
Balloon ring collapsed onto the plan view (CTC-02, regression in #888)


pzfreo:
Slow-tier failure on `main`: `test_ctc02_top_balloon_ring_hugs_dimensions`.

The plan-view balloon ring no longer stands off from the view — it sits **6.5 mm** above the plan top where the test requires **>20 mm**.

## Measured, pre vs post

Built CTC-02 AP203 at `3f92fd6` (pre-#888) and at `4165484` (#888):

| | pre-#888 | post-#888 |
|---|---|---|
| balloons | 22 | 22 |
| plan-view top | 472.93 | 481.60 |
| **highest balloon** | **552.83** (+79.9) | **488.10** (+6.5) |
| `hole_table_plan` | present | present |

The top ring **collapsed by ~65 mm onto the view**. Post-#888 all 22 balloons sit within Y 431–488, which is *inside* the plan view's own span (309–482) — they are on the holes rather than ringed above them.

Two things this rules out:

- **Not the hole table.** It is present in both; the escalation is not new.
- **Not balloon count.** 22 in both.

## Why #888 caused it indirectly

#888 touched neither `annotations/balloons.py` nor `layout.py`. The band assignment (`_assign_balloon_bands`, the min-cost max-flow solve, #516) reads what is already placed in the view, so #888 changing the plan view's occupancy changed the band solution.

## Not caused by, and not fixed by, #894

Verified independently: with the #894 collect-then-solve fix applied the standoff is **identical at 6.5 mm** (the absolute figures shift — 488.10 vs 481.60 — because the view moves, but the relationship does not). Different subsystem: #894 is the strip corridor, this is the balloon band solve. Deliberately kept out of that PR rather than putting two unrelated solver fixes in one diff.

## Repro

```
uv run pytest tests/test_e2e_standards.py::test_ctc02_top_balloon_ring_hugs_dimensions -q -m slow
```

Release blocker for 0.3.10 alongside #894.

pzfreo:
## Root cause found — it is not in the balloon code

Measured on CTC-02 AP203. The balloon ring is a **symptom**; the defect is upstream.

### The chain

```
hole_table_plan sits above the plan   (depth ~188)
  -> locx location ladder stacks BEYOND it   (rungs at 236.3, 264.8, 274.3, 302.8)
    -> balloon top band must clear that       (top_dim = 305.1)
      -> band line lands at y=801, ~400 mm from every hole
[...]

Issue #125, opened 2026-06-20
Plan-view top-side balloon leaders are far too long on dense patterns (CTC-02)


pzfreo:
## Observation

On the NIST **CTC-02** drawing, the **top-side** balloon ring in the **plan view** has leader lines that are roughly **3–4× longer** than the leaders on the other three sides (left / right / bottom). The top balloons are pushed into a row well above the view, with a tall fan of near-vertical leaders climbing the full height of the view again, while the left/right/bottom balloons hug the view edge with short, tidy leaders.

This is the four-side balloon ring introduced in #114 / refined under the compose-then-pack work (#112/#121). The packing now correctly keeps the top band clear of the front-view dimensions, but the top balloons themselves are being parked much farther out than necessary — the stand-off looks fixed/over-reserved rather than sized to the ring.

## Why it matters

- Legibility: the long near-parallel leaders are visually noisy and make it hard to trace a balloon back to its feature.
- Asymmetry: the drawing looks unbalanced — three tight sides, one sprawling side.
- Likely wasted sheet space (it contributes to the A1→A0 escalation pressure on CTC-02).

## Suspected cause

The top-side ring's leader length / stand-off appears to be derived from a band reservation rather than from the actual ring geometry, so after layout/escalation the top balloons land far above the view. Worth checking:
- whether the top band's balloon row uses the same per-side offset logic as left/right/bottom, and
- whether the reserved top band height is being consumed as literal leader length instead of a tighter ring radius.

## Acceptance

Top-side balloon leaders on CTC-02 should be comparable in length to the other three sides (within a small factor), with the balloons sitting just outside the view footprint.

## Not urgent

Cosmetic/legibility polish — file-and-defer. Separate from the #121 overlap correctness fix (which is about views not colliding, and is what surfaced this).

PR #903, opened 2026-07-28
Fix balloon perimeter assignment after hole-table escalation

### LINKED ISSUE #901: Balloon ring standoff is not guaranteed after hole-table escalation
Split out of #893 as the general defect; #893 is the CTC-02 instance.

## The defect

When `_maybe_tabulate_holes` escalates a dense plan view to a hole table + balloons, the balloon ring's standoff from the plan view is whatever falls out of residual strip occupancy. It is not bounded below, so the ring can land essentially on the view.

Measured standoffs above the plan-view top edge:

| part | standoff |
|---|---|
| synthetic 18-hole plate (below) | **19.0 mm** |
| NIST CTC-02 | **6.5 mm** |
| NIST CTC-04 | **14 mm** |

A ring at 6.5 mm is visually on top of the view it annotates.

## Why it happens

Escalation is by design (#93): when callout or location dims drop, the individual dims are removed and replaced by the table. What is *not* designed is where the ring then sits. The top band's line is placed beyond the deepest measured occupant, and the min-cost band assignment (#516) chooses by leader distance — so once the top band is pushed far out, every balloon prefers the near left/right bands and no ring forms above the view at all.

On CTC-02 the top band ends up ~400 mm from the holes against ~85 mm to the sides, with capacity 17 going unused. Capacity-vs-depth data is on #893.

## Fast repro — 14.5 s, no STEP import

```python

def _dense_plate(n=18):
    p = Box(200, 140, 14)
    for i in range(n):
        x = -90 + i * (180 / (n - 1))
        y = -60 + (i * 37) % 115
        p -= Pos(x, y, 0) * Cylinder(2.0 + 0.35 * i, 40)   # every ⌀ distinct
    return p

d = build_drawing(_dense_plate())
pl, pb, pr, pt = d.view_bounds("plan")
tops = [o.bounding_box().max.Y for _n, o in d.iter_annotations() if _n.startswith("balloon_plan")]
assert max(tops) > pt + 20      # FAILS: 19.0
```

Committed as an **xfail** in `tests/test_balloon_ring_standoff.py`, so the repro stays runnable and flips to a pass when this is fixed.

**Distinct diameters are load-bearing in the repro:** identical holes collapse into one grouped `n×` pattern callout, which never congests the strip — a plate with 20 *identical* holes shows nothing. Per-hole callouts and per-hole location dims are what fill the plan-above strip.

## Blast radius

Any part dense enough to escalate. Not exotic — a fixture plate, manifold or die shoe with many distinct hole specs. Parts with a simple bolt pattern are unaffected.

## Not the cause (checked)

- **The hole table.** `tabulate` is the last pass in `_PASS_SEQUENCE`, placed globally by `fit_box` over the whole page. On CTC-02 it sits at depth 349 while the ladder is at 236–303 — outside it, arriving after. It is a consequence of escalation, not a cause.
- **Band carving / band depth search.** Both implemented and measured against #893; neither moved the standoff, because the problem is in the *assignment*, upstream of band placement.

pzfreo:
Closes #901.

Stacked on #902, which contributes the synthetic dense-plate reproduction.

## Design
- expresses the scattered-hole-table remedy as a `perimeter` render intent
- extends the geometry-only band solver with a lexicographic activation objective: maximum cardinality, then maximum requested-band coverage, then minimum leader length
- gives perimeter rings a measured 20 mm minimum view-edge-to-outer-glyph extent
- preserves nearest-band placement and existing spacing for manual and pattern-only balloons
- records the updated solve contract in ADR 0014

## Verification
- `uv run ruff check ...`
- `uv run pytest -m smoke -q` (45 passed)
- focused balloon/solver regression selection (9 passed)

After #902 merges, this PR can be retargeted to `main` without changing its effective diff.

### DIFF
--- a/docs/adr/0014-collect-then-solve-annotation-placement.md
+++ b/docs/adr/0014-collect-then-solve-annotation-placement.md
@@ -48,11 +48,13 @@ Per view, per strip: **collect → solve → emit.**
     band assignment — a deterministic max-cardinality flow solve
     (`_assign_balloon_bands`, `layout.py`, #516).  A dense scattered-hole
     table escalation requests perimeter coverage (#901): within the maximum
-    flow, the solver lexicographically maximises the number of requested usable
-    bands activated before minimising leader length.  The render pass supplies
-    only band names, capacities, and numeric costs, keeping the solver
-    geometry-only; ordinary and manually requested balloons retain pure
-    minimum-cost assignment.  The perimeter render pass measures candidate
+    flow, the first use of each preferred usable band receives a bounded
+    distance credit before leader length is minimised. This is deliberately a
+    preference rather than a lexicographic override—a remote band stays empty
+    instead of creating a cross-part leader. The render pass supplies only band
+    names, capacities, and numeric costs, keeping the solver geometry-only;
+    ordinary and manually requested balloons retain pure minimum-cost
+    assignment. The perimeter render pass measures candidate
     lanes at obstacle boundaries and carves each into free horizontal segments;
     a geometry-only dynamic-programming solve assigns ordered members across
     those segments at minimum L1 leader cost.  This keeps a local remote
@@ -60,7 +62,9 @@ Per view, per strip: **collect → solve → emit.**
     preserving crossing-free member order.  The selected lane must hold both
     its balanced share and any member-count deficit left by the other bands, so
     lane selection cannot weaken the downstream solver's maximum-cardinality
-    guarantee.
+    guarantee whenever such a lane exists. If no lane meets the target, the
+    nearest lane wins rather than recreating the remote beyond-all-obstacles
+    geometry; any resulting capacity loss is surfaced as `balloon_dropped`.
   - **Order** — label order along the strip = site/feature order (candidates
     sort by anchor coordinate), so leaders between **distinct** strip-axis
     coordinates are **crossing-free by construction**; coincident sites
--- a/docs/adr/0014-collect-then-solve-annotation-placement.md
+++ b/docs/adr/0014-collect-then-solve-annotation-placement.md
@@ -57,7 +57,10 @@ Per view, per strip: **collect → solve → emit.**
     a geometry-only dynamic-programming solve assigns ordered members across
     those segments at minimum L1 leader cost.  This keeps a local remote
     obstacle from pushing the entire ring beyond its outer edge (#125), while
-    preserving crossing-free member order.
+    preserving crossing-free member order.  The selected lane must hold both
+    its balanced share and any member-count deficit left by the other bands, so
+    lane selection cannot weaken the downstream solver's maximum-cardinality
+    guarantee.
   - **Order** — label order along the strip = site/feature order (candidates
     sort by anchor coordinate), so leaders between **distinct** strip-axis
     coordinates are **crossing-free by construction**; coincident sites
--- a/docs/adr/0014-collect-then-solve-annotation-placement.md
+++ b/docs/adr/0014-collect-then-solve-annotation-placement.md
@@ -45,8 +45,14 @@ Per view, per strip: **collect → solve → emit.**
     assignment is (a) segment assignment within a carved strip
     (`carve_free_segments` + innermost-first fill in
     `place_strip_candidates`), and (b) the balloon pass's genuinely global
-    band assignment — a deterministic min-cost max-flow solve
-    (`_assign_balloon_bands`, `layout.py`, #516).
+    band assignment — a deterministic max-cardinality flow solve
+    (`_assign_balloon_bands`, `layout.py`, #516).  A dense scattered-hole
+    table escalation requests perimeter coverage (#901): within the maximum
+    flow, the solver lexicographically maximises the number of requested usable
+    bands activated before minimising leader length.  The render pass supplies
+    only band names, capacities, and numeric costs, keeping the solver
+    geometry-only; ordinary and manually requested balloons retain pure
+    minimum-cost assignment.
   - **Order** — label order along the strip = site/feature order (candidates
     sort by anchor coordinate), so leaders between **distinct** strip-axis
     coordinates are **crossing-free by construction**; coincident sites
--- a/docs/adr/0014-collect-then-solve-annotation-placement.md
+++ b/docs/adr/0014-collect-then-solve-annotation-placement.md
@@ -52,7 +52,12 @@ Per view, per strip: **collect → solve → emit.**
     bands activated before minimising leader length.  The render pass supplies
     only band names, capacities, and numeric costs, keeping the solver
     geometry-only; ordinary and manually requested balloons retain pure
-    minimum-cost assignment.
+    minimum-cost assignment.  The perimeter render pass measures candidate
+    lanes at obstacle boundaries and carves each into free horizontal segments;
+    a geometry-only dynamic-programming solve assigns ordered members across
+    those segments at minimum L1 leader cost.  This keeps a local remote
+    obstacle from pushing the entire ring beyond its outer edge (#125), while
+    preserving crossing-free member order.
   - **Order** — label order along the strip = site/feature order (candidates
     sort by anchor coordinate), so leaders between **distinct** strip-axis
     coordinates are **crossing-free by construction**; coincident sites
--- a/src/draftwright/_core.py
+++ b/src/draftwright/_core.py
@@ -374,12 +374,19 @@ def _dim(p1, p2, side, distance, draft, **kwargs):
 # extension-line origins) and subsequent parallel lines stack tighter and uniform (#347).
 _STRIP_GAP = 10.0  # clearance between the view outline and the first dimension line
 _STRIP_SPACING = 2.5  # clear gap between successive parallel dimension lines (beyond the label)
+_BALLOON_RADIUS_FONT_FACTOR = 1.5
+
+
+def _balloon_radius(font_size: float = _FONT_SIZE) -> float:
+    """Balloon glyph radius in page-mm for the active annotation font."""
+
+    return _BALLOON_RADIUS_FONT_FACTOR * font_size
 
 
 def _balloon_halo(font_size: float = _FONT_SIZE) -> float:
     """Reserved/rendered plan-balloon extent from the view edge, in page-mm."""
 
-    return _STRIP_GAP + 3 * font_size + _STRIP_SPACING
+    return _STRIP_GAP + 2 * _balloon_radius(font_size) + _STRIP_SPACING
 
 
 # Small lift (page-mm) of an overall/envelope witness line off the projected silhouette edge,
--- a/src/draftwright/_core.py
+++ b/src/draftwright/_core.py
@@ -374,6 +374,14 @@ def _dim(p1, p2, side, distance, draft, **kwargs):
 # extension-line origins) and subsequent parallel lines stack tighter and uniform (#347).
 _STRIP_GAP = 10.0  # clearance between the view outline and the first dimension line
 _STRIP_SPACING = 2.5  # clear gap between successive parallel dimension lines (beyond the label)
+
+
+def _balloon_halo(font_size: float = _FONT_SIZE) -> float:
+    """Reserved/rendered plan-balloon extent from the view edge, in page-mm."""
+
+    return _STRIP_GAP + 3 * font_size + _STRIP_SPACING
+
+
 # Small lift (page-mm) of an overall/envelope witness line off the projected silhouette edge,
 # so its extension line reads as distinct from the outline rather than sitting on top of it.
 _WITNESS_LIFT_MM = 2.0
--- a/src/draftwright/annotations/balloons.py
+++ b/src/draftwright/annotations/balloons.py
@@ -33,6 +33,25 @@ from draftwright.layout import (
 _MIN_PERIMETER_EXTENT = 20.0 + _STRIP_SPACING
 
 
+def _top_lane_target(member_count, other_capacities, usable_band_count):
+    """Capacity the top lane needs for both balance and maximum cardinality."""
+
+    balanced_share = math.ceil(member_count / usable_band_count)
+    capacity_deficit = max(0, member_count - sum(other_capacities))
+    return max(balanced_share, capacity_deficit)
+
+
+def _select_top_lane(lane_options, target, fallback_line):
+    """Nearest sufficient lane, else nearest maximum-capacity best effort."""
+
+    fitting = [option for option in lane_options if option[2] >= target]
+    if fitting:
+        return fitting[0]
+    if lane_options:
+        return max(lane_options, key=lambda option: (option[2], -option[0]))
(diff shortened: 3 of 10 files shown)


pzfreo:
## Review — measured on `dbafc99`

**The mechanism works, but it does not unblock the release: CTC-02 is still red, and now fails the #125 guard instead.**

### Blocker — trades one assertion for the other

```
tests/test_e2e_standards.py::test_ctc02_top_balloon_ring_hugs_dimensions
E   assert gap < 60
E   assert 265.55000000000007 < 60
```

The first assertion now passes (standoff 6.5 → 325.10). The second fails — and it is the one #125 exists for: *"Pre-#125 the top ring sat ~150 mm above the highest dim; the fix keeps the whole ring within a small standoff of it."* This puts the ring 265 mm above it, worse than the condition #125 was written to stop.

**CI would not have caught this.** `test_ctc02...` is slow-tier, so it runs post-merge on `main`, not on the PR. This would have landed and left the release blocker red with a new message.

### What does work (measured)

| | before | after |
|---|---|---|
| synthetic 18-hole plate | 19.0 | **20.0** |
| CTC-02 | 6.5 | **325.10** |
| CTC-04 | 14 | **249.05** |

Band spread on the synthetic plate goes from all-sides to left 5 / right 6 / **top 7** — the ring genuinely reforms. The `activate_bands` lexicographic-activation approach is sound, keeps the solver geometry-free, and the `activation_bonus` bound (`1 + Σ max per-member cost`) correctly dominates any real change in assignment cost.

### Why it overshoots

Activation makes the top band *chosen*, but does nothing about *where that band is*. On CTC-02 the top band line still sits ~305 mm out — beyond the deepest occupant, as before. Previously the min-cost solve quietly declined to use it; now it is forced to, so the balloons travel 300 mm to get there.

Both halves of #901 are needed: the band must be **reachable** (this PR) *and* **near** (not yet addressed).

For whoever takes the depth half — I tried two approaches on #893 and both failed, so they can be skipped: band carving (inert, because the problem is upstream in the assignment) and a shallowest-depth search (wrong, because `m_locx12`'s **label** is genuinely at depth 302.8 — there is no shallow line that clears the dims). Capacity-vs-depth data is on #893: the band holds 12 balloons at depth 0 and 17 at depth 305, so shallow placement is viable on capacity; it is the obstacle-clearance rule that forces it out.

### Smaller points

- **Zero margin.** The synthetic standoff is exactly 20.0000 against a threshold of 20 (`MARGIN = 0.000000`), which is why the `>` had to become `>= … - 1e-6`. Defensible, since `_MIN_PERIMETER_EXTENT` enforces a floor — but the test now only verifies that the floor is applied.
- **Bottom band unused** on the synthetic case (left 5 / right 6 / top 7 / **bottom 0**) despite `activate_bands` including every band with capacity > 0. Worth checking whether `has_bottom` excluded it or activation silently failed there.
- **Perimeter is forced for every escalation.** With few balloons and four usable bands the solver spreads one per side; `test_activation_is_best_effort_when_members_are_fewer_than_bands` documents the behaviour, but it is worth eyeballing on a part with ~5 holes.

### Suggestion

Either hold until the depth half lands too, or land this as the reachability half *after* resolving the CTC-02 regression — on its own it converts one red assertion into another, and the new one is the #125 condition.

pzfreo:
## Re-review — measured on `0c7e517`

**The blocker from my last review is fixed.** `test_ctc02_top_balloon_ring_hugs_dimensions` now passes **both** assertions (it was failing `gap < 60` at 265.55 after commit 1 alone), along with all three tests in `test_balloon_ring_standoff.py`, and the xfail is correctly unmarked.

The second commit is the depth half I said was missing — and it is a better design than what I attempted on #893. My shallowest-depth search failed because I required the near lane to match the *deepest* lane's capacity; `ceil(len(specs) / nbands)` is the right target and I should have got there.

### Verified by measurement

| check | result |
|---|---|
| CTC-02 e2e (both assertions) | **passes** |
| synthetic standoff | 20.0 → now with real clearance (`20.0 + _STRIP_SPACING`) |
| CTC-04 standoff | 14 → 249.05 |
| full-build determinism, same part ×2 | identical placements |
| solver determinism on exact ties | identical |
| activation vs 0-vs-1e6 cost spread | activation still wins (left 3 / top 1) |
| 1 member, 4 bands requested | goes to cheapest only — no spurious spreading |
| over-capacity | drops correctly |
| zero-capacity requested band | ignored cleanly |
| non-escalating part (identical holes) | 0 balloons — unchanged |
| lane-probe cost | **3 candidate lanes from 99 obstacles** — the X-window filter prunes hard |

All three concerns I raised last round are closed: zero margin (now has a spacing unit of real clearance), performance (non-issue, measured), determinism (verified).

An independent adversarial pass built a **3,000-case randomised exhaustive oracle** over the flow solver, comparing it against the exact lexicographic objective (cardinality → activated bands → cost). It matched exactly. It also confirmed the `activation_bonus` bound is sufficient for *multiple* activated bands, that the negative residual edges introduce no negative-cycle or nondeterminism risk under successive shortest augmenting paths, and that `perimeter=True` is confined to successfully placed scattered-hole tables.

### One finding — Medium, and I believe latent rather than live

`top_target = ceil(len(specs) / (3 + has_bottom))` ([balloons.py:128](../blob/0c7e517/src/draftwright/annotations/balloons.py#L128)) assumes the other bands can absorb equal shares. Lane selection then takes the nearest lane meeting only that target ([balloons.py:147](../blob/0c7e517/src/draftwright/annotations/balloons.py#L147)), even when side/bottom capacity is much smaller — so a near, low-capacity lane can be chosen over a farther one that would have placed everything, dropping balloons and contradicting the solver's maximum-cardinality objective.

Suggested bound: require top capacity ≥ `len(specs) - sum(other capacities)`, keeping the balanced-share figure as an aesthetic floor.

**Reachability:** I could not construct it on a real part. On the dense plate the side bands hold **41 each (82 combined) against 18 specs**, so the near lane can never be the binding constraint — the scenario needs side+bottom capacity below the spec count, which implies a page far smaller relative to the view than anything in the fixtures. I would treat it as a latent correctness gap worth closing on its own terms, not a merge blocker.

### Not verified

I did not run CTC-04's own e2e test or the full slow tier — only its standoff number (249.05). Given CTC-04 is the other escalating fixture and the slow tier is post-merge-only, that is the gap I would close before merging, since exactly this class of thing is what my last review caught.

### Verdict

The mechanism is sound, the regression I found is fixed, and the design is right. **Approve once CTC-04's e2e is green**, with the `top_target` bound as a follow-up.

pzfreo:
## Third review — deeper pass, and it reverses my verdict

A second adversarial reviewer went after the angles the first two passes did not reach. **Revising my earlier "approve once CTC-04 is green" to changes-needed.** Everything below is measured in a worktree at `0c7e517`.

### The headline: mutation testing says both engineered halves are unexercised

| mutation | result |
|---|---|
| `_MIN_PERIMETER_EXTENT = 0.0` (keep all machinery) | ring extent **19.00 mm** → the #901 test **fails** |
| `perimeter=False` entirely | also 19.00 mm |
| `top_target = 0` (disables lane escalation) | **all 26 tests pass** |
| `activate_bands=()` at the call site (disables activation) | **all tests pass** |
| `activation_bonus = 1` | correctly fails 2 unit tests |

So the entire pass on the fast repro comes from the `20.0 + _STRIP_SPACING` constant at `balloons.py:33`. The solver-level *magnitude* is pinned, but the **wiring and lane policy are not** — on six synthetic escalating plates the top band always carved to a single full-span segment (escalation removes the location dims first, so nothing is left in the top x-window) and the assignment was identical with and without activation. The only fixture exercising either mechanism is CTC-02, which is `@pytest.mark.slow` — post-merge, not a PR gate.

### A — `activate_bands` has no distance ceiling

`balloons.py:189` requests **every** usable band, and the bonus (`1 + Σ max member cost`) is lexicographically dominant over any cost difference. So a band activates regardless of distance, and the objective maximises *band count*, not ring balance.

Measured on `Box(300,100,14)` with 17 distinct-⌀ holes clustered on the right third:

| | longest leader | total leader |
|---|---|---|
| PR as-is | **229.3 mm** | 1242.9 mm |
| `activate_bands=()` | 99.3 mm | 1125.1 mm |

That leader crosses the full width of the plan view, over ~16 holes and every dimension between. Balloons set `is_centerline = True` (`balloons.py:312`), so `annotation_overlap` will not flag it — **invisible to lint**.

### B — the lane fallback re-creates the geometry #901 is about

`top_target = ceil(len(specs) / (3 + has_bottom))` is an even-share estimate. When no lane meets it, `balloons.py:150-155` falls back to `max(capacity, -line)`. Capacity is monotone in height, so maximum capacity is first reached **just above the tallest obstacle** — literally "pushed beyond the deepest occupant".

Measured on the PR's own stub harness (18 specs): three tall obstacles → lane at deepest+7, fully clear segments, **1 balloon assigned**. Band-spanning obstacle → lane at `pt+107`, **1 balloon**.

Compounding cause: the top window (plan width) is structurally shorter than left/right (page height), so an even share systematically over-asks of the shortest band. On the stub the *unobstructed* window holds 5 while `top_target` is 6 — **every lane fails the gate with no obstacles at all**.

### D — the 22.5 mm ring overruns the 21.5 mm reserved halo

`compose.py:119` reserves `_STRIP_GAP + 3·font_size + _STRIP_SPACING` = **21.5**; `balloons.py:33` now draws to **22.5**. The docstring at `balloons.py:45-47` still says the balloons "sit in clear space" because layout reserved the band — false by 1.0 mm, and ADR 0004 block-disjointness rests on that reservation. The old design had 2.5 mm of deliberate slack; it now has −1.0.

Not reachable on the six fixtures measured (min slack 26.2 mm), but there is a reachable path: when `_will_balloon` is false yet the runtime trigger fires, no halo is reserved, `first_line` can exceed `last_line`, and the top band zeroes. On a tight stub sheet `perimeter=False` places 3 and drops 7; **`perimeter=True` places 0 and drops 10.**

Cheapest fix: derive the floor from `_est_plan_halo()` so one number governs both reservation and render.

*Also resolves my earlier bottom-band question:* 0 bottom balloons on the dense plate is **correct** — `bottom` capacity is 0 there so it is never in `activate_bands`. But raising the extent makes `has_bottom` harder (`balloons.py:108`), so perimeter mode can remove a band from the ring it is trying to spread.

### E — minor: segmented placement is all-or-nothing

`balloons.py:270`: `_solve_segmented_strip_1d(...) or []` renders **zero** balloons for the band when the DP returns `None`, where the non-segmented path degrades via `prefix=True`. Reachable only within `1e-9·gap` of an exact multiple, so vanishingly rare — but the asymmetry is worth a prefix retry.

### Verified clean

- **`_solve_segmented_strip_1d` is sound** — 4,000-case fuzz: zero ordering violations, zero gap violations, zero out-of-segment positions, and `None` iff count > Σ capacity. **The ADR 0014 crossing-free guarantee holds by construction.**
- The `carve_free_segments` pad is exactly half the balloon gap, so the `ValueError` guard cannot fire from this caller.
- The negative-cost flow is valid — DAG, no negative cycles, SSP+SPFA correct, cardinality still dominates.
- Degenerate geometry is safe and in one case *safer* than before: with the plan flush to the page top the band zeroes cleanly, where the old code rendered at y=184.5 on a 180 mm page.
- Guards pass: import boundaries + carve callers (15), layout cleanliness (20, no new overlaps), the PR's unit tests (23).

### Verdict

The design is right and the solver work is solid. But the reported symptom is being fixed by the **constant**, while the two engineered halves are unexercised by any PR-gated test and actively harmful when forced to run. Before merge I would want:

1. **A** gated on distance — a preference, not a lexicographic override
2. **B**'s `top_target` reconciled with the assignment, or the fallback preferring the *nearest* lane rather than max-capacity
3. **D**'s constant derived from `_est_plan_halo`
4. At least one **PR-gated** test that fails when `activate_bands` or the lane probe is disabled

pzfreo:
## Fourth review — `24f30df`. All four findings addressed and verified; **approve**.

Re-measured everything I flagged. Each fix is real, and — the part that was missing last round — each is now **pinned by a PR-gated test**.

### A — unbounded activation → bounded preference

`activate_bands` became `prefer_bands` + `preference_limit`: a bounded preference rather than a lexicographic override.

Re-ran the case that produced the 229.3 mm leader (300×100 plate, 17 distinct-⌀ holes clustered on the right third):

| | longest balloon extent |
|---|---|
| before | **229.3 mm** |
| now | **100.0 mm** |

The leader no longer drags across the full plan view.

### B — the fallback no longer re-creates the #901 geometry

`_top_lane_target` is now `max(balanced_share, capacity_deficit)` with `capacity_deficit = member_count - Σ(other capacities)` — the exact bound suggested — and the fallback is the **nearest** lane (`lane_options[0]`) rather than max-capacity. That was the mechanism that reinstated "beyond the deepest occupant", and it is gone.

The even-share over-ask is fixed too: the target now accounts for what the sides can actually absorb, so a structurally short top window no longer fails every lane.

### D — one number governs reservation and render



The 22.5-vs-21.5 overrun is gone; ADR 0004's block-disjointness reservation is honoured again, and the `balloons.py` docstring's "sits in clear space" claim is true once more. Deriving it from `_est_plan_halo` was the suggested fix and it is exactly what landed.

### E — segmented placement degrades gracefully

`_solve_segmented_strip_1d` gains `prefix=True`, matching the non-segmented path's behaviour instead of rendering zero for the band.

### C — the mechanisms are now exercised

This was the substantive gap last round: both engineered halves could be deleted with every test still green. Re-ran the same mutations:

| mutation | before | now |
|---|---|---|
| disable the band preference at the call site | all tests passed | **fails** `test_perimeter_top_lane_carves_around_deep_local_obstacle` |
| strip `capacity_deficit` from `_top_lane_target` | (fix did not exist) | **fails** `test_top_lane_target_covers_capacity_deficit_on_other_bands` |

Baseline: `tests/test_balloon_ring_standoff.py` + `tests/test_layout.py` → 37 passed. The new coverage is in `test_make_drawing.py`, which is fast-tier — so these are genuine **PR gates**, not post-merge-only.

### Verdict: approve

Every finding from the three prior reviews is addressed, and each fix is now defended by a test that fails when the mechanism is removed. The design ends up cleaner than where it started: one halo constant shared between layout and render, a preference that expresses intent without overriding cost, and a lane target that reconciles balance with cardinality.

**One thing still not run:** CTC-04's own e2e assertions (only its standoff number). It is the other escalating fixture and the slow tier is post-merge — which is exactly how the `gap < 60` regression slipped past round one. Worth a single run before merge.

pzfreo:
**Correction to the section D code block above**, which the shell ate on the way out:

```
_balloon_halo()   = 21.5
_est_plan_halo()  = 21.5      # compose.py now delegates to _core._balloon_halo
balloons.py:73    perimeter_extent = _balloon_halo(fs) if perimeter else standoff + 2 * r
```

Verified identical and sharing one source, so the 22.5-vs-21.5 overrun is genuinely gone rather than coincidentally equal.

Issue #874, opened 2026-07-27
ADR 0016 PR6: a build must request a dimension source (breaking change)


pzfreo:
Child of #867.

Omission is only meaningful inside a set declared complete, so the dimension set gets exactly two sources and the script always says which:

- `auto_dimensions()` — the planner-selected set;
- `dimension(...)` declarations — the **complete** authored set.

Three errors: **mixing the two raises**; `add_dimension` without `auto_dimensions()` raises; and **a build that requested neither raises** rather than silently producing an undimensioned drawing. That last one is the project's standing preference for a loud failure over a plausible-looking, unbuildable sheet (consistent with #630/#631, and the #632 completeness lint).

Rejected: implicit-by-usage ('any `dimension` line turns the automatic set off') — a hand-author adding one pitch dimension would silently lose every ⌀ callout.

**Breaking change** to today's implicit-automatic `Sheet(part).build()`. Ship with PR5 so callers absorb one migration, not two.

**Gate:** all three errors fire; the emitter always writes one of the two source forms.

Issue #871, opened 2026-07-27
ADR 0016 PR3: DimensionId + the addressable-dimension handle


pzfreo:
Child of #867.

`DimensionId(feature, parameter)` where `parameter` is the `AddressableDimension`'s `ParameterId`, derivable from the plan so intent can *reference* an automatic dimension.

Durable *feature* identity is explicitly **not** minted here — structural equality on the IR feature suffices within and across script runs (ADR 0016, 'Feature identity is a separate sub-problem').

**Gate:** equal ids are interchangeable in lookup, and an id resolves against a rebuilt feature.

---

## Narrowed 2026-07-27 (during #887)

This issue originally also required wiring `DimensionId` into the ADR 0010 provenance seam so `drop`/`annotations_of` resolve through it, gated on the provenance audit. **That half is split to #886**, because implementing it surfaced a design fault rather than a wiring task:

the seam records `annotation name → one feature`, but one compound hole callout is a **single** placed leader rendering **five** addressable dimensions (bore ⌀ and depth, counterbore, spotface, countersink). So provenance keyed to one `DimensionId` per annotation is wrong by construction; the channel must become `name → tuple[DimensionId, ...]`, and a decision is needed about which passes populate it and what an unpopulated one means.

The identity half stands on its own and unblocks `add_dimension`'s idempotence (#872). The provenance half blocks the suppression children (#875/#876) and is tracked as **#886**.

Issue #894, opened 2026-07-27
Front-view dims commit to the strip instead of joining the corridor solve


pzfreo:
Slow-tier regression on `main` after #888: `test_ctc_ap203_meets_standards_no_degenerate_arcs[03]` and `[04]` lose **both** their step-height and overall-height dimensions.

```
placement_unsatisfiable: step-height dimension dropped
  (front-view right strip full; occupied by:
   m_pocket4_pos_long, m_pocket2_pos_long, m_pocket1_pos_long)
placement_unsatisfiable: overall height dimension dropped
```

## #888 is not at fault

It added occupants to a pre-existing hole.

`_place` has two paths. The corridor path (plan/side views, horizontal measure) registers a candidate and lets `drain_corridors` arbitrate. **Every other case — right/left dims, and any front-view above/below — fell to an immediate path calling `place_strip_candidates` directly.** That is a *local* solve: it spaces only the candidates in its own call against already-placed annotations, then commits. Whichever pass ran first won the space.

So #888's pocket location dims committed during the pocket pass, filled the front-view right strip, and the height dims arrived to find it full.

**The overall height carries `_MANDATORY_OVERALL_PRIORITY` (100.0) — the highest survival rank in the engine — and lost anyway**, because priority only ranks candidates that are *in* the solve. The arbitration never ran.

## The guard that should have caught it

`tests/test_carve_free_position_callers.py` (#636) is a fail-closed allowlist for exactly this bug class. Its own docstring names the failure mode: *"two 0.3.0 features, plates #559 and step positions #555, already regressed onto it before the migration — that is the failure mode this test prevents from recurring."*

But it guards only `carve_free_position`, the solver-**invisible** primitive. Direct `place_strip_candidates` is a *different* commit-now door and was never guarded. #636 locked one and left the other open.

Attribution of all 7 direct-call sites: none added by #888, #884, #891 or the ADR 0016 epic. They date from 3–17 July (#394, #478, #658, #671, #672), two of them legitimately inside the drain.

## Scope

1. Register every side into the corridor batch; opposite side becomes a post-drain `on_drop` fallthrough (the `render_plates` / `render_gdt` shape, and the #684 no-mid-drain-preemption rule).
2. Give the principal front-view chain a priority — it was registering at the implicit default and tying with feature location dims.
3. Extend the fail-closed guard to `place_strip_candidates` so the second door is shut.

Release blocker for 0.3.10, alongside #893 (balloon ring collapse — separate subsystem).

[...]

PR #905, opened 2026-07-28
feat(#872): add_dimension() — augmenting dimension intent on the Sheet facade

### LINKED ISSUE #872: ADR 0016 PR4: add_dimension() + the DimensionIntent handle

pzfreo:
`sheet.add_dimension(feature, role)` asks the planner to carry one more
measurement. Referential like every ADR 0016 intent: it names a feature and a
role and carries NO number, so the value still comes from the geometry and a
size lives in exactly one place. It changes SELECTION, not derivation — a
request can never introduce a number the part does not have.

The chain, bottom to top:

- `ir.RequestedDimension(feature, role, discriminator, pin, priority)` and
  `PartModel.requested_dimensions`. Kept deliberately distinct from
  `decorations`: a decoration enriches a dimension the planner already chose,
  a request changes WHICH dimensions it chooses. ADR 0016 calls this the
  planner's intent input.
- `plan_dimensions` consults requests and un-suppresses the named measurement,
  carrying `pin`/`priority` onto the `PlannedDimension` beside `convention`
  and `suppressed`.
- `Sheet.add_dimension(...)` returns a chainable `DimensionIntent` with
  `.pin()` / `.priority()`, plus `Sheet.auto_dimensions()`.
- `requested` threads through `build_drawing` -> `_coerce_model` ->
  `PartModel`, mirroring how `decorations` already flows.

Two design decisions worth recording.

**The call-site vocabulary accepts both a bare role and a dotted ParameterId.**
`add_dimension(bore, "bore")` emphasises every measurement under that role;
`add_dimension(bore, "bore.depth")` picks exactly one. The dotted form is the
identity #871 built; the short form is what a caller reaches for when the role
is unambiguous. ADR 0016 explicitly leaves this vocabulary open, and the first
implementation keyed on IR role alone, which made `add_dimension(bore, "bore")`
silently emphasise both the diameter and the depth.

**The auto_dimensions() requirement is checked at build(), not in the verb.**
add_dimension augments the planner's set so there must be one — but ADR 0016
requires intent to be order-independent, and a call-time check would make
`add_dimension(...)` before `auto_dimensions()` fail while the reverse
succeeded. Both orders now read the same.

`auto_dimensions()` itself stays OPTIONAL: a sheet with no augments still
builds without it, exactly as before. Making it mandatory is #874's breaking
change.

Also fixes an inconsistency found while wiring: `Sheet.model()` did not pass
requests to `_coerce_model`, so the public model-inspection surface would have
disagreed with what `build()` actually plans — the #707 class of divergence.

Mutation-tested before pushing rather than after review: gutting the planner's
request handling fails 3 tests; removing the build gate fails 1.

Closes #872.

---

**Epic:** #867 (ADR 0016 implementation), PR 4 of 8.

**Verification**
- `tests/test_add_dimension.py`: 17 new tests
- affected modules (`add_dimension`, `parameter_id`, `declare`, `tolerances`): 322 passed
- lint 4/4 clean
- **mutation-tested**: planner ignoring requests → 3 failures; build gate removed → 1 failure

**Not in this PR:** `pin`/`priority` reach `PlannedDimension` but are not yet consumed by the render layer as `CorridorCandidate.anchored`/`priority`. The existing pinned path is location-specific (`render_locations`, driven by `Drawing.finalize()`'s post-build intents), so wiring pre-build emphasis across the render passes is its own change. Filing separately rather than half-wiring it.

### DIFF
--- a/src/draftwright/builder.py
+++ b/src/draftwright/builder.py
@@ -287,6 +287,15 @@ def _assemble(
     # Detected path: reuse the model _analyse already built for sizing (#584 WP1 A) —
     # detectors run once per build (ADR 0008 Amdt 5, #602). build_model(a) remains the
     # fallback for a manually-constructed Analysis with no stored model.
+    if model is None and requested:
+        # A request names a DECLARED feature object (ADR 0016 / #872), and detection
+        # builds its own. Silently dropping them would leave a caller's add_dimension()
+        # with no effect and no diagnostic — the failure mode this project treats as
+        # worse than a visible error (#630/#631/#632).
+        raise ValueError(
+            "requested= names declared features, so it needs model= too; a detected "
+            "model builds its own feature objects that no request can target"
+        )
     pm = (
         _coerce_model(model, a.part, decorations, requested)
         if model is not None
--- a/src/draftwright/builder.py
+++ b/src/draftwright/builder.py
@@ -196,7 +196,7 @@ def _measure_blocks(dwg, a) -> dict:
 # ---------------------------------------------------------------------------
 
 
-def _coerce_model(model, part, decorations=None) -> PartModel:
+def _coerce_model(model, part, decorations=None, requested=None) -> PartModel:
     """Wrap a caller-supplied ``model=`` (ADR 0011) into a :class:`PartModel`. A
     ``PartModel`` is used verbatim; a sequence of features is wrapped with the part's
     bbox, a default corner location datum (matching ``detect.py``, so hole location
@@ -213,8 +213,14 @@ def _coerce_model(model, part, decorations=None) -> PartModel:
     ``PartModel`` is never mutated — decorations merge into a copy so the caller's
     reusable public input (ADR 0011) stays clean across builds."""
     if isinstance(model, PartModel):
-        if decorations:
-            return replace(model, decorations={**model.decorations, **decorations})
+        if decorations or requested:
+            return replace(
+                model,
+                decorations={**model.decorations, **decorations}
+                if decorations
+                else model.decorations,
+                requested_dimensions=tuple(requested) if requested else model.requested_dimensions,
+            )
         return model
     features = list(model)
     bbox = part.bounding_box()
@@ -226,6 +232,7 @@ def _coerce_model(model, part, decorations=None) -> PartModel:
         features=features,
         datums=[datum],
         decorations=decorations or {},
+        requested_dimensions=tuple(requested or ()),
     )
 
 
@@ -242,7 +249,15 @@ def detect_part_model(part, *, pmi="off") -> PartModel:
 
 
 def _assemble(
-    a, out, assembly, detail_view, auto_dims, model=None, decorations=None, trace=None
+    a,
+    out,
+    assembly,
+    detail_view,
+    auto_dims,
+    model=None,
+    decorations=None,
+    requested=None,
+    trace=None,
 ) -> Drawing:
     """Project the 4 views for analysis *a*, run the automatic annotation
     passes, and fit the iso.  This is pass 1 of :func:`build_drawing`; with a
@@ -273,7 +288,7 @@ def _assemble(
     # detectors run once per build (ADR 0008 Amdt 5, #602). build_model(a) remains the
     # fallback for a manually-constructed Analysis with no stored model.
     pm = (
-        _coerce_model(model, a.part, decorations)
+        _coerce_model(model, a.part, decorations, requested)
         if model is not None
         else (a.model if a.model is not None else build_model(a))
     )
@@ -419,6 +434,7 @@ def _repack(
     page=None,
     model=None,
     decorations=None,
+    requested=None,
     trace=None,
 ):
     """Measure the laid-out drawing's *real* per-view annotation footprints and,
@@ -557,6 +573,7 @@ def _repack(
         auto_dims=True,
         model=model,
         decorations=decorations,
+        requested=requested,
         trace=trace,
     )
     return a2, dwg2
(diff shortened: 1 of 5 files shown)


pzfreo:
## Reworked — the third pin mechanism is out, and #906 files the convergence

The first revision gave `DimensionIntent` `.pin()` / `.priority()`. Reviewing that surfaced something worth stopping for: the engine already carries **two** unconnected spellings of "keep this dimension put", and I was adding a third.

| # | mechanism | where | what it does |
|---|---|---|---|
| 1 | `Drawing.pin(name)` | `drawing.py:2249` | freezes a **placed annotation by name** — `repair` will not move it. Post-hoc, page-level. |
| 2 | `intents` `kwargs["pin"]` → `pinned_loc` → `render_locations(pinned=…)` | `drawing.py:1488` → `from_model.py:594` | ADR 0012 **post-build** intent. Sets `precedence=3`, `priority=MANDATORY`, and also calls `dwg.pin(nm)`. **Location dims only.** |
| 3 | pre-build request from `add_dimension` | this PR, now removed | — |

Two things made it worth removing rather than deferring:

- **It would have shipped a chainable verb that does nothing.** `(2)` reaches the solve only through `render_locations`, so a pre-build `pin` on a bore ⌀ or an envelope span had no consumer at all. Advertising `.pin()` on the handle would have been a promise the engine does not keep.
- **It is the #894 failure mode again** — one concept scattered across layers until no site shows it whole. That was the corridor priority scale before #894 consolidated it into one `PRIORITY` ladder beside the field it ranks.

So the emphasis fields are gone from all three layers (IR, planner, façade) and the convergence is **#906**, which also captures two things the audit turned up: `(2)` sets `priority=MANDATORY`, meaning a pinned location currently outranks authored GD&T — asserted at one call site rather than stated as policy — and `(1)` vs `(2)` are genuinely *different operations* (page-level freeze vs solve-time anchoring) sharing one verb name.

`DimensionIntent` stays as the extension point, so #906 restores `.pin()` / `.priority()` onto it once there is a single carrier to honour them.

**What this PR now ships:** pure selection intent. `add_dimension` asks the planner to carry one more measurement; it changes *which* dimensions are chosen, never what they read.

14 tests, lint 4/4, 263 passed across the affected modules. Mutation-tested: gutting the planner's request handling fails 3 tests, removing the build gate fails 1.

pzfreo:
## Seven review rounds, and the real finding is #908

Landing this as scoped. The review history is worth recording, because it ends somewhere different from where it started.

### What the rounds found

| round | finding |
|---|---|
| 1 | Tests never exercised the feature — deleting the un-suppression logic left all 20 passing |
| 1 | `_request_for` matched on **value** equality, so two equal features were one target |
| 1 | `requested=` silently discarded without `model=` |
| 2 | Re-validation checked only the role, so swapping two holes retargeted silently |
| 2 | `_feature_index` accepted a handle from **another Sheet** |
| 3 | My round-2 fix laundered a reorder: `_replace_feature` advanced on index alone |
| 4 | Declaring the intent **after** a reorder — invisible to any materialisation check |
| 5 | A reorder **before** the first intent; and my guard **over-rejected** legitimate edits |
| 6 | A size verb laundered a stale handle; and the shadow broke `hole.note()` + `add_dimension` |
| 7 | Every non-size verb still mis-targets — 13 of them |

### The actual cause

Not `add_dimension`. **`Sheet` addresses features by index into a public mutable list**, so every long-lived reference is positional where it needs to be identity-based.

Measured with no `add_dimension` anywhere:

```python
a = s.hole(diameter=10, ...).depth(12)
s.hole(diameter=4, ...).depth(7)
a.tolerance(0.05)
s.features.reverse()
s._decorations()      # -> 0.05 lands on the diameter-4 hole
```

`_tolerances` has been index-keyed since P2a; `_gdt_src` and `section` likewise. This PR did not create the class, it **surfaced** it — being the first feature to hold a reference to a specific feature and act on it visibly.

Round 7 independently enumerated the blast radius: `tolerance`, `fit`, `finish`, `note`, `knurl`, `_Params.tolerance`, GD&T `datum`/`control`, sheet-level `finish`/`note`, `section`.

Filed as **#908** with both candidate fixes (handle-as-identity with `features` derived, or a read-only `features` plus explicit verbs). Either makes the class *impossible* rather than detectable, and fixes tolerances and GD&T at the same time.

### What ships here

Selection intent, and two narrow checks marked in-code as scaffolding for #908 — they cover this PR's own paths, and round 7 found **no false positives** across the full fluent surface after the round-6 fix.

38 tests. Every mechanism mutation-verified, with the mutations now asserting their own anchor exists first — an earlier one silently edited a string that was not there and its unchanged failure read as a passing check.

### What I would do differently

Six rounds of patching before asking whether the design was wrong. The signal was there by round 3, when my own fix opened the next hole; I kept treating each variant as the last one instead of reading the pattern.

Issue #907, opened 2026-07-28
Extend the fail-closed caller guard to place_strip_candidates


pzfreo:
Split out of #894, which fixed the symptom but left the door open.

## The gap

`tests/test_carve_free_position_callers.py` (#636) is a fail-closed allowlist that stops new callers regressing onto solver-invisible placement. Its docstring names the exact failure mode it prevents:

> two 0.3.0 features, plates #559 and step positions #555, already regressed onto it before the migration — that is the failure mode this test prevents from recurring.

But it guards only `carve_free_position`. **Direct `place_strip_candidates` is a second commit-now door and is unguarded.**

The distinction:

| call | behaviour |
|---|---|
| `register_corridor(...)` + drain | the ADR 2 (was 0014) global collect-then-solve — the candidate is arbitrated against every other occupant |
| `place_strip_candidates(...)` **directly** | a *local* solve: spaces only the candidates in that call against already-placed annotations, then **commits**. Whichever pass runs first wins the space. |
| `carve_free_position(...)` | fully solver-invisible — **guarded** |

#894 was exactly this: #888's pocket location dims committed during the pocket pass, filled the front-view right strip, and the height ladder found it gone. The overall height carries the engine's top survival rank and lost anyway, because priority only ranks candidates that are *in* the solve.

## Why it was not done in #894

The allowlist must be **clean** when the guard lands — `test_architecture_docs.py` explicitly forbids "pending migration" entries:

```python
def test_carve_guard_points_to_current_placement_adr():
    guard = _read("tests/test_carve_free_position_callers.py")
    assert "Pending migration (#636)" not in guard
```

So the remaining commit-now sites have to migrate first. #894 migrated the front view only, and deliberately: routing the plan/side right-left dims through the corridor cost #885's part two pad size dims, and deferring its fallthrough cost a third — both measured, not inferred. The corridor carve is strictly more conservative than the commit path, because it marks a region obstructed from FULL annotation geometry where a label can legitimately sit between two extension lines.

## Scope

1. Migrate the remaining direct `place_strip_candidates` callers, or record each as an ADR 2 (was 0014) exemption with a rationale. Attribution at the time of #894 — none added recently; all date from 3–17 July (#394, #478, #658, #671, #672), two of them legitimately inside the drain itself.
2. Make the carve reason about label boxes where a full-geometry obstruction is over-conservative — this is what unblocks the plan/side migration without losing coverage.
[...]

PR #910, opened 2026-07-28
refactor(#908): features carry identity, so handles stop addressing by position

### LINKED ISSUE #908: Sheet handles address features by index into a public mutable list
Found across seven review rounds on #905. Every round found one more variant of the same failure, which is the signature of guarding a broken invariant rather than fixing it.

## The flaw

`Sheet` exposes `features` as a **public mutable list**, and every fluent handle (`_Hole`, `_Dim`, `_Params`) addresses its feature by **index**. So every long-lived reference is positional where it needs to be identity-based. Insert, delete or reorder the list and the reference silently points at a neighbour.

## This predates #872

Measured with no `add_dimension` anywhere:

```python
s = Sheet(part)
a = s.hole(diameter=10, ...).depth(12)
b = s.hole(diameter=4,  ...).depth(7)
a.tolerance(0.05)          # tolerance on the 10mm hole
s.features.reverse()
s._decorations()           # -> 0.05 lands on the diameter-4 hole
```

`_tolerances` is index-keyed and has been since P2a. So are `_gdt_src` and the `section` feature reference. #872 did not introduce this class — it **surfaced** it, being the first feature to hold a long-lived reference to a specific feature and act on it visibly.

## Blast radius

An independent adversarial pass enumerated the affected verbs. Reverse `features`, then use a retained handle, and each silently targets whatever now occupies the old index:

- `_Hole.tolerance`, `.fit`, `.finish`, `.note`
- `_Dim.tolerance`, `.fit`, `.finish`, `.note`, `.knurl`
- `_Params.tolerance`, `.note`
- GD&T `datum` / `control`, sheet-level `finish` / `note`
- `section(feature=…)`

Its probe: create ⌀10 and ⌀6 holes, keep the ⌀10 handle, `features.reverse()`, then `old.fit("H7")` — succeeds, stores `(0, "diameter")`, applies the fit to the ⌀6 hole.

Assessment from that pass: *"a plausible variant, not active sabotage: the public list remains documented as mutable, and retaining a fluent handle while editing that list is natural."*

## Why patching does not work

#905 added three detection mechanisms across six rounds. Each closed one route and opened another:

- round 3's fix let a stale handle be laundered through a size verb;
- round 5's fix broke a legitimate flow (`hole.note(...)` + `add_dimension` raised spuriously) because a GD&T rebind wrote the list directly;
- round 6's fix still left every non-size verb unguarded.

Detection after the fact cannot be complete when the invariant itself is positional.

## The fix

Give features **stable identity**. Two shapes worth weighing:

1. **The handle is the identity.** `Sheet` stores an ordered token → feature map; `features` becomes a derived view. A handle holds its token, so list order is irrelevant and reordering simply works rather than needing detection.
2. **`features` stops being a raw mutable list.** A read-only view plus explicit `remove(handle)` / `replace(handle, …)` verbs, so there is no way to desync.

Either makes the whole class **impossible** rather than detectable, and fixes tolerances, GD&T provenance and sections at the same time — none of which #905's guards reach.

## Interim

#905 keeps the identity check at materialisation and the stale-handle check on size verbs. They are narrow, mutation-verified, and carry no known false positives — but they are scaffolding, and should be **removed** by whichever fix lands here rather than extended to thirteen more call sites.

pzfreo:
`Sheet` exposed `features` as a public mutable list and every fluent handle
addressed its feature by INDEX. So every long-lived reference — a tolerance, a
GD&T origin, a section, a dimension intent — was positional where it needed to
be identity-based. Reorder the list and the reference silently pointed at a
neighbour.

This predates the ADR 0016 work. Measured with no add_dimension involved:

    a = s.hole(diameter=10, ...).depth(12)
    s.hole(diameter=4, ...).depth(7)
    a.tolerance(0.05)
    s.features.reverse()
    s._decorations()        # -> 0.05 landed on the diameter-4 hole

`_tolerances` has been index-keyed since P2a. Seven review rounds on #872 found
seven variants of the same failure, each fix opening the next, because
detection after the fact cannot be complete when the invariant is positional.

`features` is now a `_FeatureView` over `(token, feature)` entries:

- append mints a token — a declaration;
- `features[i] = f` KEEPS the token, because that is what a size verb does when
  it rebuilds the frozen dataclass;
- reverse/sort move ENTRIES, so a handle finds its feature at the new position;
- del drops the token, so a handle for a removed feature raises when used.

Handles hold a token and derive `_i` on read. Tolerances, GD&T provenance,
sections and dimension intents are all token-keyed.

The result is that a reorder WORKS rather than raising — the outcome detection
could never reach. Verified: fit, tolerance, handle resolution and a size verb
after a reorder all land on the intended feature.

Two things measurement caught that reasoning did not:

MutableSequence implements `reverse` via `__setitem__`, which here keeps each
slot's token — right for replacement, wrong for reordering. Without overriding
it, reversing swapped the features and left every token on its old slot: the
exact bug, reintroduced by the fix. `reverse`/`sort` now move entries.

Tokens were process-global at first, so internal keys depended on how many
sheets other tests had built. They are per-sheet now, which is deterministic
and matches this project's stance on reproducible output.

Removes the three staleness guards #872 added as scaffolding, and the
hand-written intent bookkeeping in `_replace_feature` — that method is now a
plain slot write, because the view carries identity for it.

Closes #908.

---

**Verification**
- Full fast tier: **1827 passed**
- Lint 4/4 clean
- The motivating probe now lands correctly: `fit`, `tolerance`, handle resolution and a size verb after a reorder all target the intended feature
- `test_sheet_of`'s token-key assertion passes **unchanged** once tokens are per-sheet

**Scaffolding removed:** the three staleness guards from #905, plus `_replace_feature`'s hand-written intent bookkeeping (two of the seven #872 findings were bugs in exactly that bookkeeping).

### DIFF
--- a/src/draftwright/sheet.py
+++ b/src/draftwright/sheet.py
@@ -46,7 +46,6 @@ import math
 import warnings
 from collections.abc import MutableSequence
 from dataclasses import replace
-from itertools import count
 
 from draftwright.analysis import _solids_body
 from draftwright.builder import _coerce_model, build_drawing, detect_part_model
@@ -129,8 +128,10 @@ class _FeatureView(MutableSequence):
     the feature it names:
 
     - ``append`` mints a new token — a declaration;
-    - ``features[i] = f`` **keeps** the token, because that is what a size verb does when
-      it rebuilds the frozen dataclass;
+    - ``features[i] = f`` mints a **new** token: assignment cannot express "move this
+      feature here", so inheriting the slot's identity would silently hand every
+      reference to whatever was assigned. Stale references fail loudly instead. The
+      identity-preserving rebuild a size verb needs is :meth:`_rebind`;
     - ``reverse`` / ``sort`` / ``insert`` reorder entries, so a handle simply finds its
       feature at the new position rather than silently naming a neighbour;
     - ``del`` drops the token, so a handle for a removed feature raises when used.
@@ -142,14 +143,22 @@ class _FeatureView(MutableSequence):
     instead of detectable.
     """
 
-    __slots__ = ("_entries", "_tokens")
+    __slots__ = ("_entries", "_next_token")
 
     def __init__(self, entries: list) -> None:
         self._entries = entries
-        # Per-sheet, not global: tokens appear in internal keys, and this project holds
-        # output to be deterministic — a process-wide counter would make those keys
-        # depend on how many other sheets happened to be built first.
-        self._tokens = count()
+        # A plain int, not `itertools.count`: counters lose pickle/copy support in
+        # Python 3.14 and this package supports >=3.11, so a Sheet carrying one would
+        # stop being copyable. Per-sheet rather than global because tokens appear in
+        # internal keys and this project holds output to be deterministic — a
+        # process-wide counter would make those keys depend on how many other sheets
+        # happened to be built first.
+        self._next_token = 0
+
+    def _mint(self) -> int:
+        token = self._next_token
+        self._next_token += 1
+        return token
 
     def __getitem__(self, i):
         if isinstance(i, slice):
@@ -175,9 +184,9 @@ class _FeatureView(MutableSequence):
         # `reverse()` / `sort()`, which move whole entries. Internal rebuilding goes
         # through `_rebind`, which is the only path that preserves a token.
         if isinstance(i, slice):
-            self._entries[i] = [(next(self._tokens), f) for f in value]
+            self._entries[i] = [(self._mint(), f) for f in value]
             return
-        self._entries[i] = (next(self._tokens), value)
+        self._entries[i] = (self._mint(), value)
 
     def _rebind(self, index: int, feature) -> None:
         """Replace the feature at *index* KEEPING its token — the same feature, rebuilt.
@@ -194,7 +203,7 @@ class _FeatureView(MutableSequence):
         return len(self._entries)
 
     def insert(self, i, value) -> None:
-        self._entries.insert(i, (next(self._tokens), value))
+        self._entries.insert(i, (self._mint(), value))
 
     # Reordering must move ENTRIES, not values. `MutableSequence` implements `reverse`
     # in terms of `__setitem__`, which here keeps each slot's token — right for a size
@@ -1103,14 +1112,19 @@ class Sheet:
 
     @property
     def features(self) -> MutableSequence:
-        """The declared IR features (mutable — override or drop before :meth:`build`).
+        """The declared IR features — mutable: override, drop or reorder before
+        :meth:`build`.
 
-        **Not after declaring an `add_dimension` intent.** An intent targets one specific
-        feature, and this list is addressed by index, so an insert, delete or reorder
-        silently repoints it at a neighbour. Both are refused rather than guessed:
-        editing the list after an intent exists raises at :meth:`build`, and a handle
-        whose index no longer names the feature it was issued for raises when used.
-        Declare the features you want, then the dimensions.
+        Each feature carries an identity token (#908), so a **reorder** via
+        :meth:`~_FeatureView.reverse` or :meth:`~_FeatureView.sort` moves every reference
+        with it — a handle, a tolerance, a GD&T origin, a section, an `add_dimension`
+        intent all follow their feature to its new position.
+
+        **Assignment is not a move.** ``features[i] = f`` and slice assignment mint a new
+        identity, because assignment cannot distinguish "move this feature here" from
+        "put a different feature here" — so references to the displaced feature fail
+        loudly rather than silently transferring to whatever replaced it. The same holds
+        for deletion. Use ``reverse``/``sort`` to reorder while keeping references.
         """
         return self._features
 
--- a/src/draftwright/sheet.py
+++ b/src/draftwright/sheet.py
@@ -44,7 +44,9 @@ from __future__ import annotations
 
 import math
 import warnings
+from collections.abc import MutableSequence
 from dataclasses import replace
+from itertools import count
 
 from draftwright.analysis import _solids_body
 from draftwright.builder import _coerce_model, build_drawing, detect_part_model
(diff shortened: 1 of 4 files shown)


pzfreo:
There is one remaining positional-identity leak in `_Control`.

`Sheet.control(ref)` resolves the source to an index and `_Control.__init__` retains that index in `self._src`. If the public feature view is reordered after creating the builder but before adding a characteristic, `_Control._add()` passes the stale index to `_append_gdt()`, which converts the feature now occupying that position into a token. The frame is then silently rebound to the wrong feature — the exact invariant this PR is intended to remove.

Confirmed on this PR's current head:

```python
from build123d import Box
from draftwright import Sheet

s = Sheet(Box(40, 40, 10))
a = s.hole(diameter=10, at=(-10, 0, 0), axis="z")
s.hole(diameter=4, at=(10, 0, 0), axis="z")

control = s.control(a)
s.features.reverse()
control.position(0.1)
s._prepare()

cf = next(f for f in s.features if f.kind == "control_frame")
assert cf.origin.diameter == 10  # actual: 4
```

The leak is across `Sheet.control()` → `_Control._src` → `_Control._add()` → `_append_gdt()`.

Please have `_Control` retain the source **token**, not an index. More generally, `_gdt_ref()` should ideally return stable identity for any value that can outlive the immediate call. Add a regression test that separates `control()` from `.position()` (or another characteristic) with `features.reverse()`/`sort()` between them.

pzfreo:
## Review summary — two adversarial rounds, converged

### Round 1 — two P1 identity escapes

The first draft still retargeted silently through the two most idiomatic reorderings:

```python
features[0], features[1] = features[1], features[0]
features[:] = features[::-1]
```

Both move **values** between slots, and `__setitem__` inherited the destination slot's token — so every reference followed the slot rather than the feature. Overriding `reverse`/`sort` did not help, because neither goes through them. A third hole: `features[i] = unrelated` quietly inherited the displaced feature's identity.

**Cause:** I had conflated two operations. Keeping a token is right for the *internal* rebuild a size verb does — `.depth()` replaces the frozen dataclass with an updated copy of the **same** feature. It is wrong on the public view, where assignment cannot distinguish *move this feature here* from *put a different feature here*.

They are now separate:

| operation | identity |
|---|---|
| `append` / `extend` / `insert` / `+=` | mints fresh |
| `features[i] = f`, slice, extended slice | **mints fresh** — stale references fail loudly |
| `del` / `pop` / `remove` / `clear` | drops |
| `reverse()` / `sort()` | **preserves** — entries move, references follow |
| `_rebind` (internal, size verbs only) | **preserves** |

Measured across all four routes: swap, slice permutation and unrelated replacement all raise; `reverse()` carries the tolerance to the right feature.

### Round 2 — no identity escapes

A systematic audit of every `MutableSequence` method found the model sound. That enumeration is precisely what I had failed to do three times while patching one route at a time. Three smaller findings:

- **`itertools.count` loses pickle/copy support in Python 3.14**, which this package supports and CI tests — a `Sheet` carrying one would stop being copyable, and `_Params` already anticipates copy/pickle use. Now a plain int; `deepcopy` verified.
- **Docstrings described the behaviour this PR replaces** — the view claimed `features[i] = f` *keeps* the token, and the public `features` property still documented the #905 regime where edits after an intent were refused. Both would have told callers the wrong workflow.
- **The audit's conclusions lived only in a review comment.** `TestFeatureViewContract` makes them executable: additions, removals, `sort`, extended-slice per-slot behaviour, negative indices, `deepcopy`, and detection-seeded features from `from_part`.

### Verification

- Full fast tier **1845 passed**; lint 4/4
- Both halves of the public/internal split mutation-verified (preserving in `__setitem__` fails 2 tests; minting in `_rebind` fails 4), with mutations asserting their own anchor exists first
- The motivating P2a bug is fixed: a tolerance declared on the ⌀10 hole still lands on it after a reorder

### What this removes

The three staleness guards from #905 and `_replace_feature`'s hand-written intent bookkeeping — two of the seven #872 review findings were bugs in that bookkeeping. Detection replaced by a model where the failure cannot occur.

pzfreo:
Confirmed and fixed in `9aae688` — thank you, that is a real leak and my audit missed it.

### Why the audit missed it

I enumerated `MutableSequence` methods and the three handle classes that grew `_token`. `_Control` derives from none of them, so it never appeared in either enumeration — I checked the *sites that store identity* and not the *sites that consume it*.

The distinguishing property is temporal. `datum`/`finish`/`note` resolve a ref and append in a single expression:

```python
target, src = self._gdt_ref(ref)
self._append_gdt(_declare_datum(letter, target, ...), src)
```

Nothing can mutate the view between those two lines. `control()` returns a **builder**, and the frame is appended only when a characteristic method runs — arbitrarily later. That gap is what your repro drives through, and `_Control` is the only object in the module that spans one.

### The fix

The conversion moved from *append* time to *resolve* time. `_gdt_ref` returns a token rather than an index, and `_append_gdt` accepts one, so the gap becomes unrepresentable — there is no index anywhere for a mutation to invalidate. Two consequential tidies:

- `section(feature=…)` already converted at resolve time and drops its now-redundant `_token_at`.
- `_gdt_finish` / `_gdt_note` convert at their own call site, which is also their resolve.

Your repro now yields `10`. A `sort()` variant and an append-then-reorder variant do too.

### Also found while fixing

`_Params.__getattr__`'s recursion guard still named `_i`, which this PR turned into a property — so it never routes through `__getattr__` at all, while `_token`, the actual field, was unguarded. An unpickled handle asking for `_token` would have forwarded to the owning `Sheet` rather than raising. Guard corrected.

### Verification

- Two tests added to `TestFeatureViewIdentity`, both mutation-verified: restoring the index round-trip fails them, and each asserts its anchor exists before mutating so the check cannot go vacuous
- `sheet`/`gdt`/`dimension`/`declare` selection: **525 passed**; lint 4/4 clean

### The pattern worth keeping

Every leak this PR has found sits at a **temporal gap between resolving a reference and using it** — not at the mutation sites I kept auditing. `_Control` is the last one in `sheet.py`, but that is a property of the current code rather than a guarantee. I'd rather it were enforced: I can add a check to #907's guard that no attribute stores an index into `_features`, so a future builder-style verb cannot reintroduce this class. Say the word and I'll file it.

pzfreo:
Filed as #912 — with one correction to what I said above: I referred to "#907's guard", but #907 guards the **placement** primitives (`carve_free_position` / `place_strip_candidates`). Identity is unrelated, so #912 is a new sibling in that fail-closed family rather than an extension of it.

The proposal changed shape once I looked at how the existing guards work. They AST-walk for a syntactic marker, and the marker here would be "an attribute assigned from an index into `_features`" — but *is this value an index* isn't decidable from the AST. A name-based proxy (`_src`, `_index`) is evaded by naming the field something else, which is exactly the way it would happen by accident.

So #912 proposes a **behavioural** guard instead, since the invariant is observable:

> for every public `Sheet` verb returning something other than `Sheet`, driving that returned object across a `features` reorder must produce the same `PartModel` as driving it without one.

Fail-closed by asserting the set of handle-returning verbs equals a committed set, so a new builder-style verb fails until it's added to the table and therefore exercised. Same ratchet as the allowlists, keyed on behaviour rather than syntax — and it catches a stored index under any name. Acceptance is mutation against `_Control`'s pre-#910 shape: if the guard passes on the buggy commit it's theatre.

CI on this PR: **9 passed, 7 pending, 0 failed.**

Issue #906, opened 2026-07-28
Three spellings of 'pin this dimension' — converge them


pzfreo:
Found while reviewing #905 (ADR 4 (was 0016) `add_dimension`). The engine carries **three** unconnected ways to say "keep this dimension where it is", at three layers, with no shared vocabulary.

| # | mechanism | where | what it does |
|---|---|---|---|
| 1 | `Drawing.pin(name)` | `drawing.py:2249` | pins a **placed annotation by name** — `repair` will not move it. Post-hoc, page-level. |
| 2 | `intents` `kwargs["pin"]` → `pinned_loc` → `render_locations(pinned=…)` | `drawing.py:1488` → `from_model.py:594` | ADR 2 (was 0012) **post-build** intent. Sets `precedence=3`, `priority=PRIORITY.MANDATORY`, **and** calls `dwg.pin(nm)` — so it also feeds (1). **Location dims only.** |
| 3 | a **pre-build** request from `Sheet.add_dimension(...)` | would be #905 | — |

(3) was implemented and then **removed from #905** rather than shipped: a fourth entry point advertising `.pin()` / `.priority()` that no renderer consumes is worse than not having it, and it would have entrenched the divergence.

## Why this matters

This is the same failure mode as the corridor priority scale before #894: one concept scattered across layers so that no one site shows the whole ladder, and a new contributor cannot tell which mechanism to reach for. #894's fix was a single `PRIORITY` ladder in `_common.py` beside the field it ranks.

Concretely today:

- (2) reaches the solve **only** for location dims. There is no route for pinning a bore ⌀, a step length, or an envelope span from an intent.
- (2) sets `priority=MANDATORY` — the top of the #894 ladder — so a pinned location outranks authored GD&T. That may be right, but it is asserted in one call site rather than stated as policy.
- (1) is position-freezing after the fact; (2) is solve-time anchoring. They are genuinely different operations sharing one verb name.

## What convergence looks like

ADR 2 (was 0012)'s model is that a pin is `CorridorCandidate.anchored` — "stays put while the rest flows around it" — with `priority` as the separate over-capacity rank. So:

- one carrier for emphasis that any pass can honour, not a location-dim-only parameter;
- `Sheet.add_dimension(...).pin()` and the post-build `dimension(..., pin=True)` reaching that **same** carrier;
- `Drawing.pin(name)` kept as the distinct page-level freeze it actually is, and named so it does not read as the same operation.

## Acceptance

[...]

PR #913, opened 2026-07-28
refactor(#912): one identity resolver, and a guard that fails closed on the next one

### LINKED ISSUE #912: Fail-closed guard: no Sheet object may store a positional index into _features
Split out of #910, which fixed the last known instance but leaves the class open.

## The class of bug

#908 gave features stable tokens so handles stop addressing by position. #910 then found a leak the migration missed — `_Control` stored a resolved index and converted it to a token at *append* time, so a reorder between `control(bore)` and `.position(0.1)` silently rebound the frame to whatever occupied the slot.

Both bugs share one shape, and it is **not** the mutation site:

> a reference is **resolved to an index**, then **used later**, with a window in between where the caller may mutate `features`.

That framing predicts where the next one appears. `datum`/`finish`/`note` resolve and append in a single expression and were never at risk. `_Control` was the only object in `sheet.py` spanning a window — and it was missed precisely because it derives from none of the handle classes that grew `_token`, so it appeared in neither enumeration I audited.

It is the last such window **today**. Nothing prevents the next builder-style verb from opening another, and the failure is silent: a wrong-but-plausible drawing, no exception.

## Why not the obvious AST guard

The sibling guards (`test_drawing_encapsulation.py`, `test_carve_free_position_callers.py`) AST-walk for a syntactic marker — `dwg._x = ...`, a call to a named function. The syntactic form here would be "an attribute assigned from an index into `_features`", and *"is this value an index"* is not decidable from the AST. A name-based proxy (`_src`, `_index`, `_i`) is trivially evaded by naming the field something else, which is the evasion most likely to happen by accident.

## Preferred: a behavioural guard

Cover the class directly, since the invariant is observable:

> For every public `Sheet` verb that returns something other than `Sheet`, exercising that returned object across a `features` reorder must produce the same `PartModel` as exercising it without one.

Concretely: for each such verb, build a two-feature sheet, call the verb, `features.reverse()`, drive the returned object, `_prepare()`, and assert the model matches the un-reordered run. `_Control` fails this before #910 and passes after.

Fail-closed via enumeration: assert the set of handle-returning verbs equals a committed set, so a **new** builder-style verb fails until it is added to the table and therefore exercised. That is the same ratchet as the allowlists, keyed on behaviour rather than syntax — and it catches an index stored under any name.

## Scope

1. Enumerate the handle-returning `Sheet` verbs; commit the set.
2. The reorder-invariance check, table-driven over it.
3. The fail-closed enumeration assertion, with the rationale above in the module docstring.
4. Mutation-verify against `_Control`'s pre-#910 shape — restoring the index round-trip must fail the new test, or the guard is theatre.

Step 4 is the acceptance criterion. A guard that passes on the buggy commit proves nothing (see the canary lesson in #737).

## Note

Also worth stating in the guard docstring: `section(feature=…)` and `_gdt_finish`/`_gdt_note` convert at resolve time and are correct *by construction*, not by accident. Someone "simplifying" them back to deferred conversion would reintroduce this.

pzfreo:
## The invariant

> Every long-lived reference to a feature is identity-based, never position-based. Reordering preserves the target; replacing or deleting has deliberate, documented semantics; no retained builder, intent, tolerance, GD&T origin, section request or decoration may silently retarget a neighbour.

#908 gave features tokens. #910 found the leak the migration missed. Both were the same shape — and it is **not** the mutation site I kept auditing:

> a reference is **resolved to an index**, then **used later**, with a window in between where the caller may mutate `features`.

## Part 1 — one resolver

Three ref-resolution paths had grown independently, and they disagreed:

| resolver | handle | int | Feature | b123d obj | cross-sheet check |
|---|---|---|---|---|---|
| `_index_of` (`of()`) | ✗ | ✓ negatives | ✓ | ✓ match | ✗ |
| `_gdt_ref` | ✓ | ✓ | ✓ | passthrough | ✓ |
| `_feature_index` (`add_dimension`) | ✓ | ✗ negatives | ✓ | ✗ | ✓ |

One public concept — *name a declared feature* — with three answers, two wordings of the cross-sheet error, and `of(handle)` failing as `AttributeError: '_Hole' object has no attribute 'bounding_box'` because it fell through to the object matcher. `_gdt_ref`'s was the one that leaked.

`_declared_token(ref, *, verb)` is now the single identity-bearing path. It returns a **token, never an index**, so the temporal gap is unrepresentable rather than merely unused. `None` means *not a declared reference*, and each caller decides what that means — `of()` matches the object, the GD&T verbs take a bare geometric target, `add_dimension` raises.

**User-visible:** `of(handle)` now works, negative indices are uniform, one cross-sheet message.

## Part 2 — a guard that fails closed

Behavioural, not another AST guard. Its siblings walk the AST because their marker *is* syntactic; this one's would be "an attribute assigned from an index into `_features`", and *is this value an index* is not decidable statically. A name-based proxy (`_src`, `_index`, `_i`) is evaded by naming the field something else — precisely how #910 happened.

The oracle is `Sheet.model()`. Two runs differing only by an identity-preserving mutation must agree; retargeting shows up there regardless of which layer leaked it.

- every handle-returning verb x 6 mutations (reverse, two sorts, double-reverse, added neighbour, neighbour moved in front)
- deliberate invalidation: delete / pop / clear / assign / tuple-swap / slice
- cross-sheet rejection on four verbs; token determinism; copy + pickle
- idempotent re-materialisation

**Three ratchets** — a new handle-returning verb, a new handle *class*, or a fork of the resolver each fail until an author states what the new thing does across a reorder. That is the question #910 went three rounds without being asked.

## Verification

Acceptance is **mutation against the pre-#910 shape**, not a green run. Restoring `_Control`'s stored index fails exactly 5 cases, all `control`; `double_reverse` correctly passes as a net no-op. Each ratchet independently verified to fail-close.

The resolver ratchet's first draft was **vacuous** — it searched source text, and every resolver's docstring names `_declared_token`, so the substring was there whether or not the call was. Now AST-based.

601 targeted tests pass; lint 4/4.

Closes #912.

### DIFF
--- a/src/draftwright/sheet.py
+++ b/src/draftwright/sheet.py
@@ -634,9 +634,10 @@ class Sheet:
         # P2a ± tolerances, keyed by (feature index, ParamKind) so a handle survives a later
         # feature replacement (e.g. hole().depth()); materialized to (feature, kind) at build.
         self._tolerances: dict = {}
-        # P2c GD&T provenance: (gdt_feature_index -> source_feature_index). A finish/datum stores
-        # its origin by INDEX, not the object, so a later size verb replacing the source feature
-        # (hole().depth()) doesn't strand the link; materialized to the FINAL object at build.
+        # P2c GD&T provenance: (gdt_feature_token -> source_feature_token). A finish/datum stores
+        # its origin by TOKEN, not the object, so a later size verb replacing the source feature
+        # (hole().depth()) doesn't strand the link, and a reorder cannot retarget it (#910);
+        # materialized to the FINAL object at build.
         self._gdt_src: list = []
         # Corner-block tables (notes / revision / BOM / schedule) — applied at build() via the
         # engine's generic auto-placed Drawing.add_table, AFTER the drawing is built so they sit
@@ -742,11 +743,12 @@ class Sheet:
     def of(self, ref) -> _Hole | _Dim:
         """A decoratable handle onto an **existing** feature — the hybrid seam (#463).
 
-        *ref* is a feature index, a :class:`Feature` already in :attr:`features` (e.g. seeded by
-        :meth:`from_part`), or the build123d **object** you built (matched by ⌀ + in-plane
-        position). Returns the same fluent handle the declaration verbs do, so you can
-        ``.fit(...)`` / ``.tolerance(...)`` — and, for a hole, ``.cbore(...)`` — a feature you
-        did not declare from scratch. Raises if the object matches no feature or is ambiguous."""
+        *ref* is a fluent handle this sheet issued, a feature index, a :class:`Feature` already
+        in :attr:`features` (e.g. seeded by :meth:`from_part`), or the build123d **object** you
+        built (matched by ⌀ + in-plane position). Returns the same fluent handle the declaration
+        verbs do, so you can ``.fit(...)`` / ``.tolerance(...)`` — and, for a hole, ``.cbore(...)``
+        — a feature you did not declare from scratch. Raises if the object matches no feature or
+        is ambiguous."""
         i = self._index_of(ref)
         kind = self._features[i].kind
         if kind == "hole":
@@ -755,18 +757,47 @@ class Sheet:
             return _Dim(self, i, "diameter" if kind == "boss" else "length")
         raise ValueError(f"of(): no aspect handle for a {kind!r} feature (holes / bosses / steps)")
 
-    def _index_of(self, ref) -> int:
-        if isinstance(ref, bool):
-            raise TypeError("of(): ref must be an index, a Feature, or a build123d object")
+    def _declared_token(self, ref, *, verb: str) -> int | None:
+        """The token of the declared feature *ref* names, or ``None`` if it names none.
+
+        The ONE identity-bearing resolution path (#912). Before this, three of them had grown
+        independently — ``of()``'s, the GD&T seam's and ``add_dimension``'s — accepting
+        different subsets of ref kinds, disagreeing on negative indices, and reporting the
+        cross-sheet error in two different wordings. `_gdt_ref`'s was the one that leaked
+        identity in #910, which is the argument for having a single one to audit.
+
+        Returns a **token**, never an index: a caller that stores the result cannot then be
+        invalidated by a reorder. ``None`` means *not a reference to a declared feature* — a
+        build123d face, or a :class:`Feature` from elsewhere. Each caller decides what that
+        means: :meth:`of` falls through to matching the object, the GD&T verbs treat it as a
+        bare geometric target, :meth:`add_dimension` raises.
+        """
+        if isinstance(ref, bool):  # bool is an int subclass; `of(True)` is a mistake, not index 1
+            raise TypeError(f"{verb}: ref must be a handle, an index, a Feature, or an object")
+        if isinstance(ref, (_Hole, _Dim, _Params)):
+            if ref._sheet is not self:
+                raise ValueError(
+                    f"{verb}: {type(ref).__name__} belongs to a different Sheet — a handle "
+                    "names a feature on the sheet that issued it, and this is not that sheet"
+                )
+            return ref._token
         if isinstance(ref, int):
             n = len(self._features)
             if not -n <= ref < n:
-                raise IndexError(f"of(): feature index {ref} out of range (have {n})")
-            return ref % n
+                raise IndexError(f"{verb}: feature index {ref} out of range (have {n})")
+            return self._token_at(ref % n)
         if isinstance(ref, Feature):
             for i, f in enumerate(self._features):
-                if f is ref:
-                    return i
+                if f is ref:  # identity — an EQUAL feature from elsewhere is not this one
+                    return self._token_at(i)
+            return None  # a Feature this sheet does not manage
+        return None  # build123d geometry, or something else entirely
+
+    def _index_of(self, ref) -> int:
+        token = self._declared_token(ref, verb="of()")
+        if token is not None:
+            return self._index_of_token(token)
+        if isinstance(ref, Feature):
             raise ValueError("of(): that Feature is not in this sheet's features")
         return self._match_object(ref)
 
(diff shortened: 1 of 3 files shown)


pzfreo:
Independent adversarial review found two findings, both substantiated with a working repro. Both fixed in `b6c1dc4`.

### High — the guard's central hole, and it is #910's own failure mode

`section(feature=…)` returns `Sheet`, not a handle, so my verb enumeration never saw it — yet it retains a feature reference in `_section`. The reviewer restored the positional bug:

```python
self._section = ("feature", self._index_of_token(src))
```

and all 76 tests passed, while the module docstring claimed to cover "section request". The failure it permits is a section plane silently cutting through the wrong hole — a plausible-looking wrong drawing, which is the worst kind here.

This is precisely how #910 happened: `_Control` was missed because it derived from none of the handle classes the migration swept. I then rebuilt the guard on the *same* axis that had already failed once — enumerating the things that look like handles.

**So the axis changed.** The invariant is about STATE, not verbs:

- every field `Sheet.__init__` assigns is now classified as carrying feature references or not; a new field fails until someone classifies it
- `test_every_reference_carrying_state_field_is_exercised` **proves at runtime** that some scenario populates each reference-carrying field, instead of asserting coverage by eye — that is the check that would have caught `_section`
- scenarios added for the three verbs that return `Sheet` but retain a reference: `section`, `datum`, `note`

### Medium — the `DimensionIntent` driver is vacuous

Correct: the driver is `lambda i: i`, so the retained intent is never exercised after the mutation. But it is vacuous *honestly* — the class exposes no verb that consumes its stored reference, because ADR 0012's `.pin()`/`.priority()` were removed from #905 pending the #906 convergence. There is nothing to drive.

Faking a driver would be worse than admitting that. Instead `test_dimension_intent_still_has_no_verbs` fails the moment one is added — which is exactly when a real driver becomes possible. #906 will trip it.

### Verification

Four separate reintroductions, each caught:

| mutation | failures |
|---|---|
| `section` stores an index | 1 |
| `_Control` stores an index (the #910 bug) | 5 |
| `_gdt_src` defers its conversion | 4 |
| `__setitem__` preserves instead of minting | 2 |

169 targeted tests pass; lint 4/4.

pzfreo:
Round two — three findings, all substantiated, all real. Fixed in `31a80aa`.

### High — the section scenario could not observe its own bug

Both holes sat at `Y == 0`, and the section cut materialises as a **Y coordinate**. So a positional implementation picking the wrong hole produced the identical cut, and the scenario passed while the bug was live.

What makes this worth naming: round one's mutation *did* fail one test, and I read that as coverage. It was failing incidentally — not because the oracle saw the retarget. **Counting failures is not the same as checking the failures are the ones you meant.** The guard was vacuous in exactly the way the guard exists to prevent, and a green mutation run did not reveal it.

Holes now sit at different Y. The same mutation fails **4** tests instead of 1.

### Medium — the state-field ratchet walked only `__init__`

A lazily created field was invisible:

```python
def later(self, ref):
    self._new_ref = ref     # not in __init__, so never classified
```

That is precisely the shape a new reference store takes. The walk now covers every method on the class; verified by adding such a field and watching `test_every_sheet_state_field_is_classified` fail.

Two evasions remain, now documented as deliberate acts rather than oversights: `setattr`/`__dict__` writes, and burying a reference inside a field classified reference-free (`_opts`).

### Medium — `add_dimension(-1, …)` resolves where it used to raise

Real widening, and my doing: the old `_feature_index` rejected every negative index, and the consolidated resolver wraps Python-style for all callers.

**Kept, deliberately.** The alternative is a per-caller knob, which recreates the exact divergence this PR removes, and `of(-1)` already behaved this way. But it is now pinned by `TestIntegerRefs` as a decision rather than left as an accident — along with out-of-range on every verb and the bool-is-not-an-index case (`True` is an `int` subclass, and accepting it as index 1 would let a mistyped flag dimension the second feature).

157 targeted tests pass; lint 4/4.

pzfreo:
Round three — two findings, one root cause, both real. Fixed in `42026cd`.

**A mutation that does not move the target asserts nothing, and nothing checked that they moved.**

`sort(key=repr)` left the order unchanged for 9 of 10 scenarios — the features were already in repr order. So nineteen matrix cases looked like coverage and were not. The reviewer proved it rather than arguing it: reintroducing positional tolerance materialization passed the *entire* declaration-then-build test while a plain `reverse` case failed.

Two changes:

- The sorts now key off each feature's **current position**, reversing the list by construction. That still exercises the real `sort(key=…)` and `sort(…, reverse=True)` paths — the claim under test is that `sort` moves whole *entries* and therefore preserves identity, not that any particular ordering is interesting.
- Each mutation **declares** whether it reorders, and the harness **asserts** that at runtime. A scenario whose features resist a mutation now fails loudly instead of passing vacuously.

Effect on the mutation matrix:

| reintroduced bug | before | after |
|---|---|---|
| positional tolerances | *passed the build-time test* | **36 failures** |
| `section` stores an index | 4 | 5 |
| `_Control` stores an index (#910) | 5 | 5 |
| `__setitem__` preserves | 2 | 2 |

### One disagreement

On the second finding — that `double_reverse` restores the original order for all ten scenarios — the fact is right and I don't think the remedy is. A mutation that must **not** change the model is a control, not false confidence: it caught nothing precisely because there was nothing to catch, and a round trip being harmless is a real claim worth pinning. It is now declared `moves=False` so that is explicit, rather than removed.

The thread running through all three rounds is the same one: I kept checking that mutations produced *some* failure rather than checking they produced the failure I meant. Rounds two and three were both that error, in different clothes.

106 targeted tests pass; lint 4/4.

pzfreo:
Round four — one P1, and it exposed a blind spot the whole file shared. Fixed in `7853c53`.

### The resolver was never called on an already-reordered sheet

`_scn_add_dimension` ran the verb during **setup**, with a no-op driver. So the matrix tested *"a recorded intent survives a reorder"* and never *"the resolver returns the right thing when called after one"* — two different claims. Mutating `_feature_token` to return an index passed all 106 tests, and a direct probe showed `add_dimension(ten_mm_handle, "bore")` recording the **6 mm** hole. The verb now runs in the driver, as every other scenario does.

### The sharper half: tokens and indices coincided everywhere

Every fixture built features and removed none, so token N sat at index N. On such a sheet **converting a token to an index is a no-op** — an entire class of bug was unobservable by construction, and no amount of reordering would surface it, because both sides move together.

`TestTokensDivergedFromIndices` declares four features and deletes two, so the survivor's token is 2 while its index is 0. Handles, dimension requests, GD&T provenance, tolerances and the section cut are each checked to bind to the feature they name. It also pins the contrast that makes the rest meaningful: a raw integer ref **is** positional by definition, so `of(0)` naming a different feature after a reorder is correct, not a leak.

| reintroduced confusion | before | after |
|---|---|---|
| `_feature_token` returns an index | 0 | **6** |
| `_declared_token` converts a handle | 0 | **23** |
| `_gdt_ref` returns an index | 0 | **17** |

The AST ratchet proves each resolver *calls* `_declared_token`; it cannot see the result converted back afterwards. These cases are what cover that.

112 targeted tests pass; lint 4/4.

Issue #720, opened 2026-07-18
Remove remaining 0.4.0 Sheet.dimension compatibility surfaces


pzfreo:
ADR 1 (was 0005 §4) requires every compat alias to carry a tracking issue and a removal target ("a facade with no exit date is a failure mode"). This is that issue.

## Part 1 — the seven `Drawing` aliases — **DONE** (#986, 2026-08-01)

Registry-owned: `_named`, `_anno_view`, `_pinned`, `_build_issues`.
Coverage-owned: `_pattern_callouts`, `_patterned_holes`, `_dropped_callout_diams`.

- [x] (#699c) redirect the one production reach-through (`sheet.py` `set(dwg._named)` → `dwg.annotations()`)
- [x] (#699c) migrate the test reach-throughs to the public read surface
- [x] **delete the seven properties**; `registry` / `coverage` are the public handles
- [x] widen the #639 encapsulation guard beyond `annotations/` — done by #699 slice d (`test_no_engine_module_touches_drawing_privates`, `_DWG_PRIVATE_READ_ALLOW` now empty)

`drawing.py`'s own six internal uses went to the registry's surface (`in reg`, `reg.issues`, the new `restore_issues`), **not** to `reg._named`/`reg._build_issues` — pointing them at the privates would have satisfied the deletion while recreating the same reach-through under a different owner.

Also deleted in #986, as they shared the 0.4.0 date: the `sheet_dsl` module, the `generate_script` stub (+ `__all__`), and the bespoke `--style imperative` message.

Guarded by `test_the_expired_compat_aliases_stay_deleted` and `test_the_deleted_modules_and_stubs_stay_deleted` — absence is now asserted, not just documented. (The old `_DRAWING_PRIVATES` roster is a *superset* check, so it passed with the deleted names still listed and would have passed if they came back.)

## Part 2 — the other surfaces dated "0.4.0 with #720" — **open**

These are referenced by #720 in code/ADR comments but were never in this body. Recording them here so the references resolve:

- [ ] **bare dimension-role spellings** (`sheet.py:1556`, ADR 4 (was 0016)) — `dimension(f, "width")` for `"width.length"`. **Measured scale: 1 call site** (`test_sheet_emit.py:119`) across the whole Sheet test surface. A static grep suggests ~78 candidates, but that badly over-counts: discriminated parameters legitimately keep the bare spelling, and most call sites already use canonical ids.
[...]

PR #919, opened 2026-07-28
refactor(#873): the verb that carries a number is measured_dimension

### LINKED ISSUE #873: ADR 0016 PR5: rename the materialized Sheet verb to measured_dimension
Child of #867.

`dimension` means **referential** on both `Sheet` and `Drawing`; today's materialized `Sheet.dimension(kind=…, value=…, ref_pts=…)` becomes `measured_dimension`, with `model.declare.authored_dimension` renamed in step so one vocabulary spans the façade and the model layer.

The shim must be a **transitional overload, not a `@deprecated` wrapper**: because the name is *reused* rather than retired, an old keyword call after the rename would raise `TypeError` rather than warn. `Sheet.dimension` dispatches on how it was called for one release, expiring at 0.4.0 with the ADR 0005 §4 alias removals (#720).

Independent of PR1–PR4; can land in any order relative to them.

**Gate:** generated AP242 scripts regenerate clean; the emitter change is one string.

pzfreo:
## Why

`dimension` means **referential** on both `Sheet` and `Drawing` (ADR 0016) — it names a feature and a role, and the engine reads the value off the geometry. The verb that carries an explicit number needed a name that says so:

| before | after |
|---|---|
| `Sheet.dimension(kind=…, value=…)` | `Sheet.measured_dimension(...)` |
| `model.declare.authored_dimension` | `model.declare.measured_dimension` |

One vocabulary across the façade and the model layer. A measured dimension is the one place a value does *not* come from the part — which is exactly what the name should carry.

## The shim is an overload, not a deprecation wrapper

Because the name is **reused** rather than retired, a plain rename would leave old keyword calls raising `TypeError` from the new signature's argument list rather than saying what happened. So `Sheet.dimension` dispatches on how it was called, warns, and delegates — for one release, expiring at 0.4.0 with the #720 alias removals.

A call it cannot interpret says the referential form is #874, rather than complaining about missing keywords.

The **model-layer** function is a plain rename with no alias: unlike the Sheet verb, nothing takes over the old name there, so an alias would be compat debt with no migration to smooth. Alpha project; the mandate is a clean surface.

## Gate

Generated AP242 scripts must regenerate clean — asserted **both ways**: the new spelling present *and* the old absent, so the emitter can't quietly keep writing a pre-deprecated script.

## Also

Adds the `Unreleased` changelog section this epic had been accruing debt against. #868, #869/#870/#871, #872, #894, #901/#903 and #908/#910/#912 have all merged since v0.3.9 with no entry.

637 targeted tests pass; lint 4/4.

Closes #873.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,54 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- **`sheet.add_dimension(feature, role)` — ask the planner to carry one more
+  measurement** (ADR 0016 / #872). Referential: it names a feature and a role and
+  carries **no number**, so the value still comes from the geometry and a size lives in
+  exactly one place. It changes *selection*, not derivation — a request can never
+  introduce a number the part does not have. Requesting something the planner already
+  emits is a deliberate no-op, so a script can ask without first knowing the rule set's
+  mind. `sheet.auto_dimensions()` states the source explicitly (optional in this
+  release; #874 makes it mandatory).
+- **Addressable dimension identity** (#869/#870/#871): every planned measurement now has
+  a stable `DimensionId(feature, parameter)`, and correlated measurements that must be
+  named as one — a grid pattern's row and column pitch, a step ladder — group into an
+  `AddressableDimension`. This is what a `dimension(...)` line will name, and what
+  suppression and provenance key on.
+
+### Changed
+
+- **`Sheet.dimension(kind=…, value=…)` is now `Sheet.measured_dimension(...)`**
+  (ADR 0016 / #873), and `model.declare.authored_dimension` is
+  `model.declare.measured_dimension` to match. `dimension` becomes the *referential*
+  verb on both `Sheet` and `Drawing` — name a feature and a role, read the value off the
+  geometry — so the verb that carries an explicit number needed a name that says so.
+  The old spelling is a **transitional overload**, not a deprecation wrapper: because the
+  name is reused rather than retired, `Sheet.dimension` dispatches on how it was called
+  and warns, for one release. It expires at 0.4.0 with the #720 alias removals.
+  Generated AP242 scripts emit `measured_dimension` and so arrive un-deprecated.
+- **`Sheet` handles address features by identity, not position** (#908/#910/#912). A
+  handle, tolerance, GD&T origin, section request or dimension intent now follows *its*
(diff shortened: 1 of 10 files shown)


pzfreo:
Review found one issue, and it's the kind that gets worse rather than staler. Fixed in `8acfc39`.

**The diagnostics still said `dimension()`.** Seven validation messages in the constructor, plus the emitter docstring claiming AP242 PMI emits as `dimension(...)`.

Left alone these become actively misleading rather than merely out of date: once `dimension` takes its referential meaning in #874, `dimension() kind must be one of…` implicates the *wrong API* — a caller would go looking at the verb that carries no kind at all. Renamed, and the emitter docstring now also says *why* it writes the new spelling: a regenerated script must not be born deprecated.

The reviewer found no dispatch or round-trip defect, and confirmed by probe that legacy keyword calls warn once and build the identical IR feature, referential forms are rejected pending #874, an emitted `measured_dimension` line executes and reconstructs an identical feature, and the AP242 generated-script path works end to end.

267 targeted tests pass; lint 4/4.

pzfreo:
Second review round — two findings, both correct, fixed in `29a3764`. The reviewer's verdict was **not mergeable as written**, and on the first finding that was the right call.

### The changelog described an API this PR does not ship

My release note said `dimension` *"becomes the referential verb on both `Sheet` and `Drawing`"*. It does not. `Drawing.dimension` is already referential; `Sheet.dimension` is only **reserved** by this PR and raises a `TypeError` naming #874. A user reading that note would write the call and get an exception.

Corrected to say what actually ships — the rename reserves the name, and calling it referentially raises today while naming the issue that delivers it. The `Sheet` docstring made the same overstatement and is corrected the same way.

This is worth owning plainly: I wrote the changelog entry describing the *intent* of the rename rather than its *effect*, which is precisely the failure mode release notes exist to prevent.

### Three stale references

`from_model.py`, `compose.py` and a test comment still said `Sheet.dimension()`. Same reasoning as the last round — once the name is referential these point at the wrong API.

The reviewer confirmed by probe that the implementation itself is sound: legacy keyword calls warn and produce the same feature, referential calls are rejected pending #874, emission produces `sheet.measured_dimension(...)`, and executing that emitted line round-trips the feature exactly with no deprecation warning.

267 targeted tests pass; lint 4/4.

Issue #876, opened 2026-07-27
ADR 0016 PR8: suppression by omission from an authored set


pzfreo:
Child of #867. Depends on PR6 + PR7. **Gated on #867's open question.**

A surfaced referential `dimension` line, commented out, marks that `DimensionId` suppressed in `plan_dimensions` output — marked, not filtered. No separate verb.

**Blocked until #867's open question resolves:** whether this is reachable at `Sheet.build()` time (the set is known before the solve, so plausibly no recompose is needed) or genuinely requires the automatic-candidate reconstruction that ADR 0016 attributes to the now-closed #426/#707.

ADR 0016's scheduling header and its 'Honest reconciliation needs the full recompose' constraint need amending to match whatever this resolves to.

Issue #261, opened 2026-06-29
helper footgun: HoleCallout renders a float value wider than the equivalent _fmt string


pzfreo:
**Found during #238 B1** (PR #259).

`HoleCallout` measures/renders a **numeric** value differently from the **pre-formatted string** of the same value:

```python
HoleCallout(8.0,  count=2, through=True, draft=draft).callout_width  # → 22.143
HoleCallout("8",  count=2, through=True, draft=draft).callout_width  # → 18.546
```

i.e. a float `8.0` renders as `ø8.0` (wider) while the string `"8"` renders as `ø8`. Because layout is width-driven, passing a float shifts placement and can drop callouts that would otherwise fit (this is exactly what happened in B1 when the engine started sourcing values from the IR, which carries clean floats).

**Current state:** worked around in draftwright — `callout_from_spec` (in `annotations/from_model.py`) `_fmt`s every numeric value before constructing the `HoleCallout`, matching the engine. So there's no live bug today; this issue is to remove the footgun.

**Why it matters:** any future code that passes a raw float to `HoleCallout` (depth, cbore_dia, …) silently gets the wrong width. The IR principle is "no baked labels" (values are floats), so the trap is easy to hit.

**Proposed:** `HoleCallout` should format numeric inputs with the same precision/trailing-zero rule as `_fmt` (likely **upstream in build123d-drafting-helpers**), so float and string inputs are equivalent. Until then, keep the draftwright-side `_fmt` and document the requirement.

Refs: `src/draftwright/annotations/from_model.py` `callout_from_spec`.

pzfreo:
Draftwright side **contained + documented** (PR #274): `callout_from_spec` is the single HoleCallout chokepoint and a contract comment now records why values must be `_fmt` strings. The robust fix — `HoleCallout` formatting its own numeric inputs — is **upstream in build123d-drafting-helpers** (separate pip package), so it can't land in a draftwright PR. Leaving this open to track the upstream fix; draftwright has no live bug.

pzfreo:
Draftwright side contained + documented (#274): callout_from_spec is the single HoleCallout chokepoint with a contract comment. The robust fix (HoleCallout formatting its own numeric inputs) is upstream in build123d-drafting-helpers, a separate package — not fixable in a draftwright PR, and draftwright has no live bug. Closing here; the upstream fix is tracked in that repo.

PR #920, opened 2026-07-28
feat(#875): suppression marks a dimension, it does not filter the group

### LINKED ISSUE #875: ADR 0016 PR7: suppression marks, it does not filter
Child of #867. Depends on PR0.

The rule ADR 0016 rests on:

> **No renderer may infer an engineering fact from the presence or absence of a dimension parameter.** Parameters carry values *for display*; facts live on the feature.

Two mechanics:

1. **Suppression marks.** `PlannedDimension` already carries `suppressed`/`reason` and thirteen render sites already honour it. The compound-callout path (`_first`/`hole_callout_spec`) is the one reader that ignores the flag — close that gap, so a suppressed dimension leaves the group's engineering data intact.
2. **The head of a compound callout is a declared dependency.** `hole_callout_spec` returns nothing without the bore ⌀, so suppressing it while spotface or counterbore intents remain is an authoring error and **raises**, naming the dependency. Not lint-and-drop (silently discards authored intent), not implicit restore (makes the script say something it does not say).

**Gate:** a blind hole never prints `THRU` under any suppression combination.

pzfreo:
## The rule

ADR 0016 rests on one:

> **No renderer may infer an engineering fact from the presence or absence of a dimension parameter.** Parameters carry values *for display*; facts live on the feature.

Two mechanics follow.

## 1. Suppression is honoured everywhere

Thirteen render sites already skipped suppressed dimensions. The compound-callout path did not — `_first` walked `group.dims` and returned the first match regardless — so a suppressed counterbore still printed.

It now consults the flag, with `_planned` kept beside it as the deliberate **raw** lookup. Suppression has to be *decided against* somewhere; separating the two makes which is which explicit rather than leaving a caller to wonder whether a given lookup was filtered.

## 2. The head of a compound callout is a declared dependency

`⌀20 THRU ⌴ ⌀32 ↓ 1.5` has no reading with its leading term removed. So suppressing the bore ⌀ while a counterbore, spotface or countersink segment remains **raises**, naming the orphan and saying what to do about it.

Both alternatives lose information the author supplied:

- **lint-and-drop** silently discards authored intent
- **implicit restore** makes the drawing say something the script does not

Suppressing the head *and* its dependents together is coherent and stays silent — the rule is about orphaning, not about the head being sacred.

## The gate

> a blind hole never prints `THRU` under any suppression combination

Tested over the **whole power set** of the hole's parameters rather than a chosen case, with a guard that fails if every combination happened to suppress the callout entirely (otherwise the gate could pass by testing nothing).

## Verification

Mutation-verified both mechanics: `_first` ignoring the flag fails 4 tests; dropping the head-dependency check fails 2.

One harness bug found and fixed **before** review: `_suppress` assigned the flattened `group.dims` to every unit, so a two-unit group came back with four dims. The tests passed anyway — `_first` returns the first match either way — which is exactly the shape of a harness bug that hides what it exists to measure. It now rebuilds each unit from its own members and asserts the dimension count is unchanged.

538 targeted tests pass; lint 4/4.

Closes #875.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@
   emits is a deliberate no-op, so a script can ask without first knowing the rule set's
   mind. `sheet.auto_dimensions()` states the source explicitly (optional in this
   release; #874 makes it mandatory).
+- **Suppressing a dimension marks it; it no longer leaks into the callout** (ADR 0016 /
+  #875). `PlannedDimension.suppressed` was honoured at thirteen render sites but not by
+  the compound hole-callout path, so a suppressed counterbore still printed. The group
+  keeps its engineering data either way — what changes is whether a value reaches the
+  page. Suppressing the bore ⌀ while a counterbore, spotface or countersink segment
+  remains now **raises** and names the orphan: `⌀20 THRU ⌴ ⌀32 ↓ 1.5` has no reading with
+  its leading term removed, and silently dropping the segments would discard authored
+  intent while silently restoring the head would make the drawing say something the
+  script does not.
 - **Addressable dimension identity** (#869/#870/#871): every planned measurement now has
   a stable `DimensionId(feature, parameter)`, and correlated measurements that must be
   named as one — a grid pattern's row and column pitch, a step ladder — group into an
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -76,118 +76,18 @@ from draftwright.annotations._common import (
     strip_obstacles,
 )
 from draftwright.layout import StripCandidate, plan_strip
+
+# Re-exported: `annotations/holes.py` and the tests import the spec from here, and the
+# renderer is its natural home from a caller's point of view even though the reading
+# itself now lives in the IR waist so the page estimator can share it (#875 review).
+# `hole_callout_spec` is re-exported: `annotations/holes.py` and the tests reach it here, and
+# the renderer is its natural home from a caller's point of view — the READING moved down to the
+# IR waist only so the page/scale estimator can share it instead of keeping a second copy that
+# drifts (#875 review). `as` form so the re-export is deliberate, not an unused import.
+from draftwright.model.callout import _first as _first
+from draftwright.model.callout import hole_callout_spec as hole_callout_spec
 from draftwright.model.ir import AUTHORED_DIMENSION_KINDS, HoleFeature, PatternFeature
-from draftwright.model.planner import DimensionGroup, plan_locations
-
-
-def _planned(group: DimensionGroup, kind: str, *roles: str):
-    """First PLANNED dimension matching *kind* and any of *roles*, in role order — suppressed
-    or not. The lookup suppression is decided against; :func:`_first` is what honours it."""
-    for role in roles:
-        for pd in group.dims:
-            if pd.param.kind == kind and pd.param.role == role:
-                return pd
-    return None
-
-
-def _first(group: DimensionGroup, kind: str, *roles: str) -> float | None:
-    """First **unsuppressed** parameter value matching *kind* and any of *roles*, in role order.
-
-    Honouring ``suppressed`` here is ADR 0016 / #875. Thirteen render sites already skipped
-    suppressed dimensions; the compound-callout path did not, so a suppressed segment still
-    printed. Suppression MARKS a dimension rather than removing it (the group keeps its
-    engineering data either way) — what changes is whether it reaches the page."""
-    pd = _planned(group, kind, *roles)
-    return None if pd is None or pd.suppressed else pd.param.value
-
-
-#: Segments of a compound hole callout that cannot be printed without the bore ⌀ that heads it:
-#: `⌀20 THRU ⌴ ⌀32 ↓ 1.5` has no reading with the leading term removed.
-_CALLOUT_HEAD_DEPENDENTS = (
-    ("diameter", "counterbore"),
-    ("depth", "counterbore"),
-    ("diameter", "spotface"),
-    ("depth", "spotface"),
-    ("diameter", "countersink"),
-    ("angle", "countersink"),
-)
-
-
-def _refuse_headless_callout(group: DimensionGroup, bore_pd) -> None:
-    """Raise if the bore ⌀ is suppressed while a segment that depends on it is not (ADR 0016).
-
-    The head of a compound callout is a **declared dependency**. Suppressing it while a
-    counterbore or countersink intent remains is an authoring error, and the alternatives are
-    both worse: lint-and-drop silently discards authored intent, and implicitly restoring the
-    head makes the drawing say something the script does not. So it raises and names the
-    dependency, which is the only option that leaves the author in control."""
-    orphans = [
-        f"{role}.{kind}"
-        for kind, role in _CALLOUT_HEAD_DEPENDENTS
-        if (pd := _planned(group, kind, role)) is not None and not pd.suppressed
-    ]
-    if not orphans:
-        return  # the whole callout is suppressed — coherent, and nothing to print
-    raise ValueError(
-        f"suppressing the bore diameter would leave {', '.join(orphans)} with no callout to "
-        f"head ({bore_pd.reason or 'suppressed'}). A compound callout reads as one string, so "
-        "its leading ⌀ is a dependency: suppress those segments too, or keep the bore ⌀."
-    )
-
-
-def hole_callout_spec(group: DimensionGroup) -> dict | None:
-    """A hole/pattern group's plan → `HoleCallout` kwargs, mirroring the engine's
-    convention. ``None`` if not a hole-bearing callout.
-
-    From the plan: bore from `DimParameter` roles; the cbore/spotface *step* with
-    counterbore precedence (``step = cbore or spotface``, as the engine does);
-    ``count`` and the pattern *suffix* (``EQ SP ON ø50 BC`` / ``(3×3)``) from the
-    source feature.
-
-    ``through`` is read off the FEATURE, never inferred from a missing bore-depth
-    param (#868). `HoleFeature.parameters()` only emits the depth for a blind hole,
-    so absence-as-signal would make any consumer that filters the parameter list —
-    ADR 0016 suppression above all — silently render a blind hole as ``THRU``. The
-    rule that follows (ADR 0016): a renderer may not infer an engineering *fact*
-    from the presence or absence of a dimension parameter — parameters carry values
-    for display, facts live on the feature."""
-    feat = group.feature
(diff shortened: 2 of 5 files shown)


pzfreo:
Three findings, all correct. Fixed in `abbb773`. The first two turned out to be the same defect at different scales.

### High — partial segment suppression silently discarded authored intent

My dependency rule only guarded the bore *head*. Suppressing `counterbore.diameter` while its depth survived produced `cbore_dia=None, cbore_depth=1.5`, and the renderer quietly dropped the whole counterbore — the exact *"lint-and-drop silently discards authored intent"* outcome the rule exists to forbid, left in place one level down.

The rule is now stated where it actually holds: a compound callout's **segments are atomic**. `⌴ ⌀32 ↓ 1.5` needs both terms, and half of it is not a shorter callout but a different and wrong one. Suppressing part of a segment raises; suppressing a whole segment is coherent and silent. The head becomes one case of the same rule rather than a special one — which is the formulation I should have reached first.

### Medium — the page estimator kept a second copy, and it had drifted

`compose._est_planned_bore_callout_width` re-derived bore/depth/cbore/suffix itself, with its own `_first` that ignored `suppressed` and its own `depth is None → THRU` inference — **the exact inference #868 removed from the renderer, still live in the estimator.**

| case | reserved | rendered |
|---|---|---|
| blind ⌀12×8, depth suppressed | 12.8 mm | 6.1 mm |
| suppressed counterbore | 33.7 mm | 14.0 mm |

Fixing it twice leaves two copies to drift again, so the **reading moved down to the IR waist** as `model/callout.py` — rank 0, below both consumers (`compose` 2, `annotations` 4). The renderer turns the spec into a `HoleCallout`; the estimator turns the same spec into token widths. #261's estimator/render agreement is now structural rather than maintained by hand.

### Medium — the raising is unreachable from a user path

Correct, and worth stating rather than defending. The planner suppresses only envelope spans and `add_dimension()` can merely *clear* a suppression, so #875 lands the mechanism that makes suppression safe **before** #876 makes it reachable — which is the right order, but it does mean these tests can only exercise hand-built planner output.

So that is now explicit: `test_suppression_is_still_unreachable_from_a_user_path` asserts it and **fails when #876 lands**, which is exactly when end-to-end cases become possible and should be written.

671 targeted tests pass; lint 4/4.

pzfreo:
One finding, and it was introduced **by this PR**. Fixed in `351565c`.

Teaching `_first` to honour `suppressed` silently changed every caller — including one that was never about display. `render_centermarks` read the bore ⌀ through it to size the mark, so a suppressed ⌀20 collapsed from a **42 mm** centre mark to the **2.5 mm** floor.

A centre mark is furniture derived from the hole's *physical* size. Suppression says "do not print this value", never "this hole is smaller".

So this is the governing rule biting the change that enforces it: **facts live on the feature, parameters carry display values** — and that holds for geometry just as much as for text. I applied it to the callout string and missed that the same `_first` fed a geometric consumer. The size now comes off the feature (its pattern member for a pattern).

Pinned by **two** tests, because the obvious one passes on any constant — including the 2.5 mm floor that was the bug:

- the mark keeps its size under suppression
- the size actually tracks the diameter

Mutation-verified: restoring the parameter read fails both.

347 targeted tests pass; lint 4/4.

pzfreo:
Third round — one High, correct, and it took two attempts to get right. Fixed in `b91335b`.

**The finding:** the bore segment was defined as diameter-only, so suppressing a blind hole's ⌀ while its depth survived returned `None` and silently discarded the depth — the exact half-segment discard the rule exists to forbid, in the one segment that heads the string. And the test I wrote for *"a plain hole may suppress its head"* used a **blind** hole, so it enshrined the bug rather than catching it.

**My first fix made it worse in the other direction.** Making the bore atomic with its depth meant suppressing `bore.depth` alone started raising — and `⌀12` without a depth is a perfectly good callout.

So the relation is **asymmetric**, and the code now says so. Each segment has a **head** term — always its ⌀ — and dependents that only mean something beside it:

| | |
|---|---|
| `⌴ ⌀32` | readable — a counterbore of unstated depth |
| `⌴ ↓ 1.5` | not a counterbore at all |
| `⌀12` | readable |
| `↓ 8` | not a hole |

Suppressing a head orphans its dependents and raises; suppressing a dependent is a legitimate authoring choice. The bore is not exempt from its own rule, and its depth is a *dependent* rather than — as I first had it — no part of the segment at all.

Two rounds were needed because I kept reaching for symmetry: "segments are atomic" is a tidier rule than the true one, and it is wrong in both directions.

**The power-set gate** now tolerates a raise as a legitimate outcome (refusing is not printing `THRU`) while asserting all three outcomes actually occur — so it cannot pass by everything raising or everything falling silent.

Mutation-verified: dropping the segment loop fails 5 tests; ignoring `suppressed` fails 10. 387 targeted tests pass; lint 4/4.

pzfreo:
Fourth round — two High, both correct. Fixed in `98333b0`.

### Suppression was applied *after* the counterbore/spotface precedence

Those roles are a **fallback chain**. Resolving precedence first and then nulling the winner meant a suppressed counterbore swallowed the spotface that should have taken its place — an unsuppressed `spotface=(35, 0.5)` rendered as nothing at all.

`_first` now walks the roles and returns the first **unsuppressed** match, which is what "first" meant all along. I had bolted suppression onto the outside of a lookup whose whole purpose was ordered fallback.

### The head rule only considered dimensional dependents

A thread spec and a pattern's `4× … EQ SP ON ⌀50 BC` ride the same callout string but live on the **feature**, so they survive any amount of parameter suppression — and were then discarded in silence when the head went. Suppressing the bore ⌀ on an `M10x1.5` hole returned `None` and lost the thread.

That is exactly the outcome the head rule exists to prevent, so they are now dependents of the head in the same way a counterbore is, and raise the same way.

### The pattern across four rounds

Both of these were missed because my fixtures were too clean: one counterbore, no spotface, no thread, no pattern. Looking back at all four rounds on this PR, every finding has been a **fixture gap** rather than an error in the rule — the rule kept surviving, the cases I tested it against kept not existing. Worth noting because it says where the next one will be too.

Mutation-verified: restoring precedence-before-suppression fails 1; ignoring the feature-resident dependents fails 1. 480 targeted tests pass; lint 4/4.

pzfreo:
Fifth round — one P1, and it was found by a **10,500-outcome sweep** across the hole feature space rather than by reading the code. That is the right way to have found it, because the code reads fine. Fixed in `acda0ba`.

`cbore_dia` and `cbore_depth` each applied the counterbore→spotface fallback **independently**. So for a feature carrying both, suppressing only `counterbore.depth` produced:

```
cbore=(32.0, 1.5), spotface=(35.0, 0.5),  suppress counterbore.depth
→  cbore_dia=32.0   cbore_depth=0.5
```

⌀32 from the counterbore paired with 0.5 from the spotface — **a recess present on neither feature.** 168 accepted combinations produced it.

This is a *wrong* drawing rather than a missing one, which is the worse failure: a missing dimension is visible to whoever reads the sheet; an invented one is not. And the reviewer's closing observation is the sharpest thing in this whole PR — the estimator and renderer agreed throughout, because both consumed the same erroneous spec. **Agreement between two consumers says nothing about whether what they agree on is true.** I had been treating that agreement as a correctness property; it is only a consistency one.

`_recess` now resolves the chain **once, for the segment**, and takes both terms from the role that won. A role wins on its head being unsuppressed, matching the asymmetric rule, so its depth may still be legitimately suppressed — yielding a ⌀ with no stated depth.

**The fixture is now a sweep**, since all five rounds have been fixture gaps: over the whole suppression power set of both roles, whatever the callout prints must exist *together* on one of the two features. It also asserts the sweep actually exercised each role intact, so it cannot pass by never reaching them.

Mutation-verified: restoring the independent fallbacks fails 2. 396 targeted tests pass; lint 4/4.

pzfreo:
Sixth round — two High, both from a 15,000-outcome sweep. Fixed in `5b56899`.

### A suppressed bolt-circle diameter still printed

`EQ SP ON ø50 BC` was built from `feat.bcd` directly, bypassing the planned `bolt_circle.diameter` and its mark. 384 accepted combinations leaked it — a 38.4 mm callout where 16.3 mm was intended.

This is ADR 0016's rule read in the *other* direction, and I had only internalised one half of it. "Facts live on the feature" told me to read `feat.bcd`; what it does not license is deciding **whether to print** from the feature. The value is a fact; the printing is the parameter's business.

The `(3×3)` grid suffix stays unconditional, and that is not an oversight: it is a count, not a dimension, and has no parameter to suppress.

### A grouped plain hole's count was silently discarded

`4× ⌀6 THRU` is a `HoleFeature` **with a count**, not a `PatternFeature` — and my head-dependency rule only recognised the latter, so suppressing the bore ⌀ dropped the `4×` without a word. The multiplier is a dependent of the head for both kinds.

Mutation-verified: reading `feat.bcd` directly fails 1; restricting the count check to patterns fails 1. 480 targeted tests pass; lint 4/4.

pzfreo:
Seventh round — one merge blocker, a **spurious raise**. Fixed in `cfe8062`.

The counterbore and spotface share one recess slot in the callout string, so while the counterbore's ⌀ prints the spotface is **shadowed** — not rendered before a suppression and not rendered after it. Validating every role regardless meant suppressing `spotface.diameter` on a hole with an intact counterbore raised about `spotface.depth` being orphaned, when the drawing had never carried either term.

That inverts the rule's purpose. It exists to stop authored intent vanishing *silently*; refusing a suppression that changes nothing just blocks a legitimate script. Six rounds of being told I was too permissive made the seventh error being too strict — the rule needed to know about precedence, not more coverage.

`_shadowed` makes the exemption **conditional**, which is what keeps it honest: shadowing depends on the counterbore printing, so once its head is suppressed the spotface is live again and its own half-suppression raises as it should. Both directions are tested.

Mutation-verified: ignoring shadowing fails 1.

pzfreo:
Eighth round — one P1 from a 19,500-outcome sweep (48 too-permissive, **0 too-strict**, so the shadowing fix held). Fixed in `5c2f494`.

The head-dependency rule listed the thread and the `N×` multiplier but **not the pattern suffix**. Invisible while every fixture used `count > 1`, because the multiplier fired the raise and masked the gap. `count=1` is a legal declaration, and there the multiplier is silent — so a one-member bolt circle lost its `EQ SP ON ø50 BC`, and a `1×1` grid its `(1×1)`, without a word.

`_pattern_suffix` is now shared by the spec and the rule, because **the rule has to know exactly what would print.** That is also the structural lesson rather than the local fix: round six taught the suffix about suppression, round eight taught the rule about the suffix, and against a shared derivation both would have been a single change. Two copies of "what does this callout print" is what I actually had.

Mutation-verified: dropping the suffix dependency fails 2. 483 targeted tests pass; lint 4/4.

PR #921, opened 2026-07-29
feat(#874/#876): a build must say where its dimensions come from

### LINKED ISSUE #874: ADR 0016 PR6: a build must request a dimension source (breaking change)

### LINKED ISSUE #876: ADR 0016 PR8: suppression by omission from an authored set

pzfreo:
## The decision

ADR 0016's phase 2. The dimension set gets **exactly two sources**, and the script always says which:

| source | meaning |
|---|---|
| `sheet.auto_dimensions()` | the planner's selected set, optionally augmented by `add_dimension(...)` |
| `sheet.dimension(feature, role)` | the **complete authored set** |

Three errors, all checked at `build()` rather than in the verbs so intent stays order-independent:

1. mixing the two raises
2. `add_dimension` without a source raises (as before)
3. **a build that requested neither raises**

## The breaking change

`Sheet(part).build()` no longer auto-dimensions by default.

This is the project's standing preference for a loud failure over a plausible-looking wrong drawing (#630/#631, the #632 completeness lint) — and it is what makes omission mean anything at all, which is the whole of #876. A sheet that asked for neither would otherwise build silently, with *which* set it got depending on a default the script never mentions.

**Rejected:** implicit-by-usage ("any `dimension` line turns the automatic set off"). A hand-author adding one pitch dimension would silently lose every ⌀ callout.

`Sheet.from_part()` states the automatic source itself. That is not the implicit default returning by the back door: asking for detection **is** asking for the engine's reading of the part, features and dimensions alike. A plain `Sheet(part)` still has to say.

## Suppression by omission (#876) falls out

With #875's marking in place, a parameter outside the authored set is planned `suppressed=True` with a reason and **keeps its value** — marked, not filtered — so the group retains its engineering data and the compound callout's dependency rules still refuse to orphan half a term.

**This settles the epic's open question the cheap way.** ADR 0016 said phases 2–4 sat behind the global recompose. They do not: a `Sheet` script's set is known *before* the solve, so the planner just marks. What genuinely still needs the recompose is the narrower case the ADR conflated with it — `Drawing.finalize()` dropping an *automatic* dimension **after** the solve has run. The ADR now says so.

## Shared addressing

`_resolve_measurement` is shared by both verbs. They address a measurement identically and differ only in what naming one **means** — `add_dimension` leaves the rule set's decision standing for anything unnamed, `dimension` omits it. A second copy of the addressing is exactly how the callout reading drifted in #875.

## Migration

128 test `Sheet` constructions across 18 files, plus `AGENTS.md`, `README.md` and the workflow doc. The emitter writes `sheet.auto_dimensions()` with a comment pointing at the authored alternative, so generated scripts state their source like any other.

The identity guard's state-field ratchet **fired on the new `_authored` field** — which is exactly what #912 built it for, one PR earlier.

## Verification

Full fast tier green (the 9 stragglers were in files my `-k` selection never matched — found by the full run, migrated, verified). Lint 4/4.

Closes #874. Closes #876.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,7 +24,7 @@ dwg = build_drawing("part.step")            # or build_drawing(a_build123d_solid
 dwg.export("out", formats=("pdf", "png"))
 
 # 2. Declared — the Sheet façade (ADR 0011). Statement-style: declare features + aspects on `s`.
-s = Sheet(solid)                            # declare against the build123d part
+s = Sheet(solid).auto_dimensions()          # declare against the part; state the dim source
 h = s.hole(hole_solid)                      # returns a handle you can keep
 h.fit("H7")                                 # aspects: .fit/.tolerance/.note/.thread/.finish/…
 s.slot(slot_solid)
@@ -75,7 +75,7 @@ Use the declared surface — there is deliberately **no** public "add a raw fram
 the controlled **feature first**, then target it (a handle, `Feature`, index, or a face):
 
 ```python
-s = Sheet(solid)
+s = Sheet(solid).auto_dimensions()
 hole = s.hole(hole_solid)                   # the feature the tolerance controls
 s.datum("A", top_face)
 s.control(hole).position(0.1, to="A").perpendicularity(0.05, to="A")  # chain characteristics
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@
   exactly one place. It changes *selection*, not derivation — a request can never
   introduce a number the part does not have. Requesting something the planner already
   emits is a deliberate no-op, so a script can ask without first knowing the rule set's
-  mind. `sheet.auto_dimensions()` states the source explicitly (optional in this
-  release; #874 makes it mandatory).
+  mind. It augments the planner's set, so it requires `sheet.auto_dimensions()` — see
+  the mandatory dimension source below.
 - **A build must say where its dimensions come from — BREAKING** (ADR 0016 /
   #874/#876). The dimension set has exactly two sources and the script always says
   which: `sheet.auto_dimensions()` for the planner's selected set (optionally augmented
@@ -25,7 +25,15 @@
   planned suppressed with a reason and keeps its value — marked, not filtered — so the
   group retains its engineering data and a compound callout still refuses to orphan half
   a term. This is what makes the mandatory source worth having: omission only means
-  something inside a set declared complete.
+  something inside a set declared complete. The overall height, the prismatic step
+  rungs and the step-position shoulders honour it too: those passes rebuilt their marks
+  from the feature and the bounding box, so the plan said "omitted" while the page still
+  carried the dimension. Correlated sets are kept or dropped **whole** — one
+  `dimension(step, "step_height")` line addresses the whole ladder, never a rung.
+  A model that declares no envelope has no parameter naming its overall height, so an
+  authored build no longer draws the bounding-box fallback: declare `.envelope()` to
+  make it nameable. Location dimensions stay outside the authored set until they have
+  addressable identity (#883).
 - **Suppressing a dimension marks it; it no longer leaks into the callout** (ADR 0016 /
   #875). `PlannedDimension.suppressed` was honoured at thirteen render sites but not by
   the compound hole-callout path, so a suppressed counterbore still printed. The group
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,20 @@
   emits is a deliberate no-op, so a script can ask without first knowing the rule set's
   mind. `sheet.auto_dimensions()` states the source explicitly (optional in this
   release; #874 makes it mandatory).
+- **A build must say where its dimensions come from — BREAKING** (ADR 0016 /
+  #874/#876). The dimension set has exactly two sources and the script always says
+  which: `sheet.auto_dimensions()` for the planner's selected set (optionally augmented
+  by `add_dimension`), or `sheet.dimension(feature, role)` declarations for the
+  **complete authored set**. Mixing the two raises, and so does building with neither —
+  `Sheet(part).build()` no longer auto-dimensions by default. `Sheet.from_part()` states
+  the automatic source itself, since asking for detection is asking for the engine's
+  reading of the part. **Migration:** add `.auto_dimensions()` to any `Sheet(...)` you
+  build; generated scripts already emit it.
+- **Suppression by omission** (#876): a measurement left out of an authored set is
+  planned suppressed with a reason and keeps its value — marked, not filtered — so the
+  group retains its engineering data and a compound callout still refuses to orphan half
+  a term. This is what makes the mandatory source worth having: omission only means
+  something inside a set declared complete.
 - **Suppressing a dimension marks it; it no longer leaks into the callout** (ADR 0016 /
   #875). `PlannedDimension.suppressed` was honoured at thirteen render sites but not by
   the compound hole-callout path, so a suppressed counterbore still printed. The group
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,12 @@
   exactly these" back into "draw everything" on re-run; emitting those declarations needs
   nameable feature identity in the generated script (#922). The `--script` path is
   unaffected — it emits from a detected model.
+- **`Sheet.from_part(...)` can be taken over by an authored set**: detect the features,
+  then declare exactly which of their measurements to draw. `from_part` chooses the
+  automatic source on the caller's behalf, so adding `dimension(...)` lines overrides it
+  rather than conflicting; an explicit `auto_dimensions()` still conflicts, because there
+  the script has said both things. A `callout()` naming a feature the authored set omitted
+  is refused in both the live and deferred paths rather than silently drawing nothing.
 - **Suppressing a dimension marks it; it no longer leaks into the callout** (ADR 0016 /
   #875). `PlannedDimension.suppressed` was honoured at thirteen render sites but not by
   the compound hole-callout path, so a suppressed counterbore still printed. The group
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,11 @@
   A model that declares no envelope has no parameter naming its overall height, so an
   authored build no longer draws the bounding-box fallback: declare `.envelope()` to
   make it nameable. Location dimensions stay outside the authored set until they have
-  addressable identity (#883).
+  addressable identity (#883). `emit_sheet_script()` refuses a model carrying an authored
+  set rather than emitting `auto_dimensions()` for it, which would have turned "draw
+  exactly these" back into "draw everything" on re-run; emitting those declarations needs
+  nameable feature identity in the generated script (#922). The `--script` path is
+  unaffected — it emits from a detected model.
 - **Suppressing a dimension marks it; it no longer leaks into the callout** (ADR 0016 /
   #875). `PlannedDimension.suppressed` was honoured at thirteen render sites but not by
   the compound hole-callout path, so a suppressed counterbore still printed. The group
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ from draftwright import Sheet
 plate = Box(120, 80, 20)
 bore = Pos(0, 0, 0) * Cylinder(4, 20)
 
-sheet = Sheet(plate - bore, title="Plate", number="DWG-002")
+sheet = Sheet(plate - bore, title="Plate", number="DWG-002").auto_dimensions()
 sheet.envelope()
 sheet.datum("A", plate.faces().sort_by()[-1])             # datum A on the top face
 hole = sheet.hole(bore).finish("1.6").note("M3x0.5 TAP")  # ⌀8 bore, Ra 1.6, tapped
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -41,6 +41,31 @@ planner's set (#872); dimension identity is `DimensionId(feature, parameter)`
 (#869/#870/#871). Location dimensions remain unaddressable (#883), and per-annotation
 provenance for compound callouts remains #886.
 
+### What the authored set governs
+
+An authored set replaces the planner's rule set for every dimension that has
+**addressable identity** — a parameter on an IR feature. Marking is not enough on its
+own: a renderer that rebuilds its marks from the feature or the bounding box never sees
+the mark, so the plan can say "omitted" while the page still carries the dimension.
+Three auto passes did exactly that until #921; they now read the plan
+(`env_dim_placed` / `set_dim_placed`), and the correlated sets — a `step_height` ladder,
+a `step_position` chain — are kept or dropped **whole**, matching the tier-3 identity
+rule that a single `role=` line addresses the set rather than a member.
+
+Two categories sit outside the set, both deliberately and both visible rather than
+silent:
+
+- **Location dimensions** have no addressable identity yet (`plan_locations` returns a
+  flat, cross-feature, ref-deduped list that never enters a `DimensionGroup`), so an
+  authored set neither names nor suppresses them. Giving them identity is a design
+  question — is a patterned hole's location one unit or one per member? — tracked as
+  **#883**, which also blocks the `"location"` role in the selector.
+- **A bounding-box fallback with no feature behind it.** A model that declares no
+  `EnvelopeFeature` still gets an overall height off the bbox, and no parameter anywhere
+  names it. Rather than let an authored drawing carry a dimension its author had no way
+  to ask for, the fallback is refused under an authored set: declaring `.envelope()` is
+  how the overall height becomes nameable. Automatic builds are unaffected.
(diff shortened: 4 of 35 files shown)


pzfreo:
Review found a **P0**, and it is the worst kind. Fixed in `c8deaa2`.

### The authored set never reached the planner

One `_coerce_model` call forwarded `requested` and not `authored`, so the planner never took the authored branch. **#876's omission semantics did not work through the façade they exist for** — and every unit test passed.

The reason they passed is the lesson: nothing asserted that omission changed a **drawing**, only that the model carried the intent. The gap between those two is precisely where the bug lived. `TestOmissionReachesTheDrawing` now closes it — an authored set survives the build, an omitted measurement is marked suppressed with its reason, it keeps its value, and the automatic set is unaffected as a control.

### Two bypasses of the mandatory source

- **`Sheet.model()` did not apply the gate**, so `build_drawing(part, model=sheet.model())` walked straight around it and the two public surfaces disagreed about one sheet — the #707 class of divergence. It is the model the engine *would* draw; a sheet that cannot be built must not hand one out.
- **`_coerce_model` accepted `requested=` and `authored=` together.** The mutual exclusion is a property of the **model**, not of the façade that usually builds it — `build_drawing(...)` is a public entry point under ADR 0011 and could construct the state `Sheet` refuses.

### The migration erased its own regression test

`test_a_sheet_without_augments_still_builds_without_the_source_verb` asserted `auto_dimensions()` was **optional**. My mechanical edit rewrote its setup to call the verb while leaving the name and docstring intact — so it went on passing while proving the opposite of what it claimed, and **no test asserted that a plain `Sheet(part).build()` raises.** The breaking change's own gate was unguarded.

That is the risk in a 128-site mechanical migration, and I did not check for it: a regex cannot tell a test that *uses* a behaviour from one that *asserts its absence*.

Also discharged #875's `test_suppression_is_still_unreachable_from_a_user_path`, written to fail the moment a real path could suppress a hole parameter. #876 is that path, so it fired as designed and is replaced by its success condition.

Mutation-verified: dropping the authored handoff again fails 1.

pzfreo:
The latest head still has three correctness gaps; one is a P0 against #876's headline behavior.

## P0 — omitted authored dimensions still reach the page through the height ladder

`render_height_ladder()` does not consume the planned suppression state. An envelope authored for only `width` still renders the omitted overall height:

```python
from draftwright import Sheet

part = Box(90, 60, 20) - Pos(0, 0, 12) * Cylinder(6, 20)
sheet = Sheet(part, title="T", number="N")
env = sheet.envelope()
sheet.dimension(env, "width")
dwg = sheet.build()

print([
    (name, getattr(obj, "label", None))
    for name, obj in dwg.iter_annotations()
    if name.startswith("dim_") or name.startswith("m_env")
])
```

Actual:

```text
[('dim_height', '20'), ('m_env_width', '90')]
```

The planner correctly marks envelope `height` and `depth` suppressed, but `render_height_ladder()` independently reconstructs `dim_height` from the part bbox. Step-level dimensions use the same independent path, so authored omission can leak there too.

`TestOmissionReachesTheDrawing` is not actually drawing-observable: it checks that `build().model()` carries `authored_dimensions`, then separately inspects `plan_dimensions(sheet.model())`. It never asserts that omitted annotations are absent from `Drawing`, which is why this survived the earlier P0 fix. Please add a regression that examines `dwg.annotations()`/annotation provenance and proves omitted envelope height and step levels do not render.

## P1 — detected-model `authored=` is silently ignored

`_assemble()` rejects `model is None and requested`, but not `model is None and authored is not None`. It then takes the detected model branch and drops the authored set:

```python
dwg = build_drawing(part, authored=(request,))
assert dwg.model().authored_dimensions is not None  # actual: None
```

An authored request names declared feature identity just like `requested=`, so this should fail closed and require `model=` rather than silently reverting to automatic dimensions.

## P1 — `_coerce_model()` validates arguments, not effective model state

The mutual-exclusion guard checks only `requested` and `authored` passed in this call. It misses dimension-source state inherited from a supplied `PartModel`. Both combinations currently succeed and produce a model carrying both sources:

```python
_coerce_model(
    PartModel(..., requested_dimensions=(request,)),
    part,
    authored=(request,),
)

_coerce_model(
    PartModel(..., authored_dimensions=(request,)),
    part,
    requested=(request,),
)
```

The guard needs to validate the effective post-merge `requested_dimensions` and `authored_dimensions`, and should also reject an already-invalid bare `PartModel` at the public boundary.

## Stale contract text

`Sheet.auto_dimensions()` still says it is optional and that a build without it auto-dimensions; the constructor comment says the same. Both now directly contradict this PR's breaking change and should be updated.

The platform matrix is green, but both Codecov checks currently fail. I would not merge until the P0 rendering bypass and both public-boundary fail-open paths are closed with mutation-resistant tests.

pzfreo:
The core product decision here is right, but I don't think the latest implementation is yet the right architectural shape to merge.

What looks right:

- a drawing must explicitly choose automatic or authored dimensions;
- an authored set is complete, so omission means suppression;
- conflicting sources fail loudly;
- suppressed measurements stay represented in the plan for diagnostics.

The latest commits appear to close the concrete failures from the review rounds. The concern is what those rounds reveal: height ladders, step positions, turned dimensions, callouts, deferred edits, emitted scripts, and `from_part()` each needed separate enforcement. That is evidence of a missing compiler boundary rather than simply missing tests.

At present, the planner marks dimensions as suppressed, but several renderers still receive enough raw model/geometry state to reconstruct those dimensions. Consequently every renderer must remember to consult `pd.suppressed`, correlated-set helpers, envelope helpers, or a separate callout-kind mapping. A new renderer can repeat the same failure.

I recommend making the compiled dimension plan the sole source of dimensional annotations:

```
PartModel
    -> compile dimension intent once
    -> RenderableDimensionPlan + suppression diagnostics
    -> renderers consume only approved entries
```

Concretely:

1. Renderers should not reconstruct dimensions from features or bounding boxes. Correlated dimensions such as ladders should be explicit plan groups.
2. Compound-callout eligibility should be derived from that same plan, rather than a parallel `_CALLOUT_PARAM_KINDS` policy table.
3. Live and deferred edits should pass through the same intent compilation path. `Drawing._refuse_authored_omission()` currently replans and special-cases callouts, creating another policy path.
4. I would simplify `from_part()`: let it detect features, but require callers to choose `.auto_dimensions()` or authored `.dimension(...)`. That removes the subtle `None / implicit / explicit` source state.

So my recommendation is: retain the PR's semantics, but pause the renderer-by-renderer fixes and centralise the invariant before merge. The repeated adversarial findings are valuable—they are showing exactly where ADR 0016 needs a stronger rule: **renderers may emit dimensional content only from the compiled plan**. Green CI would establish regression coverage for the current cases, but would not remove this architectural risk.

pzfreo:
## Plan: centralise the invariant before merging

Agreed, and the diagnosis is better than the one I was working from. I was treating each round's finding as a missing check. The pattern *is* the finding.

I've reverted the in-flight round-8 work (a third variant of the same wrong shape — a hand-written table replaced by outcome-sniffing) and stopped the renderer-by-renderer fixes. The branch sits at `2ddb9f7`.

### Why it kept happening

Every renderer receives the planned `groups` **and** enough raw state to reconstruct what the plan suppressed:

- 5 still take `model` directly (`render_locations`, `render_step_positions`, `render_height_ladder`, `render_pmi`, `render_gdt`);
- 17 take the `Analysis`, whose `a.bb` is the bounding box the height ladder rebuilt the overall height from;
- `PlannedDimension.suppressed` is advisory — a `bool` on a record the renderer is free to ignore, and nothing about the type says it must not.

So "honour suppression" is a convention each renderer opts into. Eight rounds found eight that hadn't. The enforcement surface has grown to four unrelated mechanisms — `pd.suppressed` reads, `env_dim_placed`, `set_dim_placed`, and `_CALLOUT_PARAM_KINDS` — which is the shape of a missing boundary, not of missing tests.

### The rule for ADR 0016

> **Renderers may emit dimensional content only from the compiled plan.**
>
> A renderer receives approved entries. It does not receive, and may not consult, the feature inventory or the bounding box to decide *what* to draw — only *where*.

The corollary is the part worth writing down: **suppression is not a flag renderers check, it is content they never receive.**

### Target shape

```
PartModel
  → compile_dimensions()            # one place: rules, requests, authored sets
      → RenderableDimensionPlan     # APPROVED entries only
      → SuppressionDiagnostics      # what was omitted and why
  → renderers consume approved entries; lint/coverage consume diagnostics
```

The load-bearing detail is that `RenderableDimensionPlan`'s entry type **has no `suppressed` field**. A renderer cannot ignore a flag that isn't there, and a new renderer cannot repeat the failure. `plan_dimensions` survives as the compiler's first stage, not as the renderer-facing surface.

Diagnostics are a real output, not a debug channel: `lint_feature_coverage` needs to know a ⌀ was deliberately omitted rather than missed, and that distinction is exactly what the current `CoverageState` side-channel fudges.

### The four concrete changes

**1. Correlated sets become explicit plan groups.**
`render_height_ladder` and `render_step_positions` rebuild their content from `StepLevelFeature` + `a.bb`. The compiler will emit the rungs, the shoulders and the overall height as entries carrying value and span; the renderers keep only layout — legibility, the leapfrog witness chain, corridor registration. This deletes `set_dim_placed` and, with it, the "no `EnvelopeFeature` ⇒ bounding-box fallback" special case: the compiler either emits an overall-height entry or it doesn't, and a model with no envelope simply has none. `env_dim_placed` goes the same way.

**2. Compound-callout eligibility from the plan.**
The segment and head-dependency rules already live in `model/callout.py`, which is the right layer — they're just consulted too late, after a renderer has already decided to try. The compiler will run them and emit a callout spec or omit the callout. `_CALLOUT_PARAM_KINDS` and `omitted_by_the_authored_set` both go. This is the finding from round 8 (a pocket authored with only `pocket_width` renders nothing, because `render_pockets` needs all three components) and it stops being a policy question the moment eligibility is computed where the components are known.

**3. One intent compilation path for live and deferred edits.**
`Drawing._refuse_authored_omission` re-plans and special-cases callouts — a second policy path, and the one that produced rounds 6 and 7. Instead: an edit compiles into the same plan. If the compiler emits a spec for the feature, the edit draws; if it emits a diagnostic instead, the edit reports that diagnostic. Live and deferred differ in *when* they compile, not in *what decides*.

**4. `from_part()` detects features only.**
Callers choose `.auto_dimensions()` or authored `.dimension(...)`. This deletes the `None`/`"implicit"`/`"explicit"` tri-state, which was my round-6 fix for a real dead end but at the cost of state whose meaning depends on how it was set. Your framing is better: the workflow is still available, it just costs one honest line.

### Honest limits of the boundary

Two things sit outside the compiled plan and the ADR rule must **say so** rather than imply completeness:

- **Location dimensions** (#883). `plan_locations` returns a flat, cross-feature, ref-deduped list that never enters a `DimensionGroup`, so `render_locations` keeps taking `model`. The rule needs an explicit carve-out naming #883, or the first person to read it will assume locations are covered.
- **Raw AP242 PMI** (`PmiFeature.parameters()` returns `[]`). It is an explicit provenance-preserving escape hatch; `render_pmi` renders from the record. Also needs naming.

An unstated exception is how a rule like this rots.

### Sequencing

I recommend **stacking**, for one specific reason: the boundary is independently testable on `main`.

`tests/test_plan_consuming_renderers.py` already proves this — it suppresses every planned measurement and asserts nothing reaches the page, and it works without an authored set. So the boundary can land, and be verified, before #876 exists to motivate it.

- **PR-1 (off `main`)** — introduce `RenderableDimensionPlan` + `compile_dimensions`; migrate all 17 renderers; delete `env_dim_placed` / `set_dim_placed`; move ladders and callout eligibility into the compiler. Oracle: the behavioural guard, plus the existing planner-rule suppressions (square footprint, X/Z-turned, rotational OD) which already exist on `main`.
- **PR-2 (#921 rebased)** — #874/#876 semantics on top. The scattered enforcement this PR accumulated deletes; the regression tests from all eight rounds stay, and become the proof that the boundary covers what they each caught individually.

That ordering also means #921's diff stops being "a breaking change plus eight fixes" and becomes "a breaking change", which is the reviewable thing it should have been.

The alternative — centralising inside #921 — keeps it one story but makes an already-large PR substantially larger, and loses the ability to verify the boundary against `main`'s behaviour before the semantics change on top of it.

### Risk

`from_model.py` is 4,200 lines and holds 17 renderers; this touches all of them. The mitigating factor is that the migration is mechanical per renderer (take approved entries, stop taking `model`) and the behavioural guard fails closed the moment one is missed. I'd rather say plainly that PR-1 is a large mechanical change than discover halfway that it wanted to be three.

I have not started implementing. Say which sequencing you want and I'll begin; if you'd rather see the `RenderableDimensionPlan` type signatures and one migrated renderer as a sketch before committing to the shape, that's a smaller first step and probably the right one.

pzfreo:
Go ahead with the stacked plan, starting with the type signatures and one migrated vertical slice before committing to the full renderer migration.

Please use `render_height_ladder` as that slice because it exercises the hardest boundary: correlated rungs, the overall dimension, geometry-derived spans, and suppression. The compiler should decide **what** is approved; the renderer should retain **where/how** placement decisions. A restricted layout-geometry input would be preferable to handing dimensional renderers the full `PartModel` or `Analysis`.

One change to the proposed limits: please do not leave location dimensions as a permanent carve-out. They are dimensions, so allowing `render_locations` to reconstruct them from the model preserves the same bypass class. They can be sequenced later within PR-1 if needed, but the completed boundary should include them before #921 is rebased. Raw AP242 PMI is a reasonable, explicitly documented non-generated exception.

Suggested sequence:

1. Amend ADR 0016 and introduce the plan/diagnostic type signatures.
2. Migrate the height-ladder vertical slice with behavioural tests.
3. Confirm the boundary keeps compilation policy separate from placement policy.
4. Migrate the remaining dimensional renderers, including locations; avoid migrating furniture/GD&T/PMI merely to satisfy a renderer count.
5. Rebase #921 on that work, delete its distributed suppression mechanisms, and retain the adversarial regressions.

Please proceed.

pzfreo:
Superseded by #925 — same semantics, reworked onto #923's compiled-plan boundary. Closing rather than rebasing, and it is worth being precise about why, because this PR's review history is the reason #923 exists.

## What happened

This PR asked each renderer to honour `PlannedDimension.suppressed`. Eight adversarial rounds each found a different one that did not:

1. the authored set never reached the planner;
2. `render_height_ladder` / `render_step_positions` rebuilt their marks from the feature and `a.bb`;
3. the whole turned family — `render_diameters`, `render_boss_diameters`, `render_step_lengths`, `render_rotational` — selected parameters with no suppression check;
4. an authored entry naming an equal-but-distinct feature copy silently suppressed everything;
5. the emitter turned an authored set back into the automatic one on round trip;
6. `from_part` could not be used with an authored set; a deferred `callout()` discarded the edit silently;
7. live and deferred disagreed; a group-wide predicate masked partial omission.

Fixing them individually produced four parallel enforcement mechanisms — `pd.suppressed` reads, `env_dim_placed`, `set_dim_placed`, `_CALLOUT_PARAM_KINDS` — and no reason to think the ninth renderer would be different. **That is what motivated the boundary in #923**: suppression stopped being a flag renderers check and became content they never receive.

So this PR was not wasted. It is the evidence base, and ADR 0016 Amendment 1 cites it directly.

## What carried over to #925

The semantics, entirely: the mandatory dimension source, the referential `dimension(feature, role)` verb, the authored set, suppression by omission, the emitter refusal, and **all 83 adversarial regressions from the rounds above**.

What did not: 441 lines of `from_model.py` and the per-renderer suppression machinery, which #923 made unnecessary. Verified before deciding — marking every parameter but `bore.diameter` suppressed leaves exactly `['bore.diameter']` in the compiled plan, for all 16 migrated renderers at once.

## Branch

`claude/874-876-dimension-source` is left in place until #925 merges, in case anything here is wanted.

Issue #883, opened 2026-07-27
Location dimensions: per-MEMBER addressable identity (per-feature landed in #925)


pzfreo:
## First AI-edit milestone — #1277

The next useful slice is precise member/axis naming for the location dimension an AI changes, including truthful ownership of shared coincident references. Carry that identity through compilation, generated declarations, placement provenance and comparison with #1006. Preserve existing per-feature omission semantics. A run-local member index is not a promise of durable identity across unrelated recognition runs. The milestone must reject an edit to the wrong member or axis, even when the rendered count is unchanged.

Child of #867. Found by adversarial review on #882. **Substantially resolved by #925 — rescoped below.**

## What #925 changed

Locations are now addressable **per feature**. `planner._LOCATION_ROLE` is the single statement of which kinds have a position, `location_datum()` says where it is measured from (`"datum_xy"`, `"bbox"`, or `None` for kinds the engine draws no position for), and `location_role()` derives the authored vocabulary from that one answer. `sheet.dimension(bore, "location")` resolves, and `compile_dimensions().locations` is the approved set every location renderer reads.

That was not a nicety: before it, a location had no `DimParameter`, so an authored set could neither include nor exclude one and **every location was drawn regardless of what the script declared**. A dimension the author cannot address is a dimension the author cannot omit.

The original framing — that the design questions had to be settled before locations could be addressed at all — turned out to be false for the completeness contract. Omission is well-formed at either granularity: "this feature's position is not in the set" means the same thing whether the unit is the feature or the member. A finer id (`location.member.3`) refines the coarse role later without contradicting it.

## What remains

The genuinely open question is **granularity**, and it is about NAMING rather than about suppression:

- is a grouped/patterned hole's location **one** addressable dimension, or one per member position?
- what is the `ParameterId` for the X and Y components the renderer derives from a single `span`?
- deduped coincident refs mean two features can share one drawn location dim — which feature owns the identity? (#925 made the dedup run over the *approved* refs, so an omitted feature can no longer absorb an approved sibling's position, but it did not decide ownership of a genuinely shared one.)

## No longer blocking

This no longer blocks the `"location"` role in the #872 selector — that role works today at feature granularity.

## Related

`from_model.render_locations` is the one location family still compiled without per-axis values, because it groups refs across features and dedups per axis before it knows which dims exist. Splitting it means moving that grouping into the compiler; tracked with the label-provenance drawdown rather than here, since it is a compiler-shape question rather than an identity one.

[...]

PR #923, opened 2026-07-29
feat(ADR 0016): compiled-plan boundary — renderers draw only what the compiler approved

pzfreo:
**PR-1 of the stack under #921.** This is the vertical slice for shape review, not the finished migration — steps 4–5 of the agreed sequence (remaining dimensional renderers incl. locations, then rebasing #921) land on this branch next.

## The invariant

> **Renderers may emit dimensional content only from the compiled plan.**

A dimensional renderer receives approved entries and decides *where* and *how*. It does not decide *what*, and it is not given the feature inventory or the bounding box it would need to decide otherwise. The corollary is the load-bearing part: **suppression is not a flag renderers check, it is content they never receive.**

## Why a boundary rather than more checks

Everything in ADR 0016 already said the planner decides suppression. It did — and the decision was routinely ignored, because `PlannedDimension.suppressed` was *advisory*. A renderer got the planned groups **and** the `PartModel` **and** the `Analysis`, so honouring the plan was a convention it had to opt into, and rebuilding a suppressed dimension from `feature.levels` or `a.bb` stayed one attribute access away.

Eight adversarial review rounds on #921 found eight renderers that hadn't opted in — the height ladder and step positions rebuilding marks from the feature and the bbox; the whole turned family selecting parameters with no suppression check; the compound callout, the deferred-edit path, the script emitter and `from_part` each needing separate enforcement. Every one was a real omission reaching a real drawing.

Fixing them individually produced **four** mechanisms for saying the same thing — `pd.suppressed` reads, `env_dim_placed`, `set_dim_placed`, and a `_CALLOUT_PARAM_KINDS` policy table — and no reason to think the ninth renderer would be different. That is a missing boundary, not missing tests.

## What landed

**`model/compiled.py`** — `compile_dimensions(model, *, include_overall) -> RenderableDimensionPlan`

- `ApprovedDimension` has **no `suppressed` field**. There is nothing to forget.
- Withheld measurements leave through `.diagnostics` (`Omission`), which lint and coverage consume — ADR 0016's "marked, not filtered" survives, but not on the path a renderer walks.
- Correlated sets arrive as explicit `ApprovedLadder` groups, preserving the tier-3 rule that a ladder is approved or omitted **whole**, never half a staircase.
- `Omission.authored` separates the author's omission from a planner rule's suppression — the distinction three #921 attempts kept blurring.
- Spans travel in **part space**; the renderer projects them.

**`_core.LayoutFrame`** — the restricted layout input: projectors, page-space view rectangles, strips, scale. No `part`, no `bb`, no feature lists. View *edges* are page rectangles rather than projected bbox extents, so anchoring beside a view needs no part geometry — the form that let the ladder read `a.bb.max.Z` while calling itself layout.

**`render_height_ladder` migrated** — `(dwg, plan, frame, *, ctx, detail_view)`, no `model`, no `a`. Chosen as the first slice because it holds every hard case at once: correlated rungs, a geometry-derived overall dimension, spans that must survive projection, and suppression.

| Decision | Before | Now |
|---|---|---|
| Which rungs exist, their values | renderer, from `feature.levels` | compiler |
| Uniform staircase → `4× 10` | renderer | compiler — what the drawing *says* |
| Overall height value | renderer, from `a.bb` | compiler |
| Whether the overall height is drawn | split across planner **and** renderer, neither aware of the other | compiler, once |
| Legibility at this scale | renderer | renderer — depends on the page |
| Leapfrog chain, strips, drops | renderer | renderer |

**Omission is not a drop.** The compiler's omission never arrives and is reported through diagnostics; the placer's drop arrived and didn't fit, and reports `placement_unsatisfiable`. Conflating them is how a deliberate suppression came to look like a layout failure.

**What the compiler may do that renderers may not:** read the bounding box. A model with no `EnvelopeFeature` has no parameter naming its overall height, and the value falls back to the bbox. That fallback is legitimate and now lives in exactly one place.

## Verification

- **Full fast tier: 2009 passed**, zero failures.
- **Output byte-identical** across six fixtures — three-tier staircase, rebate, plate, turned shaft, cylinder, uniform staircase — labels and lint codes included.
- Lint 4/4 clean; DAG guard passing (`model/compiled.py` stays rank 0).
- **12 boundary tests** pinning the property structurally, not the instances: the signature cannot name `model`/`a`; an AST walk of the body proves it never reads `features`/`bb`/`levels`/`base`; `LayoutFrame` is asserted to carry no part geometry.
- A **migration counter** names the four renderers still taking the model (`render_locations`, `render_step_positions`, `render_gdt`, `render_pmi`), so finishing is visible rather than assumed.

## Scope, stated so the exceptions can't be mistaken for completeness

- **Location dimensions are inside this rule** and sequenced later in this branch, not carved out. They are dimensions, and letting `render_locations` rebuild them from the model preserves the same bypass class (#883). The boundary is not complete until they are inside it.
- **Raw AP242 PMI** is the one documented exception: `PmiFeature.parameters()` is empty by design and the record renders verbatim.
- **Furniture** — centrelines, centre marks, section arrows — prints no value and stays outside.

## Two judgement calls worth review

1. **`include_overall` became a compile input.** It is drawing state, so it could have stayed a renderer argument — but then the renderer decides set membership again, which is the thing being removed.
2. **`ApprovedLadder.representative`.** The compiler decides the collapse; the renderer reads the flag only to *name* the mark `dim_step_typ` rather than `dim_step_0`. The thinnest coupling I found, but it is coupling — the alternative is the compiler supplying annotation names, which felt like the compiler doing layout.

### DIFF
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -97,10 +97,12 @@ being a thing the height-ladder renderer did while claiming to be doing layout.
 **The rule is the destination. The migration is at its FIRST SLICE, and the inventory
 below is the honest state of it.**
 
-Three renderers have crossed: `render_height_ladder`, `render_step_positions`, and the
-prismatic detail redraw. Everything else still takes either the legacy `DimensionGroup`
-surface — where `suppressed` remains an advisory boolean a renderer may ignore, which is
-exactly what eight #921 rounds found happening — or the raw model.
+Nine renderers have crossed: `render_height_ladder`, `render_step_positions`,
+`render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`,
+`render_boss_diameters`, and `render_boss_heights`, plus the prismatic detail redraw.
+Everything else still takes either the legacy `DimensionGroup` surface — where
+`suppressed` remains an advisory boolean a renderer may ignore, which is exactly what eight
+#921 rounds found happening — or the raw model.
 
 Two earlier versions of this section understated that, each time because the guard behind
 it measured the wrong thing (#923 reviews). Counting renderers that take `model` reported
@@ -111,8 +113,8 @@ lists, so this inventory cannot drift from the code:
 
 | Contract | Meaning | Renderers |
 |---|---|---|
-| `plan` | approved entries only — inside the rule | `render_height_ladder`, `render_step_positions`, `render_plates` (+ the detail redraw) |
-| `groups` | advisory `suppressed` — **pending** | `render_slots`, `render_centermarks`, `render_diameters`, `render_chamfers`, `render_fillets`, `render_flats`, `render_pockets`, `render_grooves`, `render_boss_diameters`, `render_boss_heights`, `render_envelope`, `render_step_lengths`, `render_rotational`, `render_pocket_patterns`, `render_slot_patterns` |
+| `plan` | approved entries only — inside the rule | `render_height_ladder`, `render_step_positions`, `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`, `render_boss_diameters`, `render_boss_heights` (+ the detail redraw) |
+| `groups` | advisory `suppressed` — **pending** | `render_slots`, `render_centermarks`, `render_diameters`, `render_pockets`, `render_envelope`, `render_step_lengths`, `render_rotational`, `render_pocket_patterns`, `render_slot_patterns` |
 | `model` | raw inventory — **pending**, except PMI | `render_locations` (#883), `render_gdt`, `render_pmi` (permitted) |
 
 Pattern pitch dimensions (`_add_furniture` → `_place_pitch_dim`) are pending too: they are
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -97,10 +97,11 @@ being a thing the height-ladder renderer did while claiming to be doing layout.
 **The rule is the destination. The migration is at its FIRST SLICE, and the inventory
 below is the honest state of it.**
 
-Thirteen renderers have crossed: `render_height_ladder`, `render_step_positions`,
+Fourteen renderers have crossed: `render_height_ladder`, `render_step_positions`,
 `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`,
 `render_boss_diameters`, `render_boss_heights`, `render_envelope`, and `render_pockets`,
-`render_pocket_patterns`, and `render_slot_patterns`, plus the prismatic detail redraw.
+`render_pocket_patterns`, `render_slot_patterns`, and `render_diameters`, plus the prismatic
+detail redraw.
 Everything else still takes either the legacy `DimensionGroup` surface — where
 `suppressed` remains an advisory boolean a renderer may ignore, which is exactly what eight
 #921 rounds found happening — or the raw model.
@@ -114,8 +115,8 @@ lists, so this inventory cannot drift from the code:
 
 | Contract | Meaning | Renderers |
 |---|---|---|
-| `plan` | approved entries only — inside the rule | `render_height_ladder`, `render_step_positions`, `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`, `render_boss_diameters`, `render_boss_heights`, `render_envelope`, `render_pockets`, `render_pocket_patterns`, `render_slot_patterns` (+ the detail redraw) |
-| `groups` | advisory `suppressed` — **pending** | `render_slots`, `render_diameters`, `render_step_lengths`, `render_rotational` |
+| `plan` | approved entries only — inside the rule | `render_height_ladder`, `render_step_positions`, `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`, `render_boss_diameters`, `render_boss_heights`, `render_envelope`, `render_pockets`, `render_pocket_patterns`, `render_slot_patterns`, `render_diameters` (+ the detail redraw) |
+| `groups` | advisory `suppressed` — **pending** | `render_slots`, `render_step_lengths`, `render_rotational` |
 | `model` | raw inventory — **pending**, except PMI | `render_locations` (#883), `render_gdt`, `render_pmi` (permitted) |
 
 Pattern pitch dimensions (`_add_furniture` → `_place_pitch_dim`) are pending too: they are
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -97,11 +97,11 @@ being a thing the height-ladder renderer did while claiming to be doing layout.
 **The rule is the destination. The migration is at its FIRST SLICE, and the inventory
 below is the honest state of it.**
 
-Fourteen renderers have crossed: `render_height_ladder`, `render_step_positions`,
+Fifteen renderers have crossed: `render_height_ladder`, `render_step_positions`,
 `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`,
 `render_boss_diameters`, `render_boss_heights`, `render_envelope`, and `render_pockets`,
-`render_pocket_patterns`, `render_slot_patterns`, and `render_diameters`, plus the prismatic
-detail redraw.
+`render_pocket_patterns`, `render_slot_patterns`, `render_diameters`, and
+`render_rotational`, plus the prismatic detail redraw.
 Everything else still takes either the legacy `DimensionGroup` surface — where
 `suppressed` remains an advisory boolean a renderer may ignore, which is exactly what eight
 #921 rounds found happening — or the raw model.
@@ -115,8 +115,8 @@ lists, so this inventory cannot drift from the code:
 
 | Contract | Meaning | Renderers |
 |---|---|---|
-| `plan` | approved entries only — inside the rule | `render_height_ladder`, `render_step_positions`, `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`, `render_boss_diameters`, `render_boss_heights`, `render_envelope`, `render_pockets`, `render_pocket_patterns`, `render_slot_patterns`, `render_diameters` (+ the detail redraw) |
-| `groups` | advisory `suppressed` — **pending** | `render_slots`, `render_step_lengths`, `render_rotational` |
+| `plan` | approved entries only — inside the rule | `render_height_ladder`, `render_step_positions`, `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`, `render_boss_diameters`, `render_boss_heights`, `render_envelope`, `render_pockets`, `render_pocket_patterns`, `render_slot_patterns`, `render_diameters`, `render_rotational` (+ the detail redraw) |
+| `groups` | advisory `suppressed` — **pending** | `render_slots`, `render_step_lengths` |
 | `model` | raw inventory — **pending**, except PMI | `render_locations` (#883), `render_gdt`, `render_pmi` (permitted) |
 
 Pattern pitch dimensions (`_add_furniture` → `_place_pitch_dim`) are pending too: they are
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -94,14 +94,22 @@ being a thing the height-ladder renderer did while claiming to be doing layout.
 
 ### Scope, stated so the exceptions cannot be mistaken for completeness
 
-**The rule is the destination, and the migration is partial. What is not yet inside it is
-scheduled, not excused — an unstated exception is how a rule like this rots.**
+**The rule is the destination; the migration is partial and its remainder is named.**
+Removing the qualification entirely once the detail path landed was wrong — two paths still
+emit dimensional content of their own, and a rule stated more broadly than it holds is worse
+than one honestly scoped (#923 review round 3). What is not yet inside is listed here, and
+`tests/test_compiled_plan_boundary.py` fails the moment an unlisted mark appears.
 
 - **Location dimensions are inside this rule.** They are dimensions, and letting
   `render_locations` rebuild them from the model preserves the same bypass class.
   `plan_locations` returns a flat, cross-feature, ref-deduped list that never enters a
   `DimensionGroup`, so they are sequenced later in the migration rather than carved out
   (#883). The boundary is not complete until they are inside it.
+- **Pattern pitch dimensions are inside this rule and not yet migrated.** `_add_furniture`
+  derives `n× pitch` from `PatternFeature.members`/`count`/`pitch` and places it directly
+  (`annotations/holes.py`), so it survives an empty compiled plan. It is grouped with
+  furniture in the code but it prints a VALUE, which is exactly what makes something
+  dimensional content under this rule — the grouping is the bug. Sequenced with locations.
 - **Raw AP242 PMI is a documented exception.** `PmiFeature.parameters()` is empty by design
   and the record renders verbatim, so there is no compiled content for it to come from. It
   is an explicit provenance-preserving escape hatch, not an oversight.
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -57,10 +57,20 @@ PartModel
   → compile_dimensions()            # one place: rules, requests, authored sets
       → RenderableDimensionPlan     # APPROVED entries only
       → diagnostics                 # what was omitted, and why
-  → renderers consume approved entries; lint/coverage consume diagnostics
+  → renderers consume approved entries (diagnostics: see below — no consumer yet)
 ```
 
 - `ApprovedDimension` has **no `suppressed` field**. There is nothing to forget.
+- Withheld measurements leave through `diagnostics`. **Nothing consumes them yet** — the
+  consumer is `linting/coverage.py` and it arrives with #921, whose authored sets are the
+  first omissions coverage must tell apart from a measurement simply missed. Until then
+  every omission is a planner rule the lint layer already knows by other means.
+- The feature a dimension came from travels as an **opaque `FeatureRef`**: identity and
+  category, no measurement. Carrying the `Feature` itself would have left the bypass one
+  attribute access away — `.feature.levels` rebuilds exactly what the compiler withheld.
+  The two seams that legitimately need the object (ADR 0010 provenance tagging, escalation
+  grouping) resolve it explicitly; a dimensional renderer doing so is a violation the
+  boundary guard catches.
 - Correlated sets (a step-height ladder, a shoulder chain) arrive as explicit
   `ApprovedLadder` groups, so a renderer never rebuilds one from a feature. This preserves
   the tier-3 identity rule — the set is approved or omitted whole, never half a staircase.
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -94,11 +94,25 @@ being a thing the height-ladder renderer did while claiming to be doing layout.
 
 ### Scope, stated so the exceptions cannot be mistaken for completeness
 
+**The rule is the destination, and the migration is partial. Two paths do not yet obey it,
+and both are scheduled rather than excused — an unstated exception is how a rule like this
+rots.**
+
 - **Location dimensions are inside this rule.** They are dimensions, and letting
   `render_locations` rebuild them from the model preserves the same bypass class.
   `plan_locations` returns a flat, cross-feature, ref-deduped list that never enters a
   `DimensionGroup`, so they are sequenced later in the migration rather than carved out
   (#883). The boundary is not complete until they are inside it.
+- **The prismatic detail redraw is inside this rule, and currently bypasses it.**
+  `_request_prismatic_detail` (`annotations/sections.py`) re-derives the step feature from
+  `dwg.model()` and rebuilds the ladder from `step.levels` against `a.bb.min.Z`, so a rung
+  the compiler withheld can still reach the detail view — demonstrated with an approved
+  ladder holding three of five rungs while the detail drew all five (#923 review). The
+  direct renderer is restricted; its escalation is not. The fix is for the escalation to
+  carry the approved rungs (or a placement-only detail request derived from them) rather
+  than rediscovering levels. Until it lands, this rule holds for the direct render path
+  only, and `tests/test_compiled_plan_boundary.py` pins the gap so closing it is
+  detectable.
 - **Raw AP242 PMI is a documented exception.** `PmiFeature.parameters()` is empty by design
   and the record renders verbatim, so there is no compiled content for it to come from. It
   is an explicit provenance-preserving escape hatch, not an oversight.
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -94,25 +94,14 @@ being a thing the height-ladder renderer did while claiming to be doing layout.
 
 ### Scope, stated so the exceptions cannot be mistaken for completeness
 
-**The rule is the destination, and the migration is partial. Two paths do not yet obey it,
-and both are scheduled rather than excused — an unstated exception is how a rule like this
-rots.**
+**The rule is the destination, and the migration is partial. What is not yet inside it is
+scheduled, not excused — an unstated exception is how a rule like this rots.**
 
 - **Location dimensions are inside this rule.** They are dimensions, and letting
   `render_locations` rebuild them from the model preserves the same bypass class.
   `plan_locations` returns a flat, cross-feature, ref-deduped list that never enters a
   `DimensionGroup`, so they are sequenced later in the migration rather than carved out
   (#883). The boundary is not complete until they are inside it.
-- **The prismatic detail redraw is inside this rule, and currently bypasses it.**
-  `_request_prismatic_detail` (`annotations/sections.py`) re-derives the step feature from
-  `dwg.model()` and rebuilds the ladder from `step.levels` against `a.bb.min.Z`, so a rung
-  the compiler withheld can still reach the detail view — demonstrated with an approved
-  ladder holding three of five rungs while the detail drew all five (#923 review). The
-  direct renderer is restricted; its escalation is not. The fix is for the escalation to
-  carry the approved rungs (or a placement-only detail request derived from them) rather
-  than rediscovering levels. Until it lands, this rule holds for the direct render path
-  only, and `tests/test_compiled_plan_boundary.py` pins the gap so closing it is
-  detectable.
 - **Raw AP242 PMI is a documented exception.** `PmiFeature.parameters()` is empty by design
   and the record renders verbatim, so there is no compiled content for it to come from. It
   is an explicit provenance-preserving escape hatch, not an oversight.
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -111,8 +111,8 @@ lists, so this inventory cannot drift from the code:
 
 | Contract | Meaning | Renderers |
 |---|---|---|
-| `plan` | approved entries only — inside the rule | `render_height_ladder`, `render_step_positions` (+ the detail redraw) |
-| `groups` | advisory `suppressed` — **pending** | `render_slots`, `render_centermarks`, `render_diameters`, `render_chamfers`, `render_fillets`, `render_flats`, `render_pockets`, `render_grooves`, `render_boss_diameters`, `render_boss_heights`, `render_plates`, `render_envelope`, `render_step_lengths`, `render_rotational`, `render_pocket_patterns`, `render_slot_patterns` |
+| `plan` | approved entries only — inside the rule | `render_height_ladder`, `render_step_positions`, `render_plates` (+ the detail redraw) |
+| `groups` | advisory `suppressed` — **pending** | `render_slots`, `render_centermarks`, `render_diameters`, `render_chamfers`, `render_fillets`, `render_flats`, `render_pockets`, `render_grooves`, `render_boss_diameters`, `render_boss_heights`, `render_envelope`, `render_step_lengths`, `render_rotational`, `render_pocket_patterns`, `render_slot_patterns` |
 | `model` | raw inventory — **pending**, except PMI | `render_locations` (#883), `render_gdt`, `render_pmi` (permitted) |
 
 Pattern pitch dimensions (`_add_furniture` → `_place_pitch_dim`) are pending too: they are
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -97,10 +97,10 @@ being a thing the height-ladder renderer did while claiming to be doing layout.
 **The rule is the destination. The migration is at its FIRST SLICE, and the inventory
 below is the honest state of it.**
 
-Ten renderers have crossed: `render_height_ladder`, `render_step_positions`,
+Eleven renderers have crossed: `render_height_ladder`, `render_step_positions`,
 `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`,
-`render_boss_diameters`, `render_boss_heights`, and `render_envelope`, plus the prismatic
-detail redraw.
+`render_boss_diameters`, `render_boss_heights`, `render_envelope`, and `render_pockets`,
+plus the prismatic detail redraw.
 Everything else still takes either the legacy `DimensionGroup` surface — where
 `suppressed` remains an advisory boolean a renderer may ignore, which is exactly what eight
 #921 rounds found happening — or the raw model.
@@ -114,8 +114,8 @@ lists, so this inventory cannot drift from the code:
 
 | Contract | Meaning | Renderers |
 |---|---|---|
-| `plan` | approved entries only — inside the rule | `render_height_ladder`, `render_step_positions`, `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`, `render_boss_diameters`, `render_boss_heights`, `render_envelope` (+ the detail redraw) |
-| `groups` | advisory `suppressed` — **pending** | `render_slots`, `render_diameters`, `render_pockets`, `render_step_lengths`, `render_rotational`, `render_pocket_patterns`, `render_slot_patterns` |
+| `plan` | approved entries only — inside the rule | `render_height_ladder`, `render_step_positions`, `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`, `render_boss_diameters`, `render_boss_heights`, `render_envelope`, `render_pockets` (+ the detail redraw) |
+| `groups` | advisory `suppressed` — **pending** | `render_slots`, `render_diameters`, `render_step_lengths`, `render_rotational`, `render_pocket_patterns`, `render_slot_patterns` |
 | `model` | raw inventory — **pending**, except PMI | `render_locations` (#883), `render_gdt`, `render_pmi` (permitted) |
 
 Pattern pitch dimensions (`_add_furniture` → `_place_pitch_dim`) are pending too: they are
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -97,9 +97,10 @@ being a thing the height-ladder renderer did while claiming to be doing layout.
 **The rule is the destination. The migration is at its FIRST SLICE, and the inventory
 below is the honest state of it.**
 
-Nine renderers have crossed: `render_height_ladder`, `render_step_positions`,
+Ten renderers have crossed: `render_height_ladder`, `render_step_positions`,
 `render_plates`, `render_chamfers`, `render_fillets`, `render_flats`, `render_grooves`,
-`render_boss_diameters`, and `render_boss_heights`, plus the prismatic detail redraw.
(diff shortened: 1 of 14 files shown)


pzfreo:
Review outcome: the direction is good, but the proposed boundary is not yet genuinely load-bearing. I recommend fixing these before expanding the renderer migration.

### P1 — the compiled plan still exposes raw feature content

`ApprovedDimension.feature` and `ApprovedLadder.feature` carry the complete `StepLevelFeature`. A renderer can therefore still read `.levels`, `.shoulders`, or `.base` and reconstruct omitted content. The AST test proves only that the current implementation does not do that; the type boundary does not remove the capability. Please replace this with an opaque provenance/escalation reference or a restricted identifier that does not expose dimensional source data.

### P1 — omission diagnostics are discarded

`compile_dimensions()` returns `diagnostics`, but `_auto_annotate()` immediately passes the plan to `render_height_ladder()` and nothing forwards those omissions to lint or coverage. The PR and ADR say lint/coverage consume diagnostics, but that is not currently implemented. Either connect this output now or narrow the claims and define the concrete owner/lifetime that will consume it in the completed PR-1.

### P1 — the load-bearing behavioural test is tautological

`test_suppressing_the_whole_plan_empties_the_ladder` creates an empty `RenderableDimensionPlan` and asserts that its ladder lookups return `None`; it never renders with that plan. It would still pass if `render_height_ladder()` ignored the plan and rebuilt all dimensions. Please patch `compile_dimensions()` to return an empty plan during a real drawing build, or directly invoke the renderer with the empty plan and assert that no `dim_step*`/`dim_height` annotations are emitted.

### P2 — the planner runs twice

`_auto_annotate()` computes `_groups = plan_dimensions(_model)`, then `compile_dimensions(_model)` calls `plan_dimensions(model)` again. That contradicts the existing plan-once invariant and creates two products that can drift during the migration. Compile once and expose the legacy groups from the compiled result, or accept the already-computed groups as the compiler input until migration completes.

### P2 — compiled identities are thrown away

Every new `ApprovedDimension` is constructed with `id=None`, despite ADR 0016 already defining `DimensionId` as the stable addressable identity. The renderer-facing result should preserve that identity for provenance, edits, diagnostics and later deduplication; otherwise the new boundary immediately creates identity debt.

### P3 — duplicated dead policy remains

`_detect_step_repeat()` remains in `from_model.py` after the repeat-collapse decision moved to `model/compiled.py`. Please delete the dead renderer-side copy so there is actually one policy location.

The height-ladder split itself looks sensible, and the targeted suite passes locally (12 tests). The blockers above matter because the first slice should prove the architecture before the larger migration follows.

pzfreo:
Independent adversarial round run against `7460511`. Two findings, both real; fixed in `ddcd5fa`.

### The worse one is mine — a dimension that said one thing and measured another

The compiler measures rungs from `StepLevelFeature.base` (the IR's own datum, and what the planner always used), while the migrated renderer still anchored every witness at the **view's bottom edge**. With a declared `base=5`, the drawn line spanned 35 mm of geometry and read `15`.

The reviewer caught it as a label difference from `main` (`15` vs `20`). The drawn geometry was the deeper half, and it was a defect I introduced rather than the semantic change it first looked like.

Fixed by projecting **both ends** of the entry's own span. The span is the compiler's statement of what is being measured, so projecting both ends is what keeps line and label the same claim. Detected parts have `base == bbox.min.Z`, so output stays byte-identical to `main` across all six fixtures.

Added the invariant that would have caught it — **a label always measures its own span** — asserted at the source for every approved entry, so it covers every renderer rather than one fixture at a time.

### The boundary has a real hole, now stated rather than implied

`_request_prismatic_detail` re-derives the step from `dwg.model()` and rebuilds the ladder from `step.levels` against `a.bb.min.Z`. An approved ladder of three rungs drew five in the detail view. The direct render path obeys the rule; its escalation does not.

Rather than leave the ADR asserting something false, Amendment 1 now says the rule holds for the **direct render path only**, names the bypass, and states what closing it requires (the escalation carrying approved rungs, or a placement-only detail request derived from them). A test asserts the bypass **still exists**, so it fails the moment someone fixes it — a reminder with a deadline, not an accepted exception.

I did not migrate it in this pass. It belongs with locations in step 4, and attempting a second render path's migration unreviewed at this depth is how the sibling PR's rounds 6 and 7 went wrong — both were my own fixes being wrong.

### Verification

- **Full fast tier: 2015 passed**, zero failures.
- Output **byte-identical to `main`** across all six fixtures.
- Lint 4/4 clean; 18 boundary tests.
- The reviewer mutation-verified the existing tests: replacing compiled labels with `"MUTANT"` failed `test_the_rung_set_arrives_decided`; synthesising `dim_height` from the `LayoutFrame` on an empty plan failed `test_an_empty_plan_draws_no_ladder_in_a_REAL_build`. Non-vacuous for the direct seam — and, as it noted, silent on the escalation bypass, which is why that is now pinned separately.

### Still open for your call

The two judgement calls from the PR description stand unchallenged by this round: `include_overall` as a compile input, and `ApprovedLadder.representative` as the one flag the renderer reads from a content decision.

pzfreo:
Latest-head review (`ddcd5fa`): the previous findings are substantially addressed.

- `FeatureRef` narrows the normal renderer surface and the resolver seams are explicit.
- The empty-plan test now exercises a real build.
- Planning is reused rather than run twice.
- `DimensionId` survives compilation.
- The duplicate repeat policy is removed.
- Diagnostic consumption is now described honestly as work for the #921 rebase.
- The declared-base line/label mismatch found in the subsequent adversarial round is correctly fixed by projecting both ends of the approved span.

The focused boundary suite passes locally: **18 passed**. GitHub’s platform matrix, lint and both Codecov checks are green.

### Remaining P1 — do not merge a boundary with a known dimensional bypass

`_request_prismatic_detail` still reconstructs the ladder from `dwg.model()` / `step.levels`, and the author has demonstrated that an approved three-rung plan can become five dimensions in the detail view. That is precisely the suppression leak this boundary exists to make impossible.

`test_the_detail_redraw_bypass_is_pinned_until_it_is_closed` makes the suite pass only while the defective implementation remains present. A test that asserts source text for a known bug is not a regression guard: it actively requires the bug and does not verify its user-visible behaviour.

Please migrate the prismatic-detail escalation before merging this PR. The escalation should carry the approved rung specifications (or a placement-only request derived from them), and a behavioural test should provide a partially approved plan and assert that the detail view emits exactly those approved dimensions and no others. Then remove the temporary ADR qualification that the rule applies only to the direct path.

### Minor test tightening

`test_a_label_always_measures_its_own_span` claims a general invariant but checks only the Z delta and skips representative ladders. For this vertical slice, the representative span is deliberately one rise and should satisfy the same invariant; there is no need to exempt it. Either make this explicitly a height-ladder Z-span test or define an axis/distance-aware invariant suitable for later renderer families.

Verdict: **the architecture is now sound enough to continue, but the known detail-view leak is a merge blocker for PR-1’s boundary claim.** Fix that path before expanding to the remaining dimensional renderers.

pzfreo:
Review of latest head `b8e1086`: the detail-view migration and shoulder migration are good fixes, but the boundary inventory is still incomplete.

### P1 — locations and pattern pitch are not the only outstanding dimensional paths

The ADR now says the remaining generated dimensional paths are locations and pattern pitch. That is not true yet. Most renderers still receive legacy `DimensionGroup` / `PlannedDimension` objects from `_groups`, where suppression remains an advisory boolean:

- `render_rotational`
- hole/callout rendering
- plates, chamfers, fillets, flats, pockets and pattern callouts
- envelope dimensions
- bosses, diameters, step lengths, slots and grooves

Those are exactly the renderer families that #921 repeatedly found ignoring `pd.suppressed`. They have not crossed the new `RenderableDimensionPlan` boundary merely because their signatures do not contain a parameter named `model`.

`test_the_migration_has_not_silently_stalled` therefore measures the wrong property: it counts `render_*` functions accepting `model`, but a renderer accepting legacy `groups` can still see suppressed entries and ignore them. The test currently reports the migration nearly complete while the majority of dimensional renderers are still on the advisory surface.

Please make the migration guard inspect the actual contract: every generated dimensional renderer must accept approved compiled entry types, not `DimensionGroup` / raw planned groups. Until those renderers migrate, list them as pending or describe this PR explicitly as only the ladder/shoulder boundary slice. Do not claim the remainder is only locations and pitch.

### P2 — compile once and share the result

`_auto_annotate()` now calls `compile_dimensions(_model, groups=_groups)` separately in the height-ladder, step-position and detail stages. The planner runs once, but the compiler runs three times. Create one `_compiled` result after `_groups` and pass that same immutable plan to all migrated consumers. Besides avoiding repeated work, this guarantees the direct ladder, shoulders and detail escalation see exactly the same compiled decision.

### P2 — the empty-plan test still uses annotation-name taxonomy

The new `_FURNITURE` allowlist includes `m_locx`, `m_locy`, `m_env` and `hc_`, all of which can carry measured values or dimensional callouts. The staircase fixture happens not to exercise most of them, so the test remains narrower than its “EVERY dimensional mark” comment. Annotation names are not a reliable semantic type system.

A stronger guard is provenance/contract based: inject an empty compiled plan into fixtures covering every renderer family and assert that no annotation tagged as generated dimensional content appears. Better still, make generated dimensional annotations impossible to construct without an approved-entry identity.

The focused boundary tests pass, and the latest detail/shoulder behaviour looks correct. This is still a scope/contract blocker rather than a local rendering bug.

pzfreo:
## Handover

**PR head: `9b8cc1b`** (verified). **Unfinished work: branch `claude/adr-0016-wip-renderer-batch` at `b7d738b`** — pushed but broken, do not merge.

---

### The rule this PR exists to enforce

> **Renderers may emit dimensional content only from the compiled plan.**

Corollary, and the thing to hold onto: **suppression is not a flag renderers check, it is content they never receive.**

It exists because eight adversarial rounds on #921 each found a *different* renderer ignoring `PlannedDimension.suppressed` and rebuilding a suppressed dimension from the `PartModel` or `Analysis.bb`. Fixing them individually produced four parallel enforcement mechanisms. The boundary makes the failure unrepresentable instead.

### What is done and verified on `9b8cc1b`

| | |
|---|---|
| Migrated | `render_height_ladder`, `render_step_positions`, `render_plates`, + the prismatic detail redraw |
| Full fast tier | 2016 passed at `a6684b4`; `9b8cc1b` has lint/mypy/parity + 118 affected tests but **no full tier** |
| Output vs `main` | byte-identical, six fixtures |
| Boundary tests | 19, in `tests/test_compiled_plan_boundary.py` |

**Key types** (`src/draftwright/model/compiled.py`):

- `compile_dimensions(model, *, include_overall, groups) -> RenderableDimensionPlan`
- `ApprovedDimension` — **no `suppressed` field**; carries `kind`/`role`/`discriminator`/`value`/`label`/`span`/`tolerance`
- `ApprovedGroup` — mirrors `DimensionGroup`'s shape (`dims`/`view`/`anchor`/`feature_kind`) so a renderer's loop barely changes
- `FeatureRef` — opaque provenance handle; identity and category only
- `FeatureFacts` + `_FACTS` — the per-kind allowlist of **non-dimensional** structure; reading anything else raises with the rule in the message
- `Omission` / `.diagnostics` — what was withheld and why. **Nothing consumes this yet**; owner is `linting/coverage.py` at the #921 rebase

**`LayoutFrame`** (`_core.py`) — the restricted layout input: projectors, page-space view rectangles, strips, scale. No `part`, no `bb`, no feature lists.

### The two rules that decide the remaining work

1. **Span vs facts.** *Positions that define what a dimension measures travel in its `span`; facts carry only what a span cannot express.* A plate's `lo`/`hi` are the two ends of its thickness → span. Its `axis` is a fact, because no span says which way a slab is thin.
2. **`FeatureRef` equality is STRUCTURAL** — the finalize drain rebuilds feature objects, and `only=` selection previously compared features with `in`. Same reasoning ADR 0016 gives for `DimensionId`, and deliberately unlike `_request_for`, which matches by `is` so two identical holes stay distinct targets.

### What remains

**16 renderers still on the advisory `groups` surface, 3 on raw `model`.** `tests/test_compiled_plan_boundary.py::test_the_migration_guard_measures_the_CONTRACT_not_the_signature` pins all three lists by name and fails the moment one moves — update it and the ADR table together.

1. **Finish the six-renderer batch** (`b7d738b`). 4 failures in `tests/test_script_detail_parity.py`, all in the drain's per-feature `only=` subsets — check the structural-equality change end to end rather than patching until green. One mypy error: `drawing.py` passes `object` to `FeatureRef`.
2. **The other 10 renderers**, mechanical modulo each `_FACTS` entry.
3. **Pattern pitch** (`_add_furniture` → `_place_pitch_dim`). Grouped with furniture in the code but it prints a value — the grouping is the bug.
4. **Locations (#883).** Contains a genuine design question the planner's own docstring flags. My proposal, unratified: *a location is one unit per addressable feature, not per member; deduped coincident refs belong to the feature whose datum the ref derived from, with the loser recorded as an omission.*
5. **Construction-time enforcement** — make a generated dimensional annotation impossible to build without an approved-entry identity. This is what retires the name-prefix allowlist in the empty-plan test, which is explicitly a stopgap.

### Two findings recorded but deliberately not fixed

- **The chamfer angle is printed but never planned.** `ChamferFeature.parameters()` emits only the leg, yet the label reads `3 × 30°`, so the angle cannot be suppressed or toleranced. A real IR gap. Closing it adds a parameter to every chamfer's plan and changes output, which does not belong in a migration claiming byte-identity.
- **Three IR parameters have no span** — `BossFeature` diameter, `GrooveFeature` width/diameter, `ChamferFeature` leg. Under rule 1 they should. Has consumers (`env_dim_placed` treats `span is None` as a signal, plus the emitter and PMI), so it wants its own verified pass.

### How to verify anything here

- Byte-identity: the six-fixture comparison against `main` — annotation names, labels and lint codes. Non-negotiable for a migration; a diff means content moved, not just plumbing.
- **The empty-plan test is the load-bearing one.** Patch `compile_dimensions` inside a real build and assert nothing reaches the page. Its first version asserted a dataclass lookup and never rendered; it would have passed with the renderer ignoring the plan entirely.
- Run the **full tier** before pushing. It caught nine identity-suite failures and a privates-ratchet failure that targeted runs missed, and a broken batch that lint and mypy both passed.

### Review history worth reading

Rounds 1–3 on this PR are in the comments above: the raw `Feature` leaking through the plan, a tautological behavioural test, the planner running twice, discarded `DimensionId`s, a declared `base` making line and label disagree, the detail escalation drawing five rungs from a three-rung plan, and the migration guard measuring signatures instead of contracts. Each is fixed; the pattern is that **the guard was wrong more often than the code**.

pzfreo:
Ready for another review at `f3d3d43`.

I took over from the handover commit and completed the recoverable compiled-plan migration batches:

- repaired and verified the six-renderer WIP batch;
- migrated envelope, pocket, pocket/slot pattern, turned diameter, rotational OD/bore, and turned step-length rendering;
- removed raw pattern-member measurement access and kept only structural member facts;
- resolved opaque `FeatureRef`s only at provenance-registration seams;
- added behavioural suppression coverage for rotational diameters;
- tightened the empty-plan guard by removing exemptions for every migrated renderer;
- corrected the ADR inventory/status so it does not claim completion.

Adversarial review found and fixed two cross-cutting problems during the work: direct leader placement was registering `FeatureRef` instead of the source feature, and the turned-detail crop still depended on a raw step diameter. The latter now derives its crop from page/view geometry.

Local validation on the final code:

- boundary + parity + tolerance suite: **92 passed**;
- affected step/diameter/rotational/pattern suites passed during each batch;
- repo-wide Ruff, Ruff format, mypy, codespell, and `git diff --check`: clean.

The remaining exceptions are stated explicitly rather than hidden behind a misleading contract:

- `render_slots` is the final legacy `DimensionGroup` renderer because its position dimension has no addressable parameter/span yet;
- general locations remain the #883 identity/design question;
- pattern pitch remains coupled to `_add_furniture`;
- raw AP242 PMI is the documented permanent exception; GD&T is still listed as raw/pending.

I deliberately did not make the guard green by passing those raw measurements under renamed arguments. GitHub CI has been triggered for this head and is currently queued.

pzfreo:
## Adversarial review of `9b8cc1b`

Reviewed my own work rather than described it. One confirmed defect, two weaknesses, one thing that survived attack.

### P1 — `render_plates` renames its dimensions

The migration changed the sort key's **shape**, not just its source:

```python
# before:  floats along the thin axis
sorted(plate_groups, key=lambda g: (g.feature.axis, g.feature.lo, g.feature.hi))
# after:   3-tuples, which compare x, then y, then the thin coordinate
sorted(plate_groups, key=lambda gp: (gp[0].facts.axis, gp[1].span[0], gp[1].span[1]))
```

A plate span is `(u, v, lo) → (u, v, hi)`, so the new key sorts by **u first** where the old one sorted by **lo**. Whenever those orderings disagree, the `dim_plate_{axis}{i}` indices swap between plates.

Reproduction (two Z-plates, the thicker one at the lower `u`):

```python
s = Sheet(Box(200, 60, 40), title="T", number="N")
s.add(PlateFeature(frame=Frame((0,0,0),'z'), axis='z', lo=5.0, hi=9.0, u=-60.0, v=0.0))  # 4 thick
s.add(PlateFeature(frame=Frame((0,0,0),'z'), axis='z', lo=1.0, hi=4.0, u= 60.0, v=0.0))  # 3 thick
```

| | |
|---|---|
| `main` | `[('dim_plate_z0', '3'), ('dim_plate_z1', '4')]` |
| `9b8cc1b` | `[('dim_plate_z0', '4'), ('dim_plate_z1', '3')]` |

Annotation names are addressable identity — `drop()`, `pin()` and `annotations_of()` all key on them — so this silently repoints a user's edit at the neighbouring plate. That is the same class of defect as #910, arriving through a sort key.

**It also falsifies this PR's byte-identity claim.** My six-fixture comparison never had two same-axis plates, so it could not see this. Fix: sort by the coordinate along the thin axis, `span[0]["xyz".index(axis)]`, which is what the old key meant.

### P2 — `ApprovedDimension.value` is annotated `float` but can be `None`

`_compile_groups` guards the label (`_fmt(...) if value is not None else ""`) and then passes `value=pd.param.value` straight through. Nothing in the fixtures produces it, so it is latent rather than live — but the type says a renderer may divide by it.

### P2 — `_FACTS["pmi"]` allows `label`

`label` is the text PMI prints. It is defensible, because raw AP242 PMI is the documented non-generated exception and the record renders verbatim — but it is the one entry in the allowlist that is a printed value, and the table's own docstring says such things must travel as approved dimensions. Either the exception is stated at the entry or the entry does not belong.

### Survived attack

- **`FeatureFacts` refuses measurements.** Audited all 19 kinds against a suspect-name list; only the `pmi.label` case above appears. A new kind arrives with no facts at all.
- **`ApprovedGroup.dim()` returning the first match** is not a masking bug: every current caller queries a `(role, kind)` pair that is unique within its feature.

### On the guards, again

The pattern across every round on this PR holds here too: **the code change was sound and the guard was too narrow.** Byte-identity was my strongest check, and it missed a name swap because no fixture had two plates of the same axis. A parity check is only as good as its fixture set, and mine was chosen before the renderer was migrated — i.e. before I knew what could move.

Worth adding a multi-instance case per feature family to the parity set, since ordering defects only appear with two or more of a kind.

pzfreo:
Addressed the latest adversarial findings in `0f00ab1`.

- **P1 fixed:** `render_plates` once again orders same-axis plates by `(axis, lo, hi)` along the thin axis, not by whole span points. Added the exact adversarial two-Z-plate regression; it asserts `dim_plate_z0 == 3` and `dim_plate_z1 == 4`, so lateral movement cannot silently retarget a pin/drop.
- **Value invariant tightened:** `DimParameter.value` is a required `float`, so compilation now formats it directly and explicitly normalises it to `float` rather than implying a nullable renderer state.
- **PMI exception clarified:** `_FACTS["pmi"].label` now documents why it is the sole printed value in the structural allowlist: raw AP242 PMI is the ADR's verbatim, non-generated exception.

Validation on this head:

- exact plate reproduction + boundary selection: **31 passed**;
- boundary + script/detail parity + tolerance suite: **92 passed**;
- repo-wide Ruff, format, mypy, codespell, and diff check: clean.

A fresh GitHub CI matrix is now running for `0f00ab1`.

pzfreo:
## Adversarial review of `9b8cc1b`, second pass

Attacked the areas the first pass did not reach — the shoulder migration and the detail redraw. Two more confirmed defects, both mine, both in code the first pass declared fine.

### P1 — `_axis_of` misclassifies a shoulder that sits on its own datum

`render_step_positions` infers a shoulder's axis from which coordinate its span varies along:

```python
def _axis_of(rung):
    lo, hi = rung.span
    return "x" if lo[0] != hi[0] else "y"
```

I justified dropping an explicit `axis` field with *"the span already says it, so an axis field would be a second statement of one fact."* **That is false when the span is degenerate.** A shoulder at position == datum has `lo == hi` in every coordinate, so an X shoulder is silently classified as Y and routed to the side view instead of above the plan.

Reproduction:

```python
step = StepLevelFeature(Frame((0,0,0),'z'), base=0.0, levels=(10.0,),
                        shoulders=(('x', -45.0),), datum=(-45.0,-30.0,0.0))
```

```text
label= 0  value= 0.0  span lo==hi -> True
_axis_of says: y   (actual axis is x)
```

Reachable for a rebate flush with the part's min edge. The dimension is degenerate either way (value 0), so this is a wrong-view routing rather than a wrong number — but the *inference* is unsound, and the design argument for removing the field was wrong. Either carry the axis explicitly, or have the compiler refuse a zero-length shoulder as an omission.

### P1 — the detail crop moved for a declared base

`_request_prismatic_detail` previously passed `crop_lo=a.bb.min.Z`; it now passes `crop_lo=datum_z`, taken from the approved span's lower end.

```text
bbox min Z = -15.0
datum_z (new crop_lo) = 5.0        # StepLevelFeature(base=5.0)
DIVERGES
```

The `DetailRequest` docstring is explicit that this argument exists to keep the part's base datum inside the projected crop *"so every witness point belongs to visible detail geometry"*. Moving it to the declared base can crop away the geometry the witnesses attach to. Detected models are unaffected (`base == bbox.min.Z`), which is exactly why the six-fixture parity check is silent on it — the same blind spot that hid the plate rename.

The ladder rungs measuring from `base` is right; the *crop* is view geometry and should stay at the bounding box.

### Cleared this pass

- **The float-keyed `label_of` map** in the detail redraw is safe: `_legible_steps` returns members of its input list verbatim, and levels are unique, so no collision or `KeyError`.
- **Shoulder ordering** is preserved — the compiler builds rungs from `sorted(step.shoulders)`, which is the order the renderer used to iterate.

### The pattern, stated plainly

Three passes, three defects, and **all three are the same mistake**: I replaced a value read from the feature with one derived from the span, and each time the derivation was correct for detected models and wrong for declared ones. The plate sort key, the shoulder axis, the detail crop. Detected models have `base == bbox.min.Z` and non-degenerate spans, so the parity fixtures cannot distinguish any of them.

The fix is not more care — it is a parity fixture set built from **declared** models with awkward values: a non-bbox `base`, a zero-length shoulder, two same-axis plates. Every defect found on this PR would have failed against that set on the first run.

pzfreo:
Addressed the second adversarial pass in `28d8578`.

- **Degenerate shoulder axis:** confirmed. `ApprovedDimension` now carries explicit structural `axis` metadata for step-position rungs. `render_step_positions` no longer guesses direction from a varying span coordinate, and the detail X-station selection uses the same compiler-owned axis. Added the exact zero-length X-shoulder regression (`span[0] == span[1]`, axis remains `x`).
- **Declared-base detail crop:** confirmed. Measurement witnesses still use the approved span's declared datum, but `DetailRequest.crop_lo` is restored to `a.bb.min.Z` because the crop describes visible part geometry. Added a regression with bbox base `-15` and declared base `5`.

Validation:

- exact new regressions: **3 passed**;
- broader boundary/shoulder/detail/step-position selection: **72 passed**;
- repo-wide Ruff, format, mypy, codespell, and diff check: clean.

Fresh CI is running for `28d8578`.

pzfreo:
## Adversarial review of `9b8cc1b`, third pass

Attacked the compiler's contract and my own guard lists this time. One confirmed hole, two latent fragilities, two things cleared with measurements.

### P1 — the empty-plan guard has a hole for the renderer I just migrated

`_PENDING_VALUE_CARRYING` in `tests/test_compiled_plan_boundary.py` still lists `m_plate`:

```text
pending prefixes: [..., 'm_locy', 'm_plate', 'm_pocket', ...]
m_plate present -> True   (render_plates is MIGRATED)
```

The empty-plan test asserts that nothing outside `_VALUE_FREE + _PENDING_VALUE_CARRYING` survives a plan with no approved entries. Because `m_plate` is still on the pending list, a plate leaking from an empty plan would be **tolerated**. Not a live leak — `render_plates` does take the plan — but the guard no longer covers the one renderer this PR migrated to prove the mechanism.

The list must shrink as each renderer crosses. It didn't, because nothing ties it to the contract test that *does* track migration. Two lists, one fact, already drifted.

### P2 — `compile_dimensions(model, groups=…)` fails open on a mismatched pair (latent)

`_suppressed_dims` keys its map on `id(group.feature)`, and `_compile_step_ladders` / `_compile_overall_height` look up `id(step)` / `id(env)` taken from `model.features`. Pass `groups` derived from a *different but equal* `PartModel` and every ladder lookup misses silently — suppression is lost rather than raising.

I could not make it produce a wrong drawing today, and said so rather than dressing it up: the planner only suppresses **envelope** parameters, and `_compile_overall_height` re-derives the two rules it cares about (`orientation == "z"`, rotational OD) directly from the model. So the `marked` map is currently redundant for ladders. It stops being redundant the moment #921's authored sets land, which is exactly when a silent fail-open becomes a wrong drawing.

Either key on the feature structurally, or have `compile_dimensions` reject `groups` whose features are not the model's.

### P2 — `LayoutFrame.edges()` names an ordering it does not guarantee

`edges()` returns `(left, right, bottom, top)` built as `front_x(bb.min.X), front_x(bb.max.X), …`. `sections.py` warns explicitly that *"projection handedness can put world max-X on the page-left"*. Every current use is behaviour-identical to what it replaced, so nothing is broken — but the field names assert an ordering the projector does not promise, and the next migration will reasonably write `max(left, right)` or assume `top > bottom`. A trap I built into the type.

### Cleared, with numbers

- **Compile cost**: 0.11 ms vs 0.08 ms for a plan on a 9-hole part — 1.49× a plan, and the orchestrator now compiles once. Not a #602 concern.
- **`groups=` reuse**: approved *group* dims are filtered from the passed groups directly, so the mismatch above cannot corrupt them; only the ladder path is exposed.

### Three passes, and the guards keep losing

| Pass | Defect | Guard that should have caught it |
|---|---|---|
| 1 | plate sort key renames dimensions | byte-identity — no two-plate fixture |
| 2 | shoulder axis on a degenerate span | byte-identity — no declared `base` |
| 2 | detail crop moved for a declared base | byte-identity — no declared `base` |
| 3 | empty-plan guard tolerates `m_plate` | nothing — two lists, one fact |

Every defect on this PR has been found by *reasoning about the change*, never by the suite. The suite has been confirming that detected models still render identically, which they do, and which is not the property at risk.

The two concrete remedies, in order: a declared-model parity fixture set (non-bbox `base`, degenerate shoulder, two same-axis plates), and deriving `_PENDING_VALUE_CARRYING` from the contract test's own pending lists instead of maintaining it by hand.

pzfreo:
Third-pass review dispositioned in `ab12837`.

- **Empty-plan `m_plate` claim:** stale on current head. `_PENDING_VALUE_CARRYING` already contains only `m_locx`, `m_locy`, `hc_`, `dim_pitch`, `m_slot`, and the documented `pmi_` exception. `m_plate` (and every other migrated prefix) was removed before this review comment.
- **Mismatched `model` / `groups=`:** valid latent fail-open. `compile_dimensions` now materialises the supplied plan and rejects any group whose feature is not an identical member of that exact `PartModel`. This protects future authored suppression rather than relying on today's redundant ladder rules. Regression uses equal-but-distinct detected models and requires `ValueError`.
- **`LayoutFrame.edges()` ordering:** valid contract trap. It now normalises projected endpoints and guarantees the documented `(left, right, bottom, top)` order under reversed projection. Added a reversed-endpoint regression.

Validation:

- focused contract regressions: **2 passed**;
- boundary + parity + tolerance suite: **95 passed**;
- repo-wide Ruff, format, mypy, codespell, and diff check: clean.

Fresh CI is running for `ab12837`.

pzfreo:
## Adversarial review of `9b8cc1b`, fourth pass

Attacked the new types under composition — copy/deepcopy/pickle, coverage of the IR, id uniqueness. One genuine bug, two design traps.

### P1 — `FeatureFacts` cannot be deepcopied or pickled

```text
copy:     ok
deepcopy: RecursionError: maximum recursion depth exceeded
pickle:   RecursionError: maximum recursion depth exceeded
```

The classic `__getattr__` + `__slots__` trap, and I walked straight into it:

```python
def __getattr__(self, name: str):
    return self._values[name]        # ← reads its own backing slot
```

`deepcopy` and `pickle` construct the object **without** `__init__`, so `_values` is unset; `__getattr__` then reads `self._values`, which re-enters `__getattr__`, and so on. Confirmed directly:

```text
FeatureFacts.__new__(FeatureFacts)._values
→ root cause: __getattr__ recurses on its own backing slot
```

Anything holding an `ApprovedGroup` inherits it, so a plan cannot cross a process boundary or be snapshotted. Nothing in the engine deepcopies a plan today, so it is latent — but `Drawing`'s transaction/rollback tests do snapshot build state, and this is the sort of thing that surfaces when someone caches or parallelises. One-line fix: refuse underscore names before touching the slot.

`FeatureRef` is unaffected — no `__getattr__`.

### P2 — four IR kinds have no `_FACTS` entry

```text
IR kinds: 23        missing: ['control_frame', 'datum_ref', 'finish', 'note']
```

The behaviour is fail-closed (no entry → no facts → every read raises), which is the right default and what I claimed for a *new* kind. But these are **existing** ADR 0011 aspect kinds, not hypothetical ones. When `render_gdt` migrates it will hit an empty facts record and the failure will read as "this kind exposes no fact `datum`" rather than "nobody has classified this kind yet". The table says 19 kinds; the IR has 23.

### P2 — `ApprovedDimension.label` is not the rendered label

```text
hole   bore.diameter   label='12'
```

The drawing prints `⌀12`. The label field carries `_fmt(value)` and nothing else — the ⌀, the `n×` multiplier, `THRU`, the tolerance suffix and every compound-callout term are added by renderers. That is fine as long as everyone knows, and the ladder renderers I migrated *do* rely on `label` being complete (`4× 10` is baked in there by the compiler). So the field means "the finished label" for ladders and "just the number" for groups.

Two meanings, one name, on the type whose entire purpose is to stop renderers deriving content themselves. Either name them apart or make the compiler own the whole label everywhere.

### Noted, and I believe correct

**All rungs of a ladder share one `DimensionId`** — 2 rungs, 1 distinct id. That is ADR 0016 tier-3 by design (a correlated set is *one* addressable dimension), but it means `rung.id` is not a per-rung key, and anything that dedups approved entries by id would collapse a staircase to one rung. Worth a comment at the field, since `id` on a per-item record reads like a per-item identity.

### Four passes

| Pass | Confirmed |
|---|---|
| 1 | plate sort key renames dimensions |
| 2 | shoulder axis wrong on a degenerate span; detail crop moved for a declared base |
| 3 | empty-plan guard tolerates `m_plate` |
| 4 | `FeatureFacts` unpicklable/undeepcopyable |

Passes 1–3 were all "correct for detected models, wrong for declared". This one is a different class — a Python-level trap in a type I wrote to be strict — which suggests the declared-model fixture set I keep recommending would **not** have caught it, and the standard composition matrix (copy/deepcopy/pickle, the one #921's mandate specified) would have.

Both are cheap. Neither exists yet.

pzfreo:
## Correction: three of my four reviews were against a stale commit

I pinned a review worktree to `9b8cc1b` and never re-fetched, so my second, third and fourth passes were reviewing a head that had already moved. The PR is at `ab12837` and you had pushed 10+ commits in between. Anyone reading those three comments would take them as findings against current code, and mostly they were not. My fault, and easily avoided — a `git fetch` between passes.

Re-run against the actual head `ab12837`. Here is what actually survives.

### Fixed — my findings, correctly dispositioned

- **Plate annotation identity** (pass 1). The divergent two-plate probe now matches `main` exactly: `[('dim_plate_z0', '3'), ('dim_plate_z1', '4')]`. `_plate_order` is the right shape.
- **Degenerate shoulder axis** and **detail crop** (pass 2) — addressed in `28d8578`.
- **Empty-plan `m_plate` hole** (pass 3) — `m_plate` is gone from the pending list.

### Still live at `ab12837`

**P1 — `FeatureFacts` cannot be deepcopied or pickled.**

```text
deepcopy: RecursionError
pickle:   RecursionError
```

Unchanged from pass 4, and this one is not a stale finding. `__getattr__` reads `self._values`, which is in `__slots__`; `deepcopy`/`pickle` construct without `__init__`, so the slot is unset and `__getattr__` re-enters itself. Reproduce with `FeatureFacts.__new__(FeatureFacts)._values`. Fix: refuse underscore names before touching the slot.

Every `ApprovedGroup` inherits it, so a plan cannot be snapshotted or cross a process boundary.

**P2 — four IR kinds still have no `_FACTS` entry.**

```text
missing: ['control_frame', 'datum_ref', 'finish', 'note']
```

Fail-closed, so nothing is wrong today — but these are existing ADR 0011 aspect kinds, and `render_gdt` will meet an empty facts record when it migrates, reporting "exposes no fact `datum`" rather than "this kind was never classified".

**P2 — `ApprovedDimension.label` still means two things.** `bore.diameter → label='12'` while the drawing prints `⌀12`; the ladder path meanwhile relies on `label` being the finished text (`4× 10`). One field, two contracts, on the type whose purpose is to stop renderers deriving content.

### On the process rather than the code

Four passes found four defects and every one came from reasoning about the change — none from the suite. Three of the four were the same failure mode (correct for detected models, wrong for declared), and the fourth was a Python-level trap that a declared-model fixture set would *not* have caught but a standard copy/deepcopy/pickle matrix would.

So there are two gaps in the verification, not one, and both are cheap:

1. a declared-model parity set — non-bbox `base`, degenerate shoulder, two same-axis plates;
2. the composition matrix #921's mandate already specified — copy/deepcopy/pickle over the new types.

Given you have migrated most of the renderer families since, the leverage is now in those two rather than in another renderer.

pzfreo:
Addressed the corrected current-head review in `c62b391`.

- **FeatureFacts composition P1:** fixed the `__slots__` / `__getattr__` recursion by refusing backing/special names before reading `_values` and using `object.__getattribute__`. Added copy, deepcopy and pickle round-trips for both `FeatureFacts` and complete `RenderableDimensionPlan` objects.
- **IR-kind classification P2:** all 23 current IR kinds are now explicit in `_FACTS`. `control_frame`, `datum_ref`, `finish`, and `note` are reviewed-empty entries, distinct from an unknown future kind. A reflection-based test requires exact coverage so this cannot drift again.
- **Label contract P2:** split the ambiguous field into `value_text` (formatted numeric fragment used by ordinary group renderers) and `rendered_label` / `final_label` (complete compiler-owned wording for correlated ladders such as `4× 10`). Ordinary dimensions now raise if a renderer asks for a complete label they do not carry. Added a contract test covering both paths.
- The shared ladder `DimensionId` remains unchanged and continues to represent the correlated set by design.

Validation:

- focused boundary/composition tests: **26 passed**;
- broad affected renderer/boundary selection: **180 passed**;
- repo-wide Ruff, format, mypy, codespell, and diff check: clean.

Fresh CI is running for `c62b391`.

pzfreo:
## Review of `c62b391` — the actual head this time

Re-fetched before reviewing. All four of my earlier findings are dispositioned, and I could not find a new defect. Reporting that as a clean pass rather than manufacturing one.

### My findings, verified fixed at head

| Finding | State |
|---|---|
| Plate sort key renamed dimensions | fixed — divergent two-plate probe matches `main` |
| Degenerate shoulder axis; detail crop moved | fixed |
| Empty-plan guard tolerated `m_plate` | fixed — off the pending list |
| `FeatureFacts` deepcopy/pickle `RecursionError` | fixed — `copy`, `deepcopy`, `pickle` all pass, on the group **and** the whole plan |
| Four IR kinds missing from `_FACTS` | fixed |
| `ApprovedDimension.label` meant two things | resolved by removing the field, which is the better answer than naming them apart |

### What I ran

**Byte-identity vs `main`: identical** across the six-fixture set.

**The declared-model probes I kept recommending** — the cases my original parity set could never reach:

```text
two_plates:          [('dim_height','40'), ('dim_plate_z0','3'), ('dim_plate_z1','4')]   == main
declared_base:       [('dim_height','30')]                                                == main
degenerate_shoulder: RAISED ValueError: Start and end points of border must be different. == main
```

The crash on a zero-length shoulder is **pre-existing** — `main` does exactly the same — so it is not a regression from this work. It is still a real bug (a degenerate declared shoulder should be refused by the compiler as an omission, not blow up in the border builder), and worth its own issue rather than being folded in here.

**Empty-plan leak sweep across six feature families** — patch `compile_dimensions` to return an empty plan inside a real build, list everything that still reaches the page:

```text
prismatic+holes: ['hc_plan0', 'm_locx0', 'm_locy0']
pattern:         ['dim_pitch_plan0', 'hc_plan0', 'm_locx0', 'm_locy0']
turned:          clean
rotational:      clean
staircase:       clean
pocket:          clean
```

Every leak is on the declared pending list (`hc_`, `m_locx`, `m_locy`, `dim_pitch`), and nothing leaks that is claimed migrated. **The inventory matches reality** — which is the thing that was wrong three times running.

**Contract state**: 16 renderers on `plan`, 1 on `groups` (`render_slots`), 3 on `model` (`render_locations`, `render_gdt`, `render_pmi`). Boundary suite: 26 passed.

### What is left, from the measurements rather than the prose

1. `render_slots` — the last legacy `DimensionGroup` renderer.
2. Hole callouts (`hc_`) and pattern pitch (`dim_pitch`) — both still emit from an empty plan.
3. Locations (#883) — still the design question, still blocking the `"location"` role.
4. `render_gdt`; `render_pmi` stays the documented exception.
5. Construction-time enforcement, which is what finally retires `_PENDING_VALUE_CARRYING`.

### One note on my own reviews

Four passes found four defects and none came from the suite. That is now less true: the empty-plan sweep above **is** a suite-shaped check, it covers six families, and it agreed with the declared inventory. The remaining gap I would still close is the composition matrix — copy/deepcopy/pickle over the new types — since that is what caught the one defect a declared-model fixture set would have missed.

Issue #714, opened 2026-07-18
Dedup the mirrored per-axis render blocks in from_model.py (render_locations X/Y, diameter row/column, render_rotational z/x/y)


pzfreo:
Follow-up to #700 (the helper-family consolidation PR folded the small mirrors — 5× anno-bbox, 2× segment/AABB, 2× `_fmt`, the compose/render table sizing, and the boss-diameter ray-exit re-inline — but deliberately left the large per-axis blocks; they are layout-affecting refactors that deserve their own review):

- **`render_locations`** — the X and Y passes are ~70-line near-mirrors (axis-swapped strip/corridor assembly).
- **`_diameter_row_below` vs `_diameter_column_left`** — the same pass mirrored per side, with a flip-cascade only one side has (divergence already begun).
- **`render_rotational`** — the z/x/y furniture branches are triplicated axis-swaps.

The consolidation shape (per ADR 1 (was 0008) — orientation is *data in the IR*, not code branches): one axis-parameterised body per pass, with the per-axis differences expressed as data (axis letter, view name, strip selector, coordinate picker). Each dedup should be its own commit/PR with layout-signature tests run before/after, since these blocks place real geometry.

From the 2026-07-18 architecture audit (#700's "same class" list).

PR #925, opened 2026-07-29
feat(#874/#876): a build must say where its dimensions come from

### LINKED ISSUE #874: ADR 0016 PR6: a build must request a dimension source (breaking change)

### LINKED ISSUE #876: ADR 0016 PR8: suppression by omission from an authored set

pzfreo:
Supersedes #921. Same semantics — reworked onto #923's compiled-plan boundary rather than rebased, because most of #921's content became dead code when that landed.

Closes #874, closes #876.

## The change

A build must say where its dimensions come from. `Sheet(part).build()` now raises; you say `.auto_dimensions()` for the planner's set, or declare the complete set with `dimension(feature, role)` lines. Omission from an authored set means suppression — which is only meaningful once a script says its set is complete.

**Migration:** add `.auto_dimensions()` to any `Sheet(...)` you build. Generated scripts already emit it.

## Why a rework and not a rebase

#921 was 13 commits ahead with main 21 ahead of it, and it accumulated eight adversarial review rounds — each finding a *different* renderer that ignored `PlannedDimension.suppressed` and rebuilt a withheld dimension from the model or the bounding box. That was the whole shape of the PR.

#923 made those checks unnecessary. `compile_dimensions` emits approved entries only, so marking a dimension suppressed removes it from all 16 migrated renderers at once. Measured directly on main before deciding:

```python
# mark everything except bore.diameter suppressed, then compile
approved after authored-style marking: ['bore.diameter']
```

So the rework carries the semantics and drops the machinery.

**Carried over whole** — six files #923 never touched: `sheet.py` (mandatory source, the referential `dimension()` verb, the authored set), `builder.py` (`authored=` threading, the effective-model guard), `planner.py` (`_authored_for`, `_check_authored_targets`), `ir.py`, `sheet_emit.py`, `callout.py`.

**Dropped as dead** — 441 lines of `from_model.py`: `set_dim_placed`, `env_dim_placed`, the turned family's `pd.suppressed` checks, the ladder and step-position gates.

## The eight rounds, now fixed by construction

| Round's finding | #921 needed | Here |
|---|---|---|
| height ladder + step positions rebuilt marks from geometry | renderer fixes | `['m_env_width']` — no rungs, no `dim_height` |
| whole turned family ignored suppression | renderer fixes | `['m_steplen0']` — no diameters |

Neither needed a line of renderer code. The regression tests from every round are retained and pass — 83 of them.

## Two things still needed a home, and both went to the compiler

- **The bbox overall-height fallback.** A model with no `EnvelopeFeature` falls back to the bounding box, and an authored set must not acquire a dimension its author had no way to name. One branch in `_compile_overall_height`, not a rule every renderer remembers.
- **`Drawing.callout` refusal.** Now asks the compiled plan whether anything was approved, replacing `_CALLOUT_PARAM_KINDS`. Three hand-written versions of that table were each wrong for some feature kind (#921 rounds 6–8); the boundary removes the need to predict at all.

## Verification

- **Full fast tier: 2069 passed**, zero failures. Lint 4/4, mypy clean.
- 83 of #921's adversarial regressions retained and passing, including every round's canary.
- The 42-site `auto_dimensions()` migration was re-applied to *current* test files, with a pre-flight check for tests asserting the old default (five flagged, all five safe on inspection) and multi-line `Sheet(...)` calls reported rather than rewritten — the failure mode that erased a regression test in #921.

## One test superseded, not deleted

`test_the_overload_refuses_a_call_it_cannot_interpret` asserted that `dimension(feature, role)` was *not yet* referential. #874 makes that false by design. #921 deleted it; here it is replaced with a test of what still matters — the transitional overload routing the measured form with its deprecation notice.


pzfreo:
## Review — two P1 blockers

The compiled boundary is the right foundation, but the PR introduces **complete authored-set** semantics before all value-carrying paths are governed by that set.

### P1 — authored omission still leaks through pending renderers

A linear hole pattern authored with only `bore.diameter` still emits:

```text
hc_plan0
2× 20       # dim_pitch_plan0 — omitted, but still drawn
30          # m_locx0 — not authored
30          # m_locy0 — not authored
```

Minimal model-level probe:

```python
part = Box(100, 60, 20)
member = HoleFeature(Frame((0, 0, 10), "z"), 4, depth=None, through=True)
pattern = PatternFeature(
    Frame((0, 0, 10), "z"), "linear", 3, member,
    members=((-20, 0, 10), (0, 0, 10), (20, 0, 10)),
    pitch=20, direction=(1, 0, 0),
)
dwg = build_drawing(
    part,
    model=[pattern],
    authored=(RequestedDimension(pattern, "bore.diameter"),),
)
```

`_add_furniture()` still reads `feat.pitch` directly and `plan_locations()` remains outside the compiled/authored selection. The opposite set, authoring only `pitch`, does not build at all: automatic callout assembly raises because the suppressed bore is treated as orphaning the pattern's `3×` multiplier.

So the headline contract — *omission from the authored set means suppression* — is not true yet. Locations are especially fundamental: they are dimensions, survive the authored set, and cannot currently be named with `Sheet.dimension()` because #883 is unresolved.

### P1 — `Drawing.callout()` asks whether anything survived, not whether the callout survived

`_refuse_authored_omission()` permits a callout whenever the feature has **any** approved compiled dimension:

```python
(g for g in approved.groups if g.ref == FeatureRef(feature) and g.dims)
```

That is not the same as the dimensions needed by this callout. Reproduction: author only a step's `step.length`, omitting its diameter, then call `dwg.callout(step)`. The live path creates `m_dia_x0` anyway because `add_feature_diameter()` reads the legacy planned diameter without checking `suppressed`. The deferred path behaves differently.

This directly contradicts the PR's claim that live and deferred callouts now refuse identically.

### Recommendation

Do not merge this yet. Either:

1. finish the hole-callout, pattern-pitch, location, and slot-position boundary migrations before enabling complete authored sets; or
2. narrow the authored API/claim so it cannot represent completeness over dimensional paths it does not control.

Adding more refusal heuristics would recreate the per-kind prediction tables #923 was intended to remove. The structured fix is to make every value-carrying generated path consume approved compiled content, then let an empty approved result mean exactly one thing.

pzfreo:
## Both P1 blockers fixed — and the recommendation's option 1, taken in full

Reproduced both before touching anything. The review was right on both counts, and the second was worse than described: `_refuse_authored_omission` asked *"does this feature have any approved dimension?"*, so a turned step with its **length** authored answered yes and the live path then drew the omitted **diameter**. That is the same defect class as rounds 3 and 7, reintroduced by me while believing the boundary had made it impossible.

Taking the review's structural fix rather than a fifth refusal heuristic: **every value-carrying generated path now consumes approved compiled content**, so an empty approved result means exactly one thing.

### #883 was not the blocker it looked like

The reason locations leaked was deeper than "unmigrated". A location has no `DimParameter` — it is synthesized from the feature and the datum — so `dimension(hole, "location")` **raised**. An authored set could neither name nor exclude one. **A dimension the author cannot address is a dimension the author cannot omit.**

`planner._LOCATION_ROLE` is now the one statement of which kinds have a position, and `location_role()` derives both `plan_locations` and the authored vocabulary from it. The authored role is the coarse `"location"` — one unit per feature.

**#883 asks a question about *naming*** (is a patterned hole's position one addressable thing or one per member?). Omission is well-formed at either granularity, and a finer id (`location.member.3`) refines the coarse role later without contradicting it. So the completeness contract did not have to wait on it. #883 stays open; the ADR now says why it is not a blocker, so the next reader does not re-derive the one that was removed.

One consequence worth naming: the coincident-ref **dedup now runs over the survivors**. Deduping first let an omitted feature's ref absorb an approved sibling's — the author names a measurement and gets nothing, which is the blank-drawing failure `_check_authored_targets` exists to prevent arriving by another route. Verified directly rather than assumed: with two coincident holes and only the second authored, the surviving planned location belongs to the second.

### What each migration turned up

| Path | Was | Now |
|---|---|---|
| `render_locations` | raw `plan_locations` — drew every ref | `plan.locations` |
| `render_slots` | legacy `groups` | `plan.of_kind(...)`; the slot object supplies **witness geometry only** |
| slot position | `s.lo - a.bb.min`, ungated | `_compile_slot_positions`, same `location_role` table |
| pattern pitch/grid | `feat.pitch` | approved group entry |
| `add_feature_diameter` | planned ⌀, read past `suppressed` | approved ⌀ |
| `add_feature_location` | raw planner list | compiled locations |

Two things the review had not seen, both found by reasoning about the change rather than by the suite:

- **Authoring only `pitch` didn't merely crash.** Once unblocked it drew *nothing*, because furniture was a side effect of placing a bore callout. That is the silent-blank failure again. `_furnish_uncalled_patterns` closes it — additive, so every drawing with a placed callout is untouched.
- **`_refuse_headless_callout` made a pattern the one feature whose callout could not be omitted at all**, because the `n×` multiplier counted as an orphan of the bore ⌀. It now turns on *who decided*: #920's refusal stands for planner suppression (the engine must not silently discard a thread spec), but an author omitting the bore is declining the string, not orphaning its prefix. That is what `Omission.authored` is for.

### Deleting the refusal restored a defect, and that is the interesting part

The two failing tests (`..._is_refused_in_both_paths`, `..._refused_identically_live_and_deferred`) pinned the mechanism I removed. What they actually guard is **silence** — round 6's defect was an edit vanishing with no annotation, no pending intent and no warning. Removing the prediction without replacing the *report* would have restored exactly that.

So the report moved to where "drew nothing" is **observed**: `_report_authored_omissions` runs after the callout/diameter drain stages, before their intents drop. Fixing it also caught a second regression I would otherwise have shipped — `add_feature_callout` still raised *"this HoleFeature exposes none"*, which is false (it exposes a bore ⌀ and was told not to draw it). That misleading message was one of the two things the deleted refusal existed to replace.

### Scope, stated honestly

- The `groups` contract list is **empty**, and the guard fails if anything reappears.
- `render_gdt` / `render_pmi` keep the model **permanently**, and the ADR now says why instead of listing them as pending: a control frame's tolerance and a PMI label are author-supplied text rendered verbatim, with no parameter to plan or suppress. An authored set does not govern them because they were never the engine's choice.
- **Hole callouts (`hc_`) remain on the legacy `DimensionGroup` surface.** They honour `suppressed` at every term, so this is structural, not behavioural — but "the renderer checks" is precisely the guarantee this boundary replaces, so it stays named in `_PENDING_VALUE_CARRYING` and in the ADR rather than assumed safe. A test now cross-checks the two lists so they cannot drift apart.
- `_compile_groups` had never emitted diagnostics — `plan.diagnostics` was empty for every ordinary feature, so a caller asking "was this the author's doing?" got *no* for an authored omission. A channel right for three kinds and silent for sixteen is worse than none: it reads as an answer.

**2078 passed** (full fast tier), all four lint checks clean. Auto-path output unchanged across hole/slot/pocket/pattern fixtures.

The meta-finding from earlier rounds holds again: every defect in this round was found by reasoning about the change, none by the suite. The new tests are behavioural — real builds, real annotations — and the dedup-ordering one was checked against the wrong ordering rather than merely asserted.

pzfreo:
Re-review of `b41db0e`: the two previous blockers are fixed, and the latest CI is green. I found one remaining blocker in the authored-set contract.

**P1 — an explicitly authored bolt-circle diameter is silently lost.**

`_refuse_headless_callout()` returns at `model/callout.py:185` whenever *any* authored omission exists. That is correct for pitch-only patterns, but too broad for a bolt circle: `bolt_circle.diameter` is itself an approved dimensional term, and its only rendering is the `EQ SP ON ø… BC` suffix on the bore callout. If the author names the BCD but omits `bore.diameter`, the early return suppresses the dependency error; `hole_callout_spec()` then returns `None`, and `_furnish_uncalled_patterns()` deliberately skips bolt circles. The result contains neither the requested 50 mm BCD nor an `authored_omission` diagnostic.

Minimal reproduction:

```python
import math
from build123d import Box
from draftwright.model.ir import Frame, HoleFeature, PatternFeature, RequestedDimension

hole = HoleFeature(Frame((0, 0, 10), "z"), 4.0, depth=None, through=True)
members = tuple(
    (25 * math.cos(i * math.pi / 2), 25 * math.sin(i * math.pi / 2), 10)
    for i in range(4)
)
pattern = PatternFeature(
    Frame((0, 0, 10), "z"), "bolt_circle", 4, hole,
    members=members, bcd=50.0,
)
dwg = build_drawing(
    Box(100, 60, 20), model=[pattern],
    authored=(RequestedDimension(pattern, "bolt_circle.diameter"),),
)

assert any("50" in str(getattr(a, "label", "")) for _, a in dwg.iter_annotations())
# Fails: there are no pattern annotations and lint is silent.
```

This needs the same explicit compound-dependency treatment as recess terms: either reject BCD-without-bore as an orphaned callout term, or provide a standalone approved BCD renderer. The authored-omission escape should only waive non-dimensional riders (count/thread/static suffix), not an approved `DimParameter` that would otherwise be printed in the suffix. Please add the authored BCD-only regression alongside the new pitch-only test.

Apart from this edge case, the prior fixes now behave correctly in my reproductions: bore-only, pitch-only, and location-only linear patterns are isolated; pitch-only no longer crashes; and the omitted live step diameter stays omitted.

pzfreo:
## P1 fixed — the BCD is a dependent, not a rider (`dd608e4`)

Reproduced your case exactly: no annotation, no lint. The diagnosis was right, and it exposed a distinction I had collapsed.

A compound callout's trailing content is **two** things:

- **Dimensional dependents** — a counterbore ⌀, a countersink angle, and `bolt_circle.diameter`. Each is a planned, addressable `DimParameter` the drawing prints.
- **Non-dimensional riders** — the `n×` multiplier, a thread spec, a grid's `(3×3)`. These live on the FEATURE with no parameter to suppress and have no existence outside the string.

My authored escape was written for riders and waived both. The BCD renders *only* as the suffix, which is precisely what made it look like a rider — but **where a term sits in the string is not what classifies it; having a parameter is.** So the BCD is now evaluated before the escape and raises like any other orphaned dependent.

Taking your first option rather than a standalone BCD renderer: the latter is new output with no callout to attach to, and that is a design decision this PR should not make unilaterally.

### The first attempt was wrong, and your own #920 tests caught it

Adding the BCD to the dependents list made it **short-circuit before the riders**, so a bolt circle whose bore was suppressed by a *planner rule* named the BCD and silently dropped the `4×` multiplier from the message — accurate but incomplete, when one edit fixes both. `test_suppressing_the_head_of_a_pattern_raises` and `test_a_one_member_pattern_suffix_is_a_dependent_too[bolt_circle]` failed on exactly that.

Restructured so the split is stated once and both halves accumulate into one raise:

```
RULE-SET suppression → "…would leave the pattern suffix 'EQ SP ON ø50 BC',
                        the 4× multiplier with no callout to head."
AUTHORED BCD-only    → "…would leave the pattern suffix 'EQ SP ON ø50 BC'
                        with no callout to head."
```

The multiplier is correctly absent from the second — the author declined the string.

A better decision fell out of the fix. Rather than adding the raw `bolt_circle.diameter` id, I classify the **suffix**: bolt circle → dependent, grid `(3×3)` → rider. That preserves #920's human message instead of listing the same term twice under two spellings, and it needed no changes to your existing tests — which is the sign the classification now sits where the concept does.

### Regressions

Four states, per your request, alongside the pitch-only test:

| authored | result |
|---|---|
| `bolt_circle.diameter` alone | raises, naming the suffix |
| BCD + `bore.diameter` | `EQ SP ON ø50 BC` prints |
| `bore.diameter` alone | suffix absent, **no raise** — omitted, so nothing is orphaned |
| nothing on the pattern | no callout, no raise |

The suffix assertions go through `hole_callout_spec` rather than the placed annotations, so they are about the STRING rather than about what happened to fit on the page.

ADR 0016 now records the dependent-vs-rider rule beside the existing head-dependency section, since that classification is what the next reader needs — the code alone would leave them re-deriving why a suffix term is sometimes waived and sometimes not.

**2082 passed** (full fast tier), all four lint checks clean. mypy also caught that I had reused `dependents`, a name the segment loop binds at function scope; reusing it silently retyped the list.

pzfreo:
Re-review of `dd608e4`: the BCD dependent/rider fix is correct, including the four authored states in the new regression coverage. I found one more P1 in the same completeness contract.

**P1 — off-axis hole locations still bypass the compiled/authored set.**

The location migration covers `render_locations`, but `_locate_off_axis_holes()` at `annotations/holes.py:849` still rebuilds X/Y-hole positions directly from `ctx.part_model` / `_ir_off_axis_holes(model)`. Both orchestrator phases call it unconditionally whenever `feature_keys` is nonempty (`orchestrator.py:450–456`, `496–503`); neither receives the compiled plan.

As a result, authoring only the bore diameter of a side-drilled hole still emits location dimensions the author omitted:

```python
from build123d import Box
from draftwright.model.ir import Frame, HoleFeature, RequestedDimension

hole = HoleFeature(Frame((10, 5, 2), "x"), 4.0, depth=None, through=True)
dwg = build_drawing(
    Box(100, 60, 20),
    model=[hole],
    authored=(RequestedDimension(hole, "bore.diameter"),),
)

print([(n, getattr(a, "label", None)) for n, a in dwg.iter_annotations()
       if n.startswith("dim_loc_")])
# [('dim_loc_side_y3500', '35'), ('dim_loc_front_z1200', '12')]
```

The same leak occurs for a Y-axis hole. There is no diagnostic because these are successfully placed raw dimensions; they simply sit outside `RenderableDimensionPlan.locations`. The existing empty-plan behavioral test does not expose this because `_staircase()` contains no off-axis holes.

This contradicts both “every value-carrying generated path now consumes approved compiled content” and the `location_role()` claim that authored vocabulary and planning cannot disagree. Structurally, off-axis locations need approved location entries with the appropriate end-on span(s), and `_locate_off_axis_holes()` must consume/filter those entries rather than rediscovering all IR holes. Please add a real-build empty/partial-plan regression with X- and Y-axis holes so the boundary guard covers this path rather than only Z-normal `render_locations`.

I would not add another authored check inside `_locate_off_axis_holes`; that would restore the renderer-side suppression convention this PR is removing.

pzfreo:
## P1 fixed — side-drilled positions are compiled (`80649eb`)

Reproduced on both axes exactly as written. You were right that this is the same contract, and right about the constraint: no authored check inside the renderer.

### The real shape of the bug

Three statements of one fact disagreed. `location_role` said a hole is locatable; `plan_locations` said only a Z-normal one is; `_locate_off_axis_holes` drew the X/Y ones from raw IR regardless. My previous comment's claim that "authored vocabulary and planning cannot disagree" was true of the table and false of the pipeline.

The organising principle that resolves it: **a position is compiled wherever it is measured from.**

- `plan_locations` owns the Z-normal ladder — measured from `datum_xy`.
- The compiler owns the ones measured from the **bounding box in a feature's own view** — a slot's near-end offset, and now a side-drilled hole's offset and height.

Splitting by *datum* rather than by *feature kind* is what keeps `_LOCATION_ROLE` a single answer: a hole is locatable, full stop, and where its span comes from is a separate question. Splitting by kind would have put "is a hole locatable?" in two tables again.

**Two approved entries per member, not one.** `dim_loc_side_y3500` and `dim_loc_front_z1200` are separate dimensions on the page; a single "this hole is located" approval would leave the renderer deciding which of the two it covered — the content decision this boundary removes. `discriminator` carries the measured axis, `axis` the hole's own; neither is derivable from the other.

### Consuming, not checking

`_ir_off_axis_holes` is gone, replaced by `_approved_off_axis_holes(plan)` — a filter over approved entries, per your instruction.

I went one step further than filtering: the three sub-passes now take their printed **value** from the entry. They were computing `abs(loc[i] - bb.min[i])` while the compiler computed the same thing, and two formulas for one number is precisely the drift #923 removed everywhere else. The callout keep-out rows come from the plan too — identical on an auto build (every off-axis hole is approved), and under an authored set it stops reserving space for dims that will not be drawn. The coaxial-bore exclusion stays in the renderer: it only ever *removes* an approved entry, which is a drop, not a leak.

### The guard, and why it was blind

Parametrised the empty-plan test over a side-drilled fixture, as you asked. `_staircase()` has no holes at all, which is exactly why the guard passed for a whole round while this path leaked — **a behavioural guard is only as wide as what its fixtures reach**, and that is now stated in the test rather than left as an accident.

Verified as a canary rather than asserted: on `dd608e4` the new case fails with

```
['dim_loc_front_x6000', 'dim_loc_front_z1800', 'dim_loc_side_y3500'] reached the page
from an EMPTY compiled plan
```

while the `staircase` case passes — which is the whole point of adding the second fixture.

Plus the authored regression on both axes:

| authored | X-drilled | Y-drilled |
|---|---|---|
| `bore.diameter` | none | none |
| `bore.diameter` + `location` | 2 dims | 2 dims |

Both halves, because omission only means something if naming it still works.

**2086 passed** (full fast tier), all four lint checks clean. ADR 0016 records the measured-from split.

pzfreo:
Re-review of `80649eb`: the ordinary X/Y inclusion and omission cases are fixed, and the renderer now consumes compiled values. One P1 remains in the new adapter.

**P1 — distinct orthogonal holes at the same point are collapsed by coordinate, silently losing an authored location.**

`_approved_off_axis_holes()` groups entries in `by_member` using only `entry.span[1]` (the 3-D member coordinate). That coordinate is geometry, not feature identity. Two legitimate cross-drilled holes—one X-axis and one Y-axis—can share the same centre. Their four approved entries are then merged into one `_OffHole`; its `axis` is whichever feature appeared first, while the second feature's planar entry sits unused in the merged `approved` dict.

Minimal reproduction:

```python
from build123d import Box
from draftwright.model.ir import Frame, HoleFeature, RequestedDimension

part = Box(100, 60, 20)
x_hole = HoleFeature(Frame((10, 5, 2), "x"), 4.0, depth=None, through=True)
y_hole = HoleFeature(Frame((10, 5, 2), "y"), 4.0, depth=None, through=True)

def locations(features):
    dwg = build_drawing(
        part,
        model=features,
        authored=tuple(RequestedDimension(h, "location") for h in features),
    )
    return [(n, getattr(a, "label", None)) for n, a in dwg.iter_annotations()
            if n.startswith("dim_loc_")]

print(locations([x_hole, y_hole]))
# [('dim_loc_side_y3500', '35'), ('dim_loc_side_z1200', '12')]

print(locations([y_hole, x_hole]))
# [('dim_loc_front_x6000', '60'), ('dim_loc_front_z1200', '12')]
```

The compiled plan correctly contains all four approved entries (`x/y`, `x/z`, `y/x`, `y/z`), but rendering keeps only the first hole's planar location. This is silent loss of an explicitly authored measurement and is feature-order-dependent.

Key the intermediate grouping by identity as well as occurrence—e.g. `(entry.ref, member)` (or equivalent stable feature/member identity)—then let the existing downstream `seen_x`/`seen_y`/`seen_z` logic perform the intentional geometric dedup. That preserves both planar dimensions while still sharing the common Z dimension. Please add the reversed-order assertion too; it is what pins that this is not accidental list-order behavior.

The BCD fix and the prior single-hole off-axis regressions continue to pass in my probes. CI is currently running; lint and Ubuntu 3.14 are green so far.

pzfreo:
## P1 fixed — grouped by identity, not by coordinate (`a38a640`)

Reproduced exactly, both orders. Your diagnosis is right and the one-line summary of the mistake is yours: **a coordinate is geometry; it does not identify a feature.** I keyed on the member point, and a cross-drilled pair is two features at one point.

Grouped by `(entry.ref, member)` — identity **and** occurrence — as you suggested. Both orders now yield all three measurements:

```
[x, y] → dim_loc_side_y3500, dim_loc_front_x6000, dim_loc_side_z1200
[y, x] → dim_loc_side_y3500, dim_loc_front_x6000, dim_loc_front_z1200
```

Each bore keeps its own planar position; the two share one height, exactly as you described — the geometric dedup happens downstream and per axis in `seen_x`/`seen_y`/`seen_z`. Merging in the adapter conflated two different jobs, and only the second one is a dedup.

### One deliberate choice in the regression

The test compares the **measurement** (axis + value) rather than the annotation name. Which *view* carries the shared height is legitimately order-dependent — the Z dim goes to whichever bore's natural strip took it, and it reads `12` either way. Asserting on names would have pinned that accident as though it were the contract, which is the failure mode I have now hit twice in this PR (`_PENDING_VALUE_CARRYING` as a prefix list, `_staircase()` as the only fixture). The docstring says so, so the next reader does not "tighten" it.

Parametrised over both orders per your request, and canaried rather than assumed: on `80649eb` it fails in **both** directions, each missing the other bore's planar dim —

```
x-first: got ['y3500', 'z1200']   — missing 'x6000'
y-first: got ['x6000', 'z1200']   — missing 'y3500'
```

which is what distinguishes a real order-independence guard from one that happens to pass.

**2088 passed** (full fast tier), all four lint checks clean.

### Standing back

Three of the last four findings share one shape: a **rule stated in one place and re-derived in another** — `location_role` vs `plan_locations` vs the off-axis pass; the dependent/rider split vs where a term sits in the string; feature identity vs the coordinate it sits at. The boundary removed the *renderer-checks-a-flag* convention, and what keeps surfacing behind it is the same thing one layer down: two representations of one fact, agreeing until an authored set makes them disagree. Worth noting in case it predicts where the next one is.

pzfreo:
Re-review of `a38a640`: the shared-centre fix is correct in both feature orders. The compiled plan retains all four entries and rendering produces both planar measurements plus one shared height. One P1 remains immediately adjacent to that path.

**P1 — `location` is accepted for an off-axis pattern but compiles to nothing.**

`_LOCATION_ROLE` declares every `PatternFeature` locatable, and `_check_authored_targets()` therefore accepts `RequestedDimension(pattern, "location")`. But `plan_locations()` skips every non-Z feature before its `PatternFeature` branch, while `_compile_off_axis_hole_locations()` handles only `HoleFeature`. An X- or Y-axis `PatternFeature` falls between the two compilers.

Minimal reproduction:

```python
from build123d import Box
from draftwright.model.ir import Frame, HoleFeature, PatternFeature, RequestedDimension

member = HoleFeature(Frame((10, 5, 2), "x"), 4.0, depth=None, through=True)
pattern = PatternFeature(
    Frame((10, 5, 2), "x"), "linear", 3, member,
    members=((10, -10, 2), (10, 5, 2), (10, 20, 2)),
    pitch=15, direction=(0, 1, 0),
)
dwg = build_drawing(
    Box(100, 60, 20),
    model=[pattern],
    authored=(RequestedDimension(pattern, "location"),),
)

assert compile_dimensions(dwg.model()).locations  # fails: ()
assert any(n.startswith("dim_loc_") for n, _ in dwg.iter_annotations())  # fails
assert any(i.code == "authored_omission" for i in dwg.lint())  # also fails
```

The same silent loss remains when `bore.diameter` and `location` are authored together: the side-view bore callout appears, but no requested location does. This is the exact “accepted target leaves the drawing silently blank” failure `_check_authored_targets()` says it prevents.

The structural options are:

1. compile off-axis pattern positions into approved entries and let the existing off-axis renderer consume them; or
2. until that output is supported, make location eligibility orientation-aware so the authored target is rejected rather than accepted and lost.

Given the public statement that a pattern is locatable and the new measured-from split, option 1 appears consistent: extend the bounding-box compiler to `PatternFeature` occurrences, while preserving whatever representative/dedup policy is intended for patterns. Either way, please add X- and Y-pattern authored-location regressions for both `location` alone and `bore + location`. The current single-hole fixture cannot reach this gap because the former raw path deliberately excluded patterns.

CI for the current head is still running; lint is green.

pzfreo:
## P1 fixed, plus a guard on the cause rather than the symptom (`09742a0`)

Reproduced: compiled locations `()`, no dims, no diagnostic.

### Why option 2, not option 1

I checked what the engine actually does before choosing. **The auto path draws nothing for an off-axis pattern** — the off-axis pass excluded patterns by construction, deliberately and since #595. So option 1 is not a boundary fix; it is *new output*, with its own layout consequences on the contended side/front-below strips, and it belongs in its own change with its own review. The vocabulary should tell the truth about today's engine instead.

Done as **one answer rather than a second orientation rule**, which is the part that matters:

```python
location_datum(feature) -> "datum_xy" | "bbox" | None
```

read by `plan_locations`, by the bbox compilers, and by the authored vocabulary. `plan_locations` had restated the orientation half inline — that is precisely how it came to disagree with the kind table. A new invariant test asserts *role-without-datum is impossible* across every locatable kind in every orientation, so this cannot arrive a third time from one side only.

### The `bore + location` case was a second bug

It survived the first fix. `_check_authored_targets` asked whether the **feature** matched anything, so a valid `bore.diameter` entry vouched for the invalid `location` beside it — one good entry laundering every bad one on the same feature. Split `_authored_addresses(authored, feature, param)` out so the check is per **entry**.

Regressions on both axes × both role sets, plus the false-positive half (a Z-normal pattern location still works). Canaried: all five fail on `a38a640`.

---

## And the prevention question

You asked how to stop this class recurring. Three of the last four findings shared one shape — **a rule stated in one place and re-derived in another, agreeing until an authored set makes them disagree**. So I added a guard on the cause:

**`tests/test_label_provenance.py`** — a renderer calling `_fmt(x)` is turning a number into printed text, which means the number reached it *as a number* rather than as the compiler's `value_text`. Every defect of this class looked exactly like that: the height ladder rebuilding rungs from `levels`, `render_slots` formatting `s.width`, `_add_furniture` printing `feat.pitch`, the off-axis pass recomputing `abs(loc[i] - bb.min[i])`.

Shrink-only per function (the `test_private_test_attr_reads` pattern), with three supporting properties: no new/grown site; **no stale ceiling** (slack in a shrink-only budget silently absorbs the next regression); every survivor carries a written reason. Excluding f-strings, log/lint calls and `_text_size` is what makes it 15 sites rather than 61 — small enough to actually read. Canaried against a synthetic site.

This complements the behavioural guard rather than duplicating it: **a behavioural test only covers the paths its fixtures reach**, which is exactly how two dimensional paths stayed outside the boundary for a round each. The source ratchet covers every path whether a fixture walks it or not.

### Writing it immediately found a contract inconsistency in this PR's own code

`_compile_off_axis_hole_locations` emits one entry **per measured axis** with its value; `_compile_locations` emitted one entry **per feature** with `value_text=""`. The non-Z pocket branch of `render_slots` drew *two* dims from that single entry — it would have printed an empty label the moment it read `value_text`. Non-Z pocket locations are now per-axis like every other family. Also fixed: `render_slots`/`render_envelope`/`render_boss_heights` re-formatting values that already carried text, and the three off-axis sub-passes now label from the entry.

The Z-normal ladder is the one location family still without per-axis values, and the budget says why rather than leaving it to be rediscovered: `render_locations` groups refs across features and dedups per axis *before* it knows which dims exist, so an entry per axis would approve a mark whose existence the renderer decides. Splitting it means moving that grouping into the compiler — worth its own change.

**2105 passed** (full fast tier), all four lint checks clean.

pzfreo:
Re-review of `09742a0`: the off-axis pattern target fix is correct. Both invalid role sets now raise per entry, and the supported Z-normal pattern case remains valid. The focused boundary/provenance tests pass locally (`62 passed`). I found one P2 in the newly added prevention guard.

**P2 — the label-provenance ratchet excludes real drawing labels whenever they use an f-string.**

`tests/test_label_provenance.py::_excused()` treats every `ast.JoinedStr` as “a message about a drawing” and skips its entire subtree. That is not a semantic distinction: production renderers build many labels with f-strings. Examples currently excluded from the count include:

- `from_model.py:858`: `label = f"ø{_fmt(dia)}..."`
- `from_model.py:1350–1351`: chamfer labels `C…` / `… × …°`
- `from_model.py:1607`: across-flats label
- `from_model.py:1678`, `1708`, `1715`: groove/pocket/slot labels
- `holes.py:1224`: pitch label `f"{n - 1}× {_fmt(pitch)}"`

Consequently, this exact regression passes the new ratchet:

```python
label = f"ø{_fmt(raw_feature.diameter)}"
```

while the equivalent `label = _fmt(raw_feature.diameter)` fails. The syntax used to decorate a value must not decide whether it is compiler-owned. The comment that “an f-string [is] a message about a drawing” is disproved by the renderer itself.

Please remove the blanket `ast.JoinedStr` exemption. Exempt prose by its call-site context (`record_issue`, logging, etc.), which the test already does through `_NOT_A_DRAWN_LABEL`; drawn f-string labels should then either consume `value_text` or appear explicitly in the shrink-only budget with their reason. Add a synthetic canary using `_fmt(raw)` *inside a drawn f-string*, since the stated canary evidently covered only the direct-call form.

There is a second AST robustness concern worth fixing at the same time: the outer function walk descends into nested `FunctionDef`s, then those nested functions are walked again, and counts are stored by `module.function` with assignment rather than accumulation. Repeated nested names such as `f` can overwrite one another. Stop traversal at nested function boundaries (and give them a qualified key), or use a visitor that maintains the lexical function stack. Otherwise the “no stale ceiling” property can be satisfied by whichever same-named nested function happens to overwrite last.

This does not invalidate the functional location fix, but the new ratchet currently provides false confidence about the class it claims to prevent. CI is still running; lint is green.

pzfreo:
## P2 fixed — both findings correct, and the first one is embarrassing (`ce9a77f`)

**The f-string exemption was the same mistake this PR keeps deleting: a heuristic standing in for the real question.** I wrote that an f-string is "a message about a drawing", and the renderers disprove it a dozen times over — `f"C{...}"`, `f"ø{...}"`, `f"{n - 1}× {...}"` are labels *on* drawings. Confirmed your regression exactly: under the old rules `label = f"ø{_fmt(raw.diameter)}"` counted **0**.

A guard that classifies by syntax rather than by provenance is the `_VALUE_FREE`-prefix-list error one level up, and I made it while writing the test that exists to prevent that class. Exemptions are now call-site facts only.

**Attribution was broken twice over**, as you said. The walk descended into nested `def`s from both their own entry *and* their parent's, and keyed on the bare name with assignment — nine duplicated names (`_place`, `_build`, `_drop` ×6) overwrote one another. Traversal now stops at each nested `def`, which gets its own dotted key (`from_model.render_step_lengths._redraw_y`).

Canaried on all three shapes: direct, inside an f-string, and nested — plus the false-positive half, that a log line stays exempt. The first version was canaried on the direct form alone, which is precisely why it looked convincing.

### What closing the hole revealed

**39 real sites where 15 had been claimed.** Fixed 13:

- the six label composers (`_chamfer_label`, `_fillet_label`, `_flat_label`, `_groove_label`, `_pocket_label`, `_slot_label`), `_dia_label`, and the pitch dims now take the compiler's `value_text` rather than a float to format;
- grid pitch **selection** stays numeric — choosing *which* entry and printing it are different jobs, and conflating them is how this class starts;
- `_chamfer_label(leg_text, leg, ch)` takes the leg twice for the same reason: `leg_text` reaches the sheet, `leg` feeds the equal-leg comparison. My first attempt passed the approved entry, which broke three tests that assert the label form and have no plan object — a pure string function should not need plan machinery to test.

The remaining 26 are budgeted with specific reasons rather than a shrug. The largest is the ø-leader item tuple, whose `dia` float is *both* printed and used for tip geometry (`az - dia / 2`); carrying the approved text alongside it is a five-call-site change I would rather name than rush at the end of a long session. A chamfer's **angle** stays too, and its reason is real: it is a form discriminator with no `DimParameter` at all — the IR gap `_FACTS` already records.

**2113 passed** (full fast tier), all four lint checks clean.

---

You asked earlier how to prevent this class. This exchange is the answer in miniature: the ratchet only became load-bearing once it was canaried against **every shape it claims to cover**, not the one that was easy to write. A guard nobody has seen fail is a guard nobody has tested — which is now itself a test in the file.

pzfreo:
## Follow-ups filed, and #883 corrected

Everything I deferred in this PR was recorded in code comments, the budget table and the ADR — which is the kind of record that gets lost. Now on the tracker:

| | |
|---|---|
| **#926** | Hole callouts (`hc_`) still take the legacy `DimensionGroup` surface — the last renderer outside the boundary. Behaviourally correct today, structurally outside; done when `_PENDING_VALUE_CARRYING` is `("pmi_",)`. |
| **#927** | Draw down the label-provenance budget — the 26 remaining sites, in three independent pieces (ø-leader item tuple, Z-ladder per-axis values, step-chain repeat-run collapse), with the permanent entries and their reasons listed as out of scope. |
| **#928** | Exhaustive authored-role property test — prevention item 3, the one this PR did **not** build. Enumerated from `feature.parameters()` + `location_role()` rather than hand-listed, so a new kind joins by existing. |
| **#929** | Off-axis pattern positions — recording the decision not to compile them, with what supporting them would cost, so it stays a choice rather than a silent gap. |

**#883 was factually wrong after this PR** and is rescoped. It said "location dimensions have no addressable identity"; they have it per feature now. What remains is per-**member** granularity, which is a naming question — and the issue now says explicitly that its original framing (settle the design questions before locations can be addressed at all) turned out to be false for the completeness contract, since omission is well-formed at either granularity. It also no longer blocks the `"location"` role in #872.

Two notes on where the boundaries fell:

- **#927 piece 2 overlaps #714** (dedup the mirrored X/Y render blocks). Moving `render_locations`' cross-feature grouping into the compiler and removing the X/Y mirror touch the same code; worth doing together, and #927 says so rather than letting two issues quietly fight over the same function.
- **#926 retires one #927 site** (`callout_from_spec.f`), so that entry is listed as out of scope in #927 rather than counted twice.

The one thing I have *not* filed is a general "authored sets are incomplete for kind X" issue, because after this series I no longer believe that shape exists — every instance turned out to be a specific duplicated rule, and a vague umbrella issue would just absorb the next one instead of naming it.

pzfreo:
Re-review of `ce9a77f`: **clean — no further findings.**

I verified the previous P2 is fixed rather than merely budgeted around:

- f-string labels are now counted; the explicit `f"ø{_fmt(x)}"` canary exercises the exact missed shape;
- nested functions are attributed lexically under dotted keys and no longer double-counted or overwritten by same-named siblings;
- exemptions are based on prose call sites, not label syntax, with a false-positive log canary;
- the newly migrated label composers consume approved `value_text`, while remaining numeric-format sites are visible in an exact, no-slack budget with reasons.

I also rechecked the functional chain from the preceding rounds: unsupported off-axis pattern locations are rejected per authored entry, supported Z-normal locations remain valid, side-drilled inclusion/omission obeys the plan, cross-drilled features retain both planar measurements independent of feature order, and the BCD dependency behavior remains coherent.

Local focused result: **70 passed** (`test_label_provenance.py` + `test_compiled_plan_boundary.py`). `git diff --check` is clean. GitHub is fully green across lint, all platform/Python jobs, and both Codecov checks; merge state is `CLEAN`.

From this review, PR #925 is ready to merge. The explicitly deferred structural debt is now tracked in #926–#929 rather than hidden in the PR.

Issue #922, opened 2026-07-29
Emit a Sheet script for a model with an authored dimension set


pzfreo:
`emit_sheet_script()` currently **refuses** a `PartModel` carrying `authored_dimensions` (added in #921 after review round 5). Before that it silently emitted `sheet.auto_dimensions()`, so running a script generated from an authored model restored every dimension the model had omitted — "width only" became the full automatic set. That is the #707 class of divergence: a generated script that draws something other than the model it came from.

Refusing removes the wrong output. This issue is the missing capability.

## Why it isn't a small fix

The emitted script binds no variables for features — `sheet.hole(...)`, `sheet.add(EnvelopeFeature(...))` — so the only way to write a declaration today is by **position**:

```python
sheet.dimension(3, "width")
```

The documented editing workflow for these scripts is to comment a feature line out, edit a value, and re-run (ADR 0011 Amdt 1). That shifts every later index and silently retargets the dimensions onto their neighbours — precisely the position-addressing defect ADR 0016's identity work exists to remove (#910/#912), rebuilt inside the artefact we hand the user.

So the prerequisite is **nameable identity for emitted features**, e.g.:

```python
bore = sheet.hole(diameter=8, at=(-30, 0, 17), axis="z")
...
sheet.dimension(bore, "bore.diameter")
```

which needs a naming scheme (stable, readable, collision-free) and a decision about which verbs return handles — `sheet.step_level(...)` returns the `Sheet`, not a handle, so it cannot be named at all today.

## Relationship to the epic

ADR 0016's epic (#867) scopes out "the emitter dimension-mirror" for the same reason: it needs a surface the ADR leaves to a follow-up. This is that work, narrowed to the authored-set case.

## Acceptance

- Emitted features can be referenced by name, not position.
- A round trip preserves the dimension source: an authored model emits `dimension(...)` declarations; a detected model emits `auto_dimensions()`.
- A round-trip test asserts the emitted script's drawing carries the same dimensions as the source model's — the assertion that would have caught the original bug.
- The refusal in `emit_sheet_script` is removed.

Refs #874, #876, #867, #707.

PR #930, opened 2026-07-30
fix(#878): a blind hole round-trips its blindness without a measured depth

### LINKED ISSUE #878: Emitted Sheet script loses blindness when a hole has through=False but depth=None
Found during the adversarial pass on #877 (child of #867). **Same rule as #868, different site, different symptom** — and it survives PR0.

Both hole emitters guard the blindness fact on the *depth* being present:

```python
# sheet_emit.py:76-78  (_hole_line, standalone hole)
if not f.through and f.depth is not None:
    line += f".depth({_n(f.depth)})"

# sheet_emit.py:96-98  (_member_hole_str, pattern member)
if not m.through and m.depth is not None:
    kw.append(f"depth={_n(m.depth)}")
    kw.append("through=False")
```

So a `HoleFeature(through=False, depth=None)` emits **neither** `depth=` nor `through=False`. `declare.hole` defaults `through=True` (`model/declare.py:219`), so the round-tripped script declares a **through hole**. The blindness is silently lost on the emit path — the script/CLI parity class of #707.

## Reachability

`declare.hole(diameter=6, at=(0, 0, 0), axis="z", through=False)` with no `depth=` is accepted today — nothing validates the pair (`_require_positive(depth=depth)` tolerates `None`). ADR 0011 makes `PartModel` a **public input**, so a caller can construct this state directly. A detection edge (`detect.py:96`, `through=(h.bottom == "through")` with an unmeasured depth) may also reach it — unverified.

## Why it is now visible

#868 fixed the *render* side: `hole_callout_spec` reads `hole.through`, so the callout is correct for this hole. The emit side still infers the fact from the dimension's presence, so render and emit now disagree.

## Candidate fixes

1. **Emit `through=False` whenever `not f.through`, independent of depth** (one line at each site). The emitter must round-trip whatever the IR holds. This is the fix that also covers any detection-sourced case.
2. **Make `declare.hole(through=False, depth=None)` raise** — a blind hole with no depth is arguably not a describable hole, consistent with #631 (raise rather than silently drop). Separate question about whether the state should exist; on its own it leaves detection-sourced cases broken.

I would do 1, and treat 2 as a follow-on judgement call.

**Gate:** a canary that must fail at the pre-fix commit — emit a `through=False, depth=None` hole, re-execute the script, assert the rebuilt feature is still blind.

pzfreo:
Closes #878. Also cleans up the ADR 0016 docs, which were wrong about the work that just merged in #925.

## The bug

Both hole emitters guarded `through=False` on the *depth* being present:

```python
if not f.through and f.depth is not None:   # _hole_line
if not m.through and m.depth is not None:   # _member_hole_str
```

So `HoleFeature(through=False, depth=None)` emitted **neither** `through=False` nor a depth, and `declare.hole` defaults `through=True`. The regenerated script declared a **through hole** — the blindness lost on the emit path, the #707 class of divergence.

This is **#868's rule at a different site**: a fact is not inferred from a dimension's presence. #868 fixed the render side (`hole_callout_spec` reads `feat.through`), which is what made this visible — the callout became correct for such a hole while the emitted script stayed wrong, so render and emit disagreed about the same feature.

## The fix

Option 1 from the issue: emit the fact independent of the depth, at both sites. Option 2 (make `declare.hole(through=False, depth=None)` raise) was not taken — on its own it leaves any detection-sourced case broken, and whether that state should exist at all is a separate judgement.

```
before:  sheet.hole(diameter=6, at=(0, 0, 0), axis="z")
after:   sheet.hole(diameter=6, at=(0, 0, 0), axis="z", through=False)
```

A measured blind hole is unchanged (`.depth(4)` already sets `through=False`).

## The canary

Asserts the **rebuilt feature**, not the emitted text: the property is that the round trip preserves the fact, and a substring check would pass on any spelling that happened to contain the token. Both emitters are exercised — they are independent sites with the same bug, and the pattern-member one is reached through a different call path.

Verified against `main`:

```
E   AssertionError: sheet.hole(diameter=6, at=(0, 0, 0), axis="z") rebuilt a THROUGH hole
```

## Docs, in the same PR because they were wrong about the same work

- **CLAUDE.md's ADR list stopped at 0015.** ADR 0016 was entirely absent despite its epic being complete — added, including what did *not* ship (the emitter dimension-mirror, #922) so the gap is visible from the index rather than only from deep in the ADR.
- **ADR 0016's scheduling header carried an open question that implementation settled.** It asked whether phase 2 was blocked on the global recompose and floated that it might not be. It was not: a `Sheet` script's authored set is known before `build()`, so the planner marks the omissions and the compiled plan withholds them — nothing is drawn that must be un-drawn. Recorded as answered, along with the note that what the work *did* need was unanticipated and structural (the compiled-plan boundary), not scheduling.
- The "honest reconciliation needs the full recompose" constraint now separates its **settled** pre-build half from its still-open post-build one, instead of marking the whole thing open.

**2114 passed** (full fast tier), all four lint checks clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -332,6 +332,30 @@ Current ADRs:
   groups are residual debt tracked by #754) and the lint/coverage carve-out
   stated properly. New kinds must add every applicable IR, planning, rendering,
   coverage, and test surface while keeping orientation data-driven.
+- **0016** — **Accepted; epic #867 complete** (PR0–PR8, #868–#876): **declared
+  dimensioning intent**. `sheet.dimension(feature, role)` is *referential* — it names
+  a measurement and carries no number; the engine still derives the value from the
+  geometry and owns placement. Three parts landed:
+  - **A build must say where its dimensions come from** (#874, breaking): either
+    `auto_dimensions()` (the planner's set, optionally augmented by `add_dimension`)
+    or an authored set of `dimension(...)` declarations. Mutually exclusive, because
+    "everything the planner chooses, plus these" and "only these" cannot both hold.
+  - **Omission from an authored set means suppression** (#876) — on *every* generated
+    dimensional path, positions included. A dimension the author cannot address is a
+    dimension the author cannot omit, so `location` became addressable per feature
+    (`planner.location_datum` is the single eligibility answer; per-*member* identity
+    remains #883).
+  - **Amendment 1 — the compiled-plan boundary**: renderers may emit dimensional
+    content only from `model/compiled.py`'s `RenderableDimensionPlan`. *Suppression is
+    not a flag renderers check, it is content they never receive* — `ApprovedDimension`
+    has no `suppressed` field. Guarded on both the symptom
+    (`tests/test_compiled_plan_boundary.py`: an empty plan draws nothing) and the cause
+    (`tests/test_label_provenance.py`: a renderer that formats a number got it as a
+    number rather than as the compiler's `value_text`). Hole callouts (`hc_`) are the
+    one renderer still on the legacy surface (#926); the label budget drawdown is #927.
+  **Not** shipped: the emitter dimension-mirror (phase 4). `emit_sheet_script` refuses a
+  model with an authored set rather than silently writing `auto_dimensions()`, because
+  naming a feature in a generated script would have to address it by position — #922.
 
 ## Dependencies
 
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -4,22 +4,28 @@
 - **Date:** 2026-07-26 (accepted 2026-07-27; implementation epic **#867**)
 - **Deciders:** Paul Fremantle (pzfreo)
 
-> **Scheduling reality (corrected 2026-07-27).** Phase 1 (identity + `add_dimension`)
-> is reachable on today's machinery. Phases 2–4 were written as sitting behind "the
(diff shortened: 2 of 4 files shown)


pzfreo:
Review of `511d617`: **clean — no blocking findings.**

The fix preserves the right distinction: `through` is an engineering fact and `depth` is an optional measurement. Both independent emit paths now round-trip `HoleFeature(through=False, depth=None)` without inventing a depth or falling back to `declare.hole`'s `through=True` default:

- standalone holes emit `through=False` when no depth exists;
- pattern member templates emit `through=False` independently and include `depth=` only when measured;
- measured standalone blind holes retain the existing `.depth(...)` spelling, whose declared semantics also set `through=False`.

The tests assert the rebuilt IR state rather than a source substring and exercise both emitters, which is the correct property-level canary. I also checked that these helpers are the paths used by full generated scripts and that existing measured-depth cases remain covered.

Local result: **96 passed** in `tests/test_sheet_emit.py`; `git diff --check` is clean. GitHub lint and the completed platform jobs are green; four matrix jobs are still pending at review time, with no failures.

Non-blocking style note only: `_hole_line` could append `through=False` for every blind hole and then append `.depth(...)` when present, making the fact visually independent in both branches. The current output is nevertheless correct and preserves established generated-script text for measured blind holes, so I would not hold the PR for that change.

The ADR/CLAUDE updates accurately distinguish the shipped pre-build authored-set behavior from the still-open post-build recompose and emitter-mirror work. From this review, #930 is ready once the remaining CI jobs finish green.

PR #931, opened 2026-07-30
feat(#922): every declaration verb hands back a nameable feature

pzfreo:
First of three for #922. Prerequisite only — no emitter change yet.

## The gap

A `dimension(feature, role)` line names a feature, so every declaration verb has to return one. **Eleven did not:** nine returned the `Sheet` (`chamfer`, `fillet`, `flat`, `groove`, `plate`, `step_level`, `pattern`, `pocket_pattern`, `slot_pattern`), plus `add` and `measured_dimension`. Those features were unreferenceable — `sheet.dimension(chamfer, "chamfer")` could not be written.

#922 describes this as "the feature verbs bind no variables". That understates it in one direction and overstates it in another: `hole`/`boss`/`step`/`slot`/`pocket`/`pad` already returned token-carrying handles and `sheet.dimension(handle, role)` already worked. The real gap was these eleven.

## Why uniform, rather than only where it's needed

`add` is the case that settles it. The **envelope** is emitted through that escape hatch rather than through `envelope()`, so while `add` returned the sheet, ADR 0016's own worked example —

```python
sheet.dimension(env, "width")
```

— could not be written against a generated script. Naming would have been uniform across the verbs and silently absent for one feature in the middle of the file. That is worse than absent everywhere, because it is invisible until someone tries it.

## This is not only a dimension feature

It removes **positional addressing** from the generated artefact. Today, commenting a feature line out shifts every later index, so anything addressing by position silently retargets onto a neighbour. A binding turns that into a `NameError` at the line you edited — the same defect ADR 0016's identity work removed everywhere else in the codebase, still present in the file we hand the user.

## Guards, both of which caught something

**The #912 identity invariant fired immediately** — every handle-returning verb must prove its token survives a reorder. I could have added eleven names to `_SAME_PATH_AS_ENVELOPE` and moved on. Instead `test_the_same_path_verbs_really_share_the_route` parses each exempt verb and asserts it still appends exactly one feature and returns the envelope handle verbatim, so the exemption is a checked claim rather than a comment. *An exemption list nobody verifies is how a guard stops guarding.* Writing it caught a bug in itself first: filtering every `ast.Expr` to drop the docstring also drops `self._features.append(...)`, which is an expression statement — it reported "appends 0 features" for every verb.

**A new nameability ratchet** fails any verb that appends a feature and returns `self`. That is what found `measured_dimension`; my hand-written list had ten of eleven.

## Compatibility

`_Params.__getattr__` forwards to the owning `Sheet`, so chaining is unchanged:

```python
Sheet(part).chamfer(...).hole(...)   # still works
```

Additive for existing scripts.

**2123 passed** (full fast tier), all four lint checks clean.

### DIFF
--- a/src/draftwright/sheet.py
+++ b/src/draftwright/sheet.py
@@ -718,11 +718,18 @@ class Sheet:
 
     # -- feature declaration --------------------------------------------------
 
-    def add(self, feature) -> Sheet:
+    def add(self, feature) -> _Params:
         """Append a pre-built IR :class:`~draftwright.model.Feature` (escape hatch for
-        the constructors this façade does not surface directly, e.g. PMI)."""
+        the constructors this façade does not surface directly, e.g. PMI).
+
+        Returns a handle, like every declaration verb (#922). It matters here more than it
+        looks: the ENVELOPE is emitted through this escape hatch rather than through
+        :meth:`envelope`, so while `add` returned the sheet, `sheet.dimension(env, "width")`
+        — ADR 0016's own worked example — could not be written against a generated script.
+        Naming would have been uniform across the verbs and silently absent for one feature
+        in the middle of the file, which is worse than being absent everywhere."""
         self._features.append(feature)
-        return self
+        return _Params(self, len(self._features) - 1)
 
     #: The keyword set that identifies a call to the pre-rename `dimension(...)`. `kind` alone
     #: would do today, but the referential form may grow keywords of its own, so the test is the
@@ -786,7 +793,7 @@ class Sheet:
         lower_tol: float | None = None,
         source: str = "sheet",
         source_kind: str | None = None,
-    ) -> Sheet:
+    ) -> _Params:
         """Declare a drafting dimension from explicit **measured** values.
 
         Named for what it carries (ADR 0016 / #873). ``dimension`` is *referential* on
@@ -818,7 +825,11 @@ class Sheet:
                 source_kind=source_kind,
             )
         )
-        return self
+        # A handle like every other declaration verb (#922). A measured dimension carries its
+        # own number, so a referential `dimension(handle, role)` on it will correctly raise —
+        # but it is still a declared feature in the emitted script, and exempting it would put
+        # one unnameable line back in the middle of a file where naming is otherwise uniform.
+        return _Params(self, len(self._features) - 1)
 
     def of(self, ref) -> _Hole | _Dim:
         """A decoratable handle onto an **existing** feature — the hybrid seam (#463).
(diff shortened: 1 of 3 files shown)


pzfreo:
## Codex adversarial review — two findings, both fixed (`HEAD` updated)

Ran an open-ended review (no defect-class checklist, explicit licence to reject the approach) against the PR head in an isolated worktree. It found two, and **both were mine** — it landed on the two things I'd flagged as most attackable, which is a reasonable sign the prompt was open enough to be useful rather than agreeable.

### 1. Medium — a silently dropped tolerance. A regression **this PR introduced**.

```python
h = sheet.measured_dimension(...)
h.tolerance(0.2)      # succeeds, records nothing
```

`_Params.tolerance()` iterates `_roles()`, empty for a parameterless feature, and returns `self`. Unreachable before this PR, because those verbs returned the `Sheet` — handing out a handle that advertises aspects the feature cannot carry is exactly what opened the path. A drafting instruction vanishing in silence is the failure this codebase ranks below a visible raise (#630/#631).

Now raises, naming the kind, and pointing a measured dimension at its own `upper_tol=`/`lower_tol=` kwargs.

I did **not** take the suggested type split (identity-only base class, `_Params` as a subclass). `add()` is polymorphic — it accepts any IR feature, parameterful or not — so the *type* cannot know at declaration time what the *call* can determine. The check belongs at the call site, and the split would have left `add(EnvelopeFeature(...))` unable to tolerance a feature that genuinely has parameters.

### 2. Low — the ratchet asserted syntax, not the property

It matched `source.rstrip().endswith("return self")`. A verb returning `None`, falling off the end, or returning anything unrelated passed while leaving its feature unnameable — and the "property" test parametrized six verbs while claiming to cover all of them. Both criticisms correct.

Replaced with the behavioural contract the review specified: declaration verbs discovered from source, **every one** invoked, handle resolved, feature list reordered, resolution re-checked. The reorder is the half that matters — a handle carrying a bare index passes a resolve-once check and then names a neighbour the moment a script comments a feature out, which is the positional-addressing defect this work exists to remove. Plus a table-coverage ratchet: a new verb with no entry fails rather than being skipped.

Canaried against the review's exact case — `chamfer` returning `None` now fails three tests, and I confirmed the old check would have passed it.

### One note on myself

That is the third guard this session that asserted a proxy instead of the property (`_VALUE_FREE` as a prefix list; the f-string exemption in the label ratchet; this). The pattern is consistent: I write the guard *after* the fix, so the shape I check is the shape I just produced. Worth watching for in review rather than trusting me to catch it.

**2153 passed, 1 skipped** (full fast tier), all four lint checks clean.

pzfreo:
Review of `61f6e98`: **clean — no further findings.**

I reviewed the approach as an API and identity change, not just the eleven return statements:

- all feature-declaration verbs now return the same token-resolved `_Params` route, so handles survive reorder rather than retaining positional indices;
- `add()` correctly participates—important because emitted envelopes use that escape hatch;
- fluent chaining remains compatible through `_Params.__getattr__`;
- the two `_Params` method-name overrides are deliberate: `note()` targets the declared feature by default, while `tolerance()` records supported parameters and now raises for parameterless features instead of silently dropping intent;
- measured dimensions give a useful error directing callers to `upper_tol=`/`lower_tol=`;
- the revised behavioral table discovers every declaration verb, fails if a new verb lacks a fixture, resolves every returned handle, and verifies identity after reorder. This is materially stronger than the removed source-ending proxy.

I also checked the uniform-route exemption in `test_sheet_identity_invariant.py`: it verifies exactly one append and the token-bearing envelope return path, while genuinely different verbs remain in dedicated scenarios.

Local focused result: **242 passed, 1 skipped** (`test_add_dimension.py` + `test_sheet_identity_invariant.py`). `git diff --check` is clean. GitHub lint is green and Codecov reports full patch coverage; the platform test matrix is still pending at review time, with no failures.

The scope is appropriately narrow for the first #922 slice: it establishes nameable feature identity without prematurely changing emitted scripts or removing the authored-model refusal. From this review, PR #931 is ready once CI finishes green.

PR #932, opened 2026-07-30
feat(#922): the emitted script binds a name for every feature

pzfreo:
Second of three for #922. Builds on #931 (every declaration verb hands back a handle).

```python
hole1 = sheet.hole(diameter=8, at=(-20, 0, 4), axis="z")   # ⌀8 THRU
hole2 = sheet.hole(diameter=6, at=(20, 10, 4), axis="z")   # ⌀6 THRU
envelope1 = sheet.add(EnvelopeFeature(...))                # envelope 80 × 8 × 50
```

## Why this is not (only) a dimension feature

It removes **positional addressing** from the artefact we hand the user. The documented workflow is to comment a feature line out and re-run (ADR 0011 Amdt 1) — which shifts every later index, so `sheet.of(2)` silently retargets onto a neighbour. `sheet.of(hole1)` fails at the line you edited.

Verified by *running* an edited script rather than asserting about it:

```
NameError: name 'hole1' is not defined. Did you mean: 'hole'?
```

## Uniform, not only where referenced

Binding only the features a `dimension(...)` line names would make the emitter's **output** depend on the **dimension source** — a formatting decision keyed on something unrelated to formatting, and the exact coupling shape this codebase has been paying down. It is also less useful: the benefit above has nothing to do with dimensions.

## Decisions

- **Names derive from `Feature.kind`**, which is already a valid identifier — not from the `_NOUN` display table, whose values have spaces (`"measured dimension"`, `"step-ladder"`). Two spellings of one fact is the drift this codebase keeps paying for.
- **The numeric suffix is unconditional.** The generated script *imports* `hole` for pattern members, so an unsuffixed `hole = ...` would shadow it and break every later pattern line. Tested explicitly.
- **A kind with no declarative verb binds nothing** — its line is a comment, and `rotational1 = # …` does not parse. Easy to miss, because the binding is applied in a loop where a comment looks like any other line.

## A wart the naming scheme creates, fixed

That `Did you mean: 'hole'?` points at the imported constructor. Following it reached `Sheet.of` with a function and leaked:

```
AttributeError: 'function' object has no attribute 'bounding_box'
```

from the match-by-object path, which assumed anything not a handle/index/`Feature` was a solid. Now it says what it expected and what it got. Pre-existing, but this change puts a signpost on the road to it.

I considered renaming the import (`hole as hole_spec`) to kill the near-collision and rejected it: `sheet.pattern(hole(diameter=6, …))` is the shape the docs use, and the fix belongs in the error message rather than in contorting the generated script around one editor hint.

## Seven tests failed on the first full run, and both causes were real

- **The header promised `sheet.dimension(hole1, ...)`, which the emitter does not yet write.** A test asserting the emitted script contains no `sheet.dimension(` caught me documenting PR3's capability as though it had landed. Removed rather than excused — that is the assertion doing its job.
- **Tests locating a feature line by prefix, then `eval`ing it.** A bound line is an assignment, which `mode="eval"` cannot parse; and bare-substring matching hits the header's own `sheet.hole(my_bore)` example. Two helpers now do it properly — a declaration is an assignment whose value is that call.

## Tests

Four naming properties, canaried: stripping the binding fails three of them. Plus the `of()` error and the comment-line case.

**2158 passed, 1 skipped** (full fast tier), all four lint checks clean.


pzfreo:
Review of `cd1fbef`: **clean — no blocking findings.**

The naming scheme is appropriate for this second #922 slice:

- every emitted declaration statement is bound uniformly, independent of dimension source;
- names derive from the IR kind and use an unconditional numeric suffix, keeping built-in names valid, distinct, readable, and clear of the generated script's constructor imports;
- counters are per kind and advance only for emitted statements, so unsupported/comment-only kinds neither create invalid `name = # ...` syntax nor consume names;
- the binding wraps the whole fluent declaration expression, including blind-hole `.depth(...)` chains, so the variable receives the final token-based handle;
- commenting out a named declaration now produces a local `NameError` for downstream references rather than shifting a positional index onto a neighbour.

I also checked the test changes that parse/evaluate feature lines: treating a declaration as an assignment whose RHS is the call is the right update, and avoids matching examples in the generated module docstring. The executed-script test confirms the bindings resolve to the intended feature kinds, rather than merely asserting identifier syntax.

The `Sheet.of()` error hardening is sensible and scoped: the newly likely constructor-function mistake now fails at the public boundary with an actionable `ValueError`, while handles, indices, in-sheet features, and build123d objects retain their existing routes.

Local result: **101 passed** in `tests/test_sheet_emit.py`; `git diff --check` is clean. GitHub lint and all completed platform jobs are green; five matrix jobs remain pending at review time, with no failures.

The scope is correctly limited: this PR makes names available but does not claim the authored `dimension(...)` mirror has landed yet. From this review, #932 is ready once the remaining CI jobs finish green.

PR #933, opened 2026-07-30
feat(#922): emit the authored dimension set as declarations

### LINKED ISSUE #922: Emit a Sheet script for a model with an authored dimension set

pzfreo:
Closes #922. Third and last of the series — #931 (every verb hands back a handle), #932 (the emitter binds a name per feature), this.

```python
hole1 = sheet.hole(diameter=8, at=(-20, 0, 4), axis="z")   # ⌀8 THRU
envelope1 = sheet.add(EnvelopeFeature(...))                # envelope 80 × 8 × 50

# ── Dimensions ─────────────────────────────────────────────
# THIS IS THE COMPLETE SET (ADR 0016). A measurement with no line here is omitted
# deliberately — comment a line out to drop that dimension, add one to declare it.
sheet.dimension(hole1, "bore.diameter")
sheet.dimension(envelope1, "width")
```

## Why it was refused, and what changed

`emit_sheet_script` wrote `sheet.auto_dimensions()` for any model, so a script generated from an authored model restored every dimension that model had omitted — "width only" became the full automatic set (#921 review round 5). Refusing removed the wrong output; it did not add the right one.

The blocker was that the only way to name a feature was its position, and the documented workflow is to comment a feature line out and re-run — which shifts every later index and retargets the declarations onto their neighbours. That is the positional-addressing defect ADR 0016 removed everywhere else, and rebuilding it inside the artefact we hand the user would have been worse than refusing.

## The refusal is narrowed, not deleted

A kind with no declarative verb emits a **comment**, so it binds no name and a declaration on it has nothing to reference. Emitting the rest would produce a script drawing a different set from its source — exactly what the blanket refusal prevented. So it still refuses, for that case alone: from *"any authored model"* to *"an authored dimension on an unemittable feature"*.

## Design notes

- **Names are computed once and returned** from `_feature_block`, not recomputed by the dimension block. A second derivation here would be *silently* wrong rather than loudly — a mismatch emits `sheet.dimension(hole2, …)` naming the wrong hole, and it still runs.
- **Declarations follow the features**, since each names a binding. The source stays stated up front as a pointer to them, so a reader still learns at the top whether omission means anything in this script.
- **Double-quoted roles.** `!r` rendered single quotes and made the generated file read as two dialects next to `axis="z"`.

## Acceptance is on drawn annotations, not emitted text

Build the model directly, run the generated script, compare the annotation sets. A text assertion passes on a script that names the *wrong feature* — precisely the failure positional addressing would have produced.

Checked non-vacuous: the authored fixture draws **3** annotations where the planner draws **7**, so the original `auto_dimensions()` bug fails it loudly.

## One test retargeted, not deleted

`test_a_model_with_an_authored_set_is_refused_not_silently_converted` pinned the blanket refusal. Its **property** — the authored set must never become the planner's — still holds and still matters; only its expectation changed, from "refuses" to "emits the declarations". The docstring carries the whole history so the next reader sees why it moved.

**2162 passed, 1 skipped** (full fast tier), all four lint checks clean.


pzfreo:
Review of `54432d4`: the ordinary authored-set round trip is correctly identity-based, but there is one P1 source-boundary case.

**P1 — an explicitly empty authored set emits a script that cannot build.**

ADR 0016 deliberately distinguishes `authored_dimensions=None` (“planner chooses”) from `authored_dimensions=()` (“the author chose no dimensions”). The direct public `PartModel`/`build_drawing` path accepts the latter and produces a drawing with no generated dimensions. `_dimension_block()` also recognizes it as authored (`is not None`) and emits the “THIS IS THE COMPLETE SET” header with zero declarations.

However, the generated `Sheet` never enters authored mode: the only public transition is executing at least one `sheet.dimension(...)` line. With no lines, both `sheet._authored` and `_auto_dimensions` remain empty/`None`, so `sheet.build()` raises the mandatory-source error.

Minimal reproduction:

```python
import dataclasses
from build123d import Box
from draftwright.builder import detect_part_model
from draftwright.sheet_emit import emit_sheet_script

part = Box(40, 20, 10)
model = dataclasses.replace(detect_part_model(part), authored_dimensions=())

direct = build_drawing(part, model=model, authored=model.authored_dimensions)
assert not [n for n, _ in direct.iter_annotations() if n.startswith(("m_", "dim_", "hc_"))]

src = emit_sheet_script(model, "part", "empty", title="T", number="N")
ns = {"part": part}
body = src.replace("\npart\n", "\n", 1)
body = body[:body.index("sheet.export(")]
exec(compile(body, "<emit>", "exec"), ns)
ns["sheet"].build()
# ValueError: this sheet does not say where its dimensions come from
```

This is a generated script diverging from a valid source model—the exact #707 class this PR closes. Please add an end-to-end empty-authored-set canary.

The structural fix needs an explicit way for a script to state “the complete authored set is empty” (for example a source-selection verb), because absence of `dimension(...)` lines cannot carry that meaning under the mandatory-source rule. If adding that API is outside this slice, retain a narrowed refusal for `authored_dimensions == ()`; emitting a script known not to build is worse than refusing it. Do not map it to `auto_dimensions()`, which would recreate the original widening bug.

The non-empty authored path, feature-name map, discriminator-to-`axis=` spelling, and refusal for dimensions on unemittable features otherwise look sound. CI is still running; lint is green.

pzfreo:
## P1 fixed — and it was an API gap, not an emitter bug (`HEAD` updated)

Reproduced exactly: the direct path builds a drawing with zero dimensions, the generated script raises `this sheet does not say where its dimensions come from`.

### The real finding

`auto_dimensions()` is an **explicit verb**. The authored source was entered **implicitly** — as a side effect of calling `dimension(...)` at least once. So one thing was unsayable: *"the complete set is empty."* Absence of `dimension(...)` lines is also what a script with no source at all looks like, which `_check_dimension_source` must reject.

That is the same ambiguity #874 removed at build level, left standing at the entry. A valid `PartModel` could be built but not written.

```python
sheet.authored_dimensions()      # the missing half of the pair
```

`dimension(...)` still selects the source on its own, so nothing existing changes.

### The emitter now writes it unconditionally

Not only for the empty case — which fixes a compromise in the previous commit I should have flagged at the time: an authored script stated its source in a **comment**. The automatic path states it with a verb, and a reader should not have to trust prose to know whether an absent dimension means anything. Both paths are now machine-checkable.

### On the fallback

I did not take it. Narrowing the refusal to `authored_dimensions == ()` would have made that refusal **permanent** — there would be no API to lift it with, and the gap would have sat behind a "not yet" that nothing could resolve.

### The state-field ratchet then caught me

`_authored_source` was unclassified, and #912's guard failed — correctly. Every field a feature reference could live in must be declared as carrying one or not. It does not (the set's *members* live in `_authored`, already classified and scenario'd), so it is classified beside `_auto_dimensions` for the same reason.

Both cases round-trip end to end now: empty authored set builds with zero dimensions, `None` builds with the planner's three. Canaried — removing the emitted verb fails both new tests.

**2164 passed, 1 skipped** (full fast tier), all four lint checks clean.

pzfreo:
Re-review of `6d574ae`: the P1 is fixed structurally. The empty authored set now round-trips, mixed sources still refuse, and `Sheet.from_part(...).authored_dimensions()` correctly takes over the implicit automatic source. One small P2 remains in the public contract text.

**P2 — the mandatory-source error still says an authored source can only be selected with `dimension(...)` lines.**

The new verb exists specifically because that advice is impossible to follow for an empty complete set. Yet a source-less build still raises:

```text
Call auto_dimensions() for the planner's set, or declare the complete set with
 dimension(feature, role) lines.
```

That was accurate before this fix; it is now the stale API guidance that hides the only way to express “author chose no dimensions.” Please update it to name both source verbs, e.g. call `authored_dimensions()` for the complete authored set and then add zero or more `dimension(feature, role)` lines. Add an assertion on the error text so this new entry point remains discoverable.

There is also a directly contradictory `_authored_set()` docstring at `sheet.py:1478`: it still says an empty tuple is something `_check_dimension_source` rejects, immediately above the new code that intentionally returns `()` when `_authored_source` is true. Please correct that sentence in the same pass.

These are documentation/error-contract findings, not a functional blocker. The behavior now checks out:

- empty authored set → `PartModel.authored_dimensions == ()` and zero generated dimensions;
- explicit auto + authored, in either order → visible conflict;
- authored + `add_dimension` → visible conflict;
- `from_part()` + empty authored takeover → `()` and automatic source cleared;
- non-empty authored and automatic emitter paths remain distinct.

Local focused result: **349 passed, 1 skipped** (`test_sheet_emit.py`, `test_add_dimension.py`, and `test_sheet_identity_invariant.py`). `git diff --check` is clean. CI is still running; lint is green.

pzfreo:
## Both P2s fixed (`HEAD` updated)

Both mine, and the second worse than "stale".

### The error text

It said *"declare the complete set with `dimension(feature, role)` lines"* — advice impossible to follow for the empty set, which is the entire reason the verb exists. **Stale error text about an API is worse than none**: it reads as an exhaustive list, and this one hid the only route to the case a caller would be stuck on.

Now names both verbs and says why the verb exists when a line would usually do. Asserted in a test, per your request, so it stays discoverable from the failure a caller actually hits rather than from a review.

### The `_authored_set()` docstring

Worse than stale — I **contradicted it with code directly beneath it**. It still said an empty tuple is "what `_check_dimension_source` rejects before we get here", sitting immediately above the `return () if self._authored_source else None` I had just added. I changed the code and left the paragraph explaining why the code couldn't do that. It now states what is true, why it changed, and what *is* still rejected (naming no source at all).

### One thing you didn't ask for

ADR 0016's two-sources section showed only the two old spellings, so it was about to become the third place describing the pre-verb contract. A design document that disagrees with the API is how the next person re-derives a constraint that was lifted — the same failure I spent this PR series fixing in #883. Added the verb and the empty-set example there too.

**2165 passed, 1 skipped** (full fast tier), all four lint checks clean.

pzfreo:
Re-review of `892e906`: **clean — no further findings.**

The remaining P2s are fixed at the right contract surfaces:

- the mandatory-source error now names both `auto_dimensions()` and `authored_dimensions()`, explains that zero or more `dimension(...)` lines follow the latter, and identifies why the explicit verb is needed for an empty set;
- the error behavior is pinned so the empty-authored-set route remains discoverable;
- `_authored_set()` now documents the actual load-bearing distinction: `None` means planner source, `()` means a legitimate complete-but-empty authored source, while only “no source selected” is rejected;
- ADR 0016 now records both source verbs and an executable empty-set example, so the design document no longer describes the pre-verb API.

I reran the source-selection matrix and the broader focused suites. Empty authored sets round-trip with zero generated dimensions; non-empty authored sets retain only their declarations; automatic sets remain automatic; mixed automatic/authored and authored/augmenting sources fail visibly; `from_part()` takeover behaves correctly.

Local result: **350 passed, 1 skipped** (`test_sheet_emit.py`, `test_add_dimension.py`, and `test_sheet_identity_invariant.py`). `git diff --check` is clean. GitHub is fully green across lint, every platform/Python job, and both Codecov checks; merge state is `CLEAN`.

From this review, PR #933 is ready to merge.

Issue #927, opened 2026-07-30
Draw down the label-provenance budget (26 sites where a renderer formats a number)


pzfreo:
Follow-up to #925, which added `tests/test_label_provenance.py`.

## The invariant

A renderer calling `_fmt(x)` is turning a number into printed text, which means the number reached it **as a number** rather than as the compiler's `value_text`. Every defect of this class in #921/#923/#925 looked exactly like that: the height ladder rebuilding rung labels from `StepLevelFeature.levels`, `render_slots` formatting `s.width` past a suppression flag, `_add_furniture` printing `feat.pitch`, `_locate_off_axis_holes` recomputing `abs(loc[i] - bb.min[i])` beside a compiler that had already computed it.

The ratchet is shrink-only and 26 sites remain. **A budget nobody is drawing down becomes furniture**, so this issue is the drawdown, in three independent pieces.

## 1. The ø-leader item tuple (6 sites)

`_diameter_row_below` / `_diameter_column_left` take `(anchor, dia, feature, tolerance, thread)` where `dia` is a float that is **both printed and used for tip geometry** (`az - dia / 2`). Carrying the approved text alongside the number is a five-call-site change (`render_diameters`, `render_boss_diameters`, `add_feature_diameter`, plus the two placers).

Sites: `from_model._diameter_row_below` (2), `_diameter_column_left` (2), `render_diameters` (1), `render_boss_diameters` (1).

## 2. The Z-normal location ladder's per-axis values (6 sites)

The one location family compiled without per-axis values. `render_locations` groups refs **across features** and dedups per axis *before* it knows which dims exist, so an entry per axis would approve a mark whose existence the renderer decides. Splitting it means moving that grouping into the compiler.

Every other family is already per-axis: off-axis holes, non-Z pockets and slot positions each carry one approved entry per measured coordinate with its own value and span.

Sites: `from_model.render_locations` (4), `holes.add_feature_location` (2). Related to #714 (dedup the mirrored X/Y render blocks) — the two are worth doing together, since the grouping move and the mirror removal touch the same code.

## 3. The turned step-chain repeat-run collapse (8 sites)

`render_step_lengths` composes labels while collapsing repeat runs, so the text is decided by the collapse rather than per approved entry. The collapse is a *content* decision and belongs in the compiler, exactly as `_step_repeat` moved there in #923 for the height ladder.

[...]

PR #934, opened 2026-07-30
fix(#889): give the overall height an intent, so a generated script can replay it

### LINKED ISSUE #889: Generated script replay omits the automatic overall height dimension
## Problem

A drawing generated with `generate_script()` does not fully reproduce the automatic drawing for the Y-axis stepped-flange case used by `TestTurnedDiameters._issue_881_y_step_flange()`.

The automatic drawing contains `dim_height`, while executing the generated detect-only script does not. Both drawings currently lint cleanly, so lint does not expose the parity gap.

This was observed while reviewing PR #888 and is separate from its pad/pocket coverage work.

## Reproduction

1. Build the Y-axis stepped-flange solid from `tests/test_make_drawing.py::TestTurnedDiameters._issue_881_y_step_flange`.
2. Create the automatic drawing with `build_drawing(part)`.
3. Export the part to STEP, call `generate_script()`, and execute the generated script.
4. Compare annotation names.

Current difference:

```text
automatic only: dim_height
generated-script only: none
```

Both `auto.lint_summary()["by_code"]` and `replayed.lint_summary()["by_code"]` are empty.

## Expected

The generated intent-level script should reproduce the automatic overall height furniture, or explicitly document and diagnose that the dimension is intentionally outside replay fidelity.

## Acceptance criteria

- Add a regression test comparing the relevant automatic and generated-script annotation semantics for the Y-axis stepped-flange case.
- Replay includes the overall height dimension without duplicating the Y-step length chain.
- Automatic and replayed drawings remain lint-clean across supported platforms.

pzfreo:
Closes #889.

## What was happening

The Y-axis stepped flange replayed without its 46 mm overall height — silently, and lint-clean on both drawings, which is why lint never surfaced it.

```
auto only     : ['dim_height']
generated only: []
```

## Root cause

**The overall height is the one dimension with no feature to name.**

A model with an `EnvelopeFeature` carries a `height` parameter, and `generate_script` already emits `dimension(f, "length", role="height")` for it. This part has **no envelope feature at all** — the height comes from the compiler's bounding-box fallback (`_compile_overall_height`), a decision only the compiler may make. So an intent-level script had no way to say *"and the 46 mm overall height"*. It could not be replayed, only lost.

`Drawing.overall_height()` is the verb for exactly that case. The emitter writes it only when the model has no envelope feature, so it never duplicates the existing height line.

## The first fix was wrong, and the existing suite caught it

I made the drain draw the overall height whenever the compiler approved one. That restored parity — and broke `test_finalize_replay_equals_live_placement`.

Under `auto_dims=False` the recorded verbs **are** the drawing, so injecting an automatic dimension nobody asked for meant record-then-finalize no longer equalled placing live. **That is this bug's mirror image**: I fixed a silent omission by creating a silent addition.

As an intent it is absent unless recorded — which preserves the live/deferred contract *and* makes it commentable, the property the whole intent-level script rests on (ADR 0016, "the script records intent").

I would have shipped the wrong fix on the targeted tests alone; the full tier is what caught it.

## Tests

- **Full annotation-set parity**, not "`dim_height` is present". The issue's acceptance has two halves — replay includes the height *and* does not duplicate the Y-step length chain — and a presence check only tests one. Canaried on `main`: `replay differs — missing ['dim_height'], extra []`.
- **Commentability / live-vs-deferred**: not recorded ⇒ not drawn; live == deferred. This is the assertion the first fix would have failed.
- The fixture asserts it has *no* envelope feature, so the test cannot silently drift onto the already-working path.

**2167 passed, 1 skipped** (full fast tier), all four lint checks clean.

### DIFF
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -2822,6 +2822,33 @@ def render_step_lengths(dwg, plan, *, ctx, only=None) -> int:
     return _draw_step_chain(dwg, view, fsegs, "m_steplen", ctx=ctx, start=start)
 
 
+def ladder_plan_for(plan, *, step_height: bool, overall: bool):
+    """*plan* narrowed to the ladders a caller actually asked :func:`render_height_ladder` for.
+
+    The renderer draws TWO independent things — a `step_level` feature's correlated rungs and
+    the envelope/bbox overall height — and reads a third, `step_position`, only for its
+    PRESENCE (short-rung placement). Handing it a plan containing more than was asked for
+    draws more than was asked for: the #889 drain passed the whole compiled plan once either
+    intent was recorded, so `overall_height()` alone also rebuilt the step rungs — a
+    dimension nobody recorded, and live/deferred divergence in the one PR relying on their
+    equivalence (#934 review).
+
+    Exists so the live verb and the finalize drain project the plan the SAME way. Two
+    spellings of "which ladders did they ask for" is how they diverged in the first place.
+
+    `step_position` rides `step_height`: it is not content here, it is how those rungs are
+    placed, so it is meaningless without them.
+    """
+    from dataclasses import replace
+
+    kinds = []
+    if step_height:
+        kinds += ["step_height", "step_position"]
+    if overall:
+        kinds.append("overall_height")
+    return replace(plan, ladders=tuple(lad for lad in plan.ladders if lad.kind in kinds))
+
+
 def render_height_ladder(dwg, plan, frame, *, ctx, detail_view: bool = False) -> int:
     """Front-view right ladder: prismatic step heights stacked inner→outer, then the overall
     height outermost — registered as :class:`CorridorCandidate`s in the shared
--- a/src/draftwright/builder.py
+++ b/src/draftwright/builder.py
@@ -1054,6 +1054,21 @@ def _feature_listing(a: Analysis) -> str:
         for p in feat.parameters():
             if p.span is not None or kind in ("slot", "pad"):
                 body.append(f'dwg.dimension(f, "{p.kind}", role="{p.role}")   # {display(p)}')
+    # The overall height, when it comes from the compiler's BOUNDING-BOX fallback rather
+    # than from an `EnvelopeFeature`. An enveloped model already emitted
+    # `dimension(f, "length", role="height")` in the loop above; a model without one has no
+    # feature to name, so the replay simply lost the dimension — silently and lint-clean
+    # (#889). `overall_height()` is the verb for exactly that case.
+    if not any(f.kind == "envelope" for f in feats):
+        from draftwright.model.compiled import compile_dimensions
+
+        if compile_dimensions(model).ladder("overall_height") is not None:
+            body += [
+                "",
+                "# Overall height (from the bounding box — this part declares no envelope",
+                "# feature, so there is none to hang a dimension(...) intent on).",
+                "dwg.overall_height()",
+            ]
     if plan_sections(model, feature_hole_keys(model, a)) is not None:
         body += [
             "",
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -192,6 +192,9 @@ class _IntentRouting:
     len_ids: set
     slot_ids: set
     height_ladder_ids: set
+    #: `Drawing.overall_height()` intents — the bbox-fallback overall height, which has no
+    #: feature to hang a `dimension(...)` intent on (#889).
+    overall_height_ids: set
     step_position_ids: set
     explicit_envelope_height: bool
     user_dim_ids: set
@@ -1212,6 +1215,53 @@ class Drawing:
             self, feature, self._part_model, self._analysis, view=view, name=name, ctx=ctx
         )
 
+    def overall_height(self) -> list[str]:
+        """Add the part's **overall height** — the one dimension with no feature to name.
+
+        Every other add verb takes a feature, because every other dimension belongs to one.
+        The overall height usually does too: a model with an `EnvelopeFeature` carries a
+        `height` parameter, and `dimension(env, "length", role="height")` is the verb for it.
+
+        A model WITHOUT one still gets an overall height — the compiler falls back to the
+        bounding box, which is a decision only the compiler may make (`_compile_overall_height`).
+        There is then no feature to record an intent against, so an intent-level script had no
+        way to say "and the 46 mm overall height", and a generated script replayed without it,
+        silently and lint-clean (#889).
+
+        This verb is that line. It is deliberately NOT "draw it whenever the compiler approves
+        one": `auto_dims=False` means the verbs are the whole drawing, so a dimension nobody
+        recorded must not appear — record-then-finalize has to equal placing live.
+
+        Returns the placed names (empty when the compiler withholds the height — a Z-turned
+        part whose step chain already tiles it, or an X/Y rotational OD that conveys it).
+        """
+        if self._defer_intents:  # #426: record, don't place — finalize() drains it
+            self._intents.append(Intent("overall_height", None, {}))
+            return []
+        model, a = self._part_model, self._analysis
+        if model is None or a is None:
+            raise ValueError("overall_height(): no detected model — build the drawing first")
+        from draftwright._core import layout_frame
+        from draftwright.annotations._common import drain_corridors
+        from draftwright.annotations.from_model import render_height_ladder
+        from draftwright.model.compiled import RenderableDimensionPlan, compile_dimensions
+
+        before = set(self.annotations())
+        ctx = PlacementContext(registry=self._registry, coverage=self._coverage, items=self.items)
+        overall = compile_dimensions(model).ladder("overall_height")
+        if overall is not None:
+            # ONLY the overall height: passing the whole plan would also rebuild the step
+            # ladder, which is a different intent with its own verb.
+            render_height_ladder(
+                self,
+                RenderableDimensionPlan(ladders=(overall,)),
+                layout_frame(a),
+                ctx=ctx,
+                detail_view=self._build.detail_view,
+            )
+            drain_corridors(ctx, self)
+        return sorted(set(self.annotations()) - before)
+
     def furniture(self, feature, *, view=None) -> list[str]:
         """Add a hole/pattern's non-dimensional **sheet furniture** (#419) — centre marks
         (every member) plus a pattern's centre-cross (bolt circle) or pitch/grid dims.
@@ -1448,6 +1498,7 @@ class Drawing:
             and it.kwargs.get("param") == "length"
             and it.kwargs.get("role") in (None, "step_height")
         }
+        overall_height_ids = {id(it) for it in self._intents if it.kind == "overall_height"}
         explicit_envelope_height = any(
             routable
             and it.kind == "dimension"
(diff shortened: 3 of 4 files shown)


pzfreo:
Adversarial review of `3febc90`: the Y-flange replay is fixed, but there is one P1 in the new shared drain.

**P1 — deferred `overall_height()` also draws an unrecorded step-height ladder.**

The live verb correctly constructs a `RenderableDimensionPlan` containing only the `overall_height` ladder. The deferred `_s_height_ladder()` path does not: once either `height_ladder_ids` **or** `overall_height_ids` is present, it passes the entire result of `compile_dimensions(model, ...)` to `render_height_ladder()`. That plan can contain both `step_height` and `overall_height`, so an overall-height-only intent draws the step rungs too.

Minimal reproduction:

```python
from build123d import Box
from draftwright.model.ir import Frame, PartModel, StepLevelFeature

part = Box(80, 60, 30)
step = StepLevelFeature(
    Frame((0, 0, 0), "z"), base=-15, levels=(-5, 5),
    shoulders=(("x", 0),), datum=(-40, -30, -15),
)
model = PartModel(
    bbox=part.bounding_box(), orientation="prismatic", features=[step], datums=[]
)

live = build_drawing(part, model=model, auto_dims=False)
live.overall_height()
assert {n for n, _ in live.iter_annotations() if n.startswith("dim_")} == {"dim_height"}

deferred = build_drawing(part, model=model, auto_dims=False)
with deferred.deferred():
    deferred.overall_height()
assert {n for n, _ in deferred.iter_annotations() if n.startswith("dim_")} == {"dim_height"}
# Fails: deferred also contains dim_step_0.
```

Observed output:

```text
live     [('dim_height', '30')]
deferred [('dim_height', '30'), ('dim_step_0', '10')]
```

This violates the live/deferred equivalence the PR explicitly relies on and silently adds a dimension nobody recorded. It also breaks the claimed commentability in generated scripts for a no-envelope model that has `step_level`: `_feature_listing()` emits both the step-height intent and `dwg.overall_height()`, but commenting only the step-height line still lets the overall-height drain rebuild the rungs; conversely, the existing step-height drain currently carries the overall ladder, so commenting only `dwg.overall_height()` need not remove the overall height.

The shared renderer is fine; the plan passed to it must be projected to the recorded intents. Compile once, then include `step_height` only when `r.height_ladder_ids` is nonempty and `overall_height` only when `r.overall_height_ids` is nonempty (plus only the structural `step_position` presence actually recorded/needed for short-rung placement). The deferred route should mirror the live verb’s partial-plan construction rather than using “either intent means all ladders.”

Please add a model with both approved ladders and assert all four intent states: neither, overall-only, step-only, and both; for each, live and deferred annotation sets must agree. The current fixtures each expose only one ladder, so they cannot catch cross-contamination.

`git diff --check` is clean and CI is currently running with lint green.

pzfreo:
## P1 fixed (`HEAD` updated)

Reproduced exactly. And you were right about the converse as well — I checked the direction you predicted but didn't demonstrate, and it fails too.

### The fix

`ladder_plan_for(plan, step_height=…, overall=…)`, used by **both** routes — the live verb and the drain. The live verb was already building a partial plan by hand; the drain built a different one. **Two spellings of "which ladders did they ask for" is how they diverged**, which is the defect shape this whole series keeps hitting, so the projection is now one function rather than a corrected copy.

`step_position` rides `step_height`: the renderer reads only its *presence*, for short-rung placement, so it is meaningless without the rungs.

### One existing test asserted the conflation

`test_deferred_dimension_rebuilds_prismatic_step_height_ladder` asserted `dim_height` appears from a step-height intent. Updated — but **checked rather than assumed** that no replay regression rides on it:

- a part **with** an `EnvelopeFeature` emits `dimension(f, "length", role="height")`, which sets `explicit_envelope_height` and takes the overall height out of the compile entirely — so the ladder never carried it there anyway;
- verified against `main` that the one such fixture diverges *identically* before and after this change;
- a part **without** one now emits `dwg.overall_height()`.

So the conflation only ever changed the hand-driven case that test drove.

### Coverage, as requested

Four intent states on a model with **both** approved ladders — neither / overall-only / step-only / both. Your diagnosis of why this was invisible is exactly right: every earlier fixture exposes one ladder, and with a single ladder "pass everything" and "pass what was asked for" are the same plan.

Asserted on **which** ladders appear rather than how many marks — with both recorded the strip legitimately drops a rung for room, and pinning counts would fail on placement changes unrelated to the property.

Canaried: three of the new tests fail on the previous head.

### Separate, pre-existing, not from this PR

While checking the above I found that an enveloped part's **height/width/depth replay as generic `dim_length0/1/2`** instead of `dim_height`/`m_env_width`/`m_env_depth`. Identical on `main`, so it is untouched by this change — but it is a real replay-fidelity gap of the same family as #889. Say the word and I'll file it.

Lint, smoke and the targeted suites (`TestFeatureEdits`, `TestTurnedDiameters`: **115 passed**) are green; CI has the full matrix.

pzfreo:
Filed as #935 — envelope dimensions replay as generic `dim_length*` instead of `dim_height` / `m_env_width` / `m_env_depth`.

Characterised rather than just reported: the emitter is **correct** (three properly-roled intents, right values). What is lost is that they replay through the generic user-dimension path instead of `render_envelope` / the overall-height ladder — so it costs identity (`dim_length1` says nothing about which measurement it is, and its index depends on emit order) and placement (the envelope pass has its own strip priority; the generic path competes as an ordinary corridor candidate).

It interacts with this PR in one specific way worth recording: an explicit `role="height"` intent sets `explicit_envelope_height`, which deliberately removes the overall height from the compile so the intent path can draw it — and the intent path then draws a `dim_length*`. The ladder's `dim_height` is correctly suppressed and never replaced by an equivalent.

Suggested direction is the same shape as this PR's fix: route a recorded envelope dimension back through the pass that owns that content rather than flattening it into a generic candidate. `render_envelope` already consumes the compiled plan (#925), so `ladder_plan_for` is the pattern.

Verified pre-existing — `main` and this branch diverge identically.

pzfreo:
Re-review of `bfc7881`: the previous cross-contamination is fixed on the four requested ladder states. I found one adjacent P1 in the new public verb’s supported domain.

**P1 — `overall_height()` plus an envelope height is order-dependent live and disagrees with deferred replay.**

`overall_height()` is documented as the verb for the case with **no** `EnvelopeFeature`, where there is no feature to name. But it does not enforce that precondition. On an enveloped model it can be combined with the sanctioned `dimension(env, "length", role="height")` path. If `overall_height()` is placed first live, the later feature dimension produces a duplicate; the deferred route detects `explicit_envelope_height`, removes the overall ladder from the compile, and produces only the feature dimension.

Minimal reproduction:

```python
from build123d import Box
from draftwright.model.declare import envelope

part = Box(80, 60, 30)

def drawing(deferred):
    env = envelope(part)
    dwg = build_drawing(part, model=[env], auto_dims=False)
    feature = dwg.model().features[0]
    if deferred:
        with dwg.deferred():
            dwg.overall_height()
            dwg.dimension(feature, "length", role="height")
    else:
        dwg.overall_height()
        dwg.dimension(feature, "length", role="height")
    return [(n, a.label) for n, a in dwg.iter_annotations()
            if "height" in n or n.startswith("dim_length")]

print(drawing(False))
# [('dim_height', '30'), ('dim_length0', '30')]
print(drawing(True))
# [('dim_length0', '30')]
```

Reversing the two live calls happens to avoid the duplicate because `overall_height()` then places nothing, so the live API is order-dependent as well as inconsistent with deferred replay.

This is exactly the “silent addition” mirror image described in the PR, now reachable by composing the two public spellings of the same measurement. Since the new verb exists specifically for the featureless fallback, the cleanest contract is to reject `overall_height()` when the model contains an `EnvelopeFeature`, with an error directing the caller to `dimension(envelope, "length", role="height")`. Apply the same check before recording in deferred mode so both routes fail identically. Alternatively, if both spellings are intentionally supported, they need one semantic identity/dedup rule that is order-independent in live and deferred paths.

Please add envelope-present tests for overall-only and both call orders, live and deferred. The current four-state fixture deliberately has no envelope, so it proves plan projection but cannot exercise the overlap that `explicit_envelope_height` exists to arbitrate.

The previous fix itself is sound: overall-only, step-only, both, and neither now isolate correctly on a model carrying both compiled ladders. CI is still running; lint and Ubuntu 3.14 are green.

pzfreo:
## P1 fixed (`HEAD` updated)

Reproduced exactly, including the order-dependence:

```
live  fwd: [('dim_height', '30'), ('dim_length0', '30')]     ← duplicated
live  rev: [('dim_length0', '30')]
defer fwd: [('dim_length0', '30')]
```

### The fix: refuse it

Took your first option. `overall_height()` exists **only** for the featureless fallback — that is the entire justification for adding public API rather than reusing the feature verb. Supporting both spellings would mean two ways to say one measurement, which is the problem #906 already tracks and the one this whole series keeps paying for. So it raises on an enveloped model and names the verb to use instead.

**Refused before the deferred/live split**, so both routes answer identically — the shape #925 settled for `callout()`, where a check on one side made the answer depend on whether you were inside `deferred()`.

The emitter is unaffected: it already writes the verb only when the model has no envelope feature.

### One of my own tests was standing on the overlap

`test_the_overall_height_intent_is_commentable_not_injected` asserted the commentability property on an *enveloped* plate — a part where the verb should never have been usable. It now uses the Y-flange, the verb's actual domain, and asserts the fixture has no envelope feature so it cannot drift back onto the enveloped path.

That is the third fixture in this PR that quietly tested the easy case; your point about the four-state fixture needing *both* ladders was the same observation one level down.

### Tests, as requested

- envelope present × {live, deferred} → refuses, identically;
- envelope present, feature verb alone × {live, deferred} → height drawn **once** (the false-positive half: refusing the second spelling must not cost the first).

Lint, smoke, and the targeted suites (`TestFeatureEdits`, `TestTurnedDiameters`: **119 passed**) are green; CI has the full matrix.

---

Separately, per your instruction: filed **#936** for the chamfer-angle IR gap (printed, but no `DimParameter`, so unsuppressible and untolerable — the one permanent entry in #927's budget justified by an IR gap rather than by not being a dimension). The ADR deletion-discipline rule is going out as its own PR.

pzfreo:
Re-review of `e301131`: **clean — no further findings.**

The previous overlap is fixed at the correct boundary:

- `overall_height()` now enforces its actual domain—models with no `EnvelopeFeature`, where the compiler’s bbox fallback has no feature to name;
- the guard runs before the deferred/live split, so both routes reject an enveloped model identically and no stale intent is recorded;
- the error directs the caller to the single sanctioned spelling, `dimension(envelope, "length", role="height")`;
- the feature-owned envelope height still draws exactly once in both live and deferred paths;
- generated scripts are unaffected because `_feature_listing()` already emits `overall_height()` only when no envelope exists.

I rechecked the earlier partial-plan correction as well. On a model with both compiled ladders, neither / overall-only / step-only / both now expose exactly the recorded ladder families, and live `overall_height()` matches deferred replay. The helper shared by both routes removes the duplicate derivation that caused the original divergence.

Local targeted result: **12 passed** (`overall_height`, ladder-intent, and prismatic-step-height cases); `git diff --check` is clean. GitHub lint and Ubuntu 3.14 are green; the remaining platform jobs are still running with no failures.

The separately discovered envelope replay-identity gap is correctly outside this PR and tracked as #935. From this review, PR #934 is ready once the remaining CI jobs finish green.

Issue #928, opened 2026-07-30
Exhaustive authored-role property test: every addressable role, drawn or omitted


pzfreo:
Follow-up to #925. The third of three prevention mechanisms discussed on that PR; the other two landed (`tests/test_compiled_plan_boundary.py`, `tests/test_label_provenance.py`).

## Why

Every defect in the #925 review series shared one shape: **a rule stated in one place and re-derived in another, agreeing on the auto path and disagreeing under an authored set.**

- `location_role` said a hole is locatable; `plan_locations` said only a Z-normal one is; `_locate_off_axis_holes` drew the X/Y ones from raw IR anyway.
- Then the same table said a *pattern* is locatable while neither compiler emits one off-axis — so `dimension(x_pattern, "location")` was accepted and silently drew nothing.
- `bolt_circle.diameter` was classified by *where it appears in the string* rather than by whether it has a `DimParameter`.

The existing behavioural guard is fixture-driven, and **a behavioural test only covers the paths its fixtures reach**. `_staircase()` has no holes, which is exactly how two whole dimensional paths stayed outside the boundary for a review round each. Adding a fixture per discovered gap is chasing instances.

## Scope

For a corpus of parts covering every IR feature kind in every principal orientation, **enumerate features × addressable roles automatically** and assert, for each:

1. authoring exactly that role produces a mark attributable to it;
2. authoring the complement produces none;
3. a role the vocabulary rejects raises rather than building blank.

Generated, not hand-listed — a hand-maintained expectation table is how `_PENDING_VALUE_CARRYING` came to be treated as the definition of "dimensional". The enumeration should come from `feature.parameters()` plus `location_role()`, i.e. from the same source the authored vocabulary uses, so a new feature kind joins the sweep by existing rather than by someone remembering.

## Notes

- Costs real CI time; likely `-m slow` (post-merge on `main`) rather than a PR gate, per the #153 tiering.
- Expect it to find gaps on first run. That is the point; file them as children.
- Where a kind legitimately has no addressable role (aspect features — `control_frame`, `datum_ref`, `finish` — carry author-supplied text with empty `parameters()`), the sweep should assert that explicitly rather than skipping, so "known empty" stays distinct from "never checked".

Issue #926, opened 2026-07-30
Hole callouts (hc_) still take the legacy DimensionGroup surface


pzfreo:
Child of #867. The last renderer outside the ADR 0016 compiled-plan boundary, named in the ADR and pinned by `tests/test_compiled_plan_boundary.py::_PENDING_VALUE_CARRYING`.

## What is true today

`_annotate_holes` → `_assemble_view_callouts` → `hole_callout_spec(group)` consumes `DimensionGroup`/`PlannedDimension`, where `suppressed` is an advisory boolean. It *is* honoured — `model/callout.py` checks it at every term, head and dependent — so **this is a structural gap, not a behavioural one**. No known authored omission leaks through it.

That is exactly why it needs an issue rather than a shrug. "The renderer checks the flag" is the guarantee the boundary exists to replace: eight adversarial rounds on #921 found eight renderers that were supposed to check and did not, and fixing them one at a time produced four mechanisms for saying the same thing. The ninth renderer is not different in kind because it currently happens to be correct.

## Scope

- `hole_callout_spec` consumes `ApprovedGroup` — approved entries only, no `suppressed` field to honour or forget.
- The compound-dependency rules (`_refuse_headless_callout`, `_recess`, `_shadowed`) become simpler, not harder: `if pd is None or pd.suppressed` collapses to `if pd is None`. Keep the dependent/rider split #925 established — a term with a `DimParameter` is a dependent and is never waived; a rider (thread spec, `n×` multiplier, grid `(3×3)`) lives on the feature and is waived only for an *authored* omission.
- The spec dict carries approved TEXT rather than floats, which also retires `callout_from_spec` / its `f()` from the label-provenance budget (the #261 "every value crosses as a `_fmt` string" invariant moves into the compiler, where it belongs).
- Delete `hc_` from `_PENDING_VALUE_CARRYING`; the guard then fails if it reappears.

## Also: split the tuple, because it conflates two categories

`_PENDING_VALUE_CARRYING` currently holds `("hc_", "pmi_")` — one pending migration and one **permanent** exception, in a list named "pending" and described as shrink-only. That is why this issue's original "done when" was the awkward `("pmi_",)`: a pending list whose finished state is non-empty.

Raised by the #937 review, which caught ADR 0016 citing the whole tuple as debt that "can only ever get smaller". The ADR now names the two members separately; the code should too:

```python
_PENDING_MIGRATION = ("hc_",)        # shrink-only; empty when this issue closes
[...]

PR #937, opened 2026-07-30
docs(ADR 0016): record the deletion discipline — the rule that actually held

pzfreo:
Docs only, no behaviour change.

## Why

During the #925 review you asked how to prevent the recurring defect shape. I named three mechanisms. Two shipped — the label-provenance ratchet (#925) and the exhaustive authored-role sweep (#928). The third I described as *"costs nothing, write it into ADR 0016 as a one-line rule"* — and then never did. A tracker audit caught it.

It is the mechanism that most reliably held, and the one that explains why some fixes stuck and others did not.

## The rule

**Designating an owner means DELETING the alternative reader, not redirecting the caller.**

| | |
|---|---|
| #921 | fixed eight renderers to check `suppressed` — **the ninth leaked** |
| #923 | **deleted the field** — none has leaked since; `ApprovedDimension` has nothing to forget |
| #925 | removed `_ir_off_axis_holes` rather than leaving it beside `_approved_off_axis_holes` |
| #933 | removed the emitter's blanket refusal rather than adding a flag beside it |

A redirect leaves the trap armed: the next author reaches for the reader that still exists, *because it still exists*. Fixing N call sites is O(N) work repeated for every new call site; removing the thing they call is O(1) and permanent.

And when it cannot yet be removed, it is **named as pending** — `_PENDING_VALUE_CARRYING`, `_FMT_BUDGET`, `_SAME_PATH_AS_ENVELOPE` — rather than left to be rediscovered.

## The corollary

Recorded because it otherwise reads as a stylistic habit: this ADR keeps producing **tables rather than checks**, because a check must be remembered at each site while a table is consulted from one. `location_datum`, `_FACTS` and `_LOCATION_ROLE` are each the deletion of a rule that had been restated in three places.

Sits beside the existing "two ratchets, on the symptom and on the cause" paragraph, so the three prevention mechanisms are documented together.

### DIFF
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -153,9 +153,9 @@ reach (`_staircase()` has no holes, which is how two dimensional paths stayed ou
 boundary unnoticed); the source ratchet covers every path whether a fixture walks it or not.
 It is shrink-only, and each survivor carries a written reason.
 
-**Designating an owner means DELETING the alternative reader, not redirecting the caller.**
-The rule that most reliably held while this ADR was implemented, and the one that explains
-why some fixes stuck and others did not:
+**Once a semantic owner is designated, parallel INFERENCE paths must be deleted, not
+redirected.** The rule that most reliably held while this ADR was implemented, and the one
+that explains why some fixes stuck and others did not:
 
 - #921 fixed eight renderers to check `suppressed`. The ninth leaked.
 - #923 **deleted the field**. No renderer has leaked it since — `ApprovedDimension` has
@@ -165,10 +165,25 @@ why some fixes stuck and others did not:
 
 A redirect leaves the trap armed: the next author reaches for the reader that still exists,
 because it still exists. Fixing N call sites is O(N) work that must be repeated for every
-new call site; removing the thing they call is O(1) and permanent. So when this ADR names a
-single source of truth, the same change removes the other way to get the same answer —
-and if it cannot yet be removed, it is NAMED as pending (`_PENDING_VALUE_CARRYING`,
-`_FMT_BUDGET`, `_SAME_PATH_AS_ENVELOPE`) rather than left to be discovered.
+new call site; deleting the thing they call is O(1) and permanent.
+
+**Scoped to competing readers of one semantic fact — not to compatibility forwarding.** Two
+ways to *ask* for an answer is an API convenience and this codebase keeps several on purpose
+(`make_drawing`'s facade, the `sheet_dsl` alias, `dimension`'s transitional overload). Two
+ways to *decide* an answer is the defect: `location_role` and `plan_locations` and the
+off-axis pass each independently concluding whether a hole has a position. A forwarding shim
+computes nothing, so it cannot disagree; a second inference path exists precisely to
+conclude, and will.
+
+When a parallel path cannot yet be deleted, it is named — and the two ways of naming one are
+different mechanisms that should not be confused:
+
+- **Unresolved debt** is inventoried AND **shrink-only**, so it can only ever get smaller:
+  `_PENDING_VALUE_CARRYING`, `_FMT_BUDGET`.
+- **An intentional shared route** is inventoried AND **behaviourally verified**, so the
+  listing stays a checked claim rather than an assertion: `_SAME_PATH_AS_ENVELOPE`, whose
+  members are proven by `test_the_same_path_verbs_really_share_the_route` to reach their
+  handle by the same two lines `envelope` does.
 
 The corollary is why this ADR keeps producing *tables* rather than checks: a check must be
 remembered at each site, a table is consulted from one. `location_datum`, `_FACTS` and
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -153,6 +153,27 @@ reach (`_staircase()` has no holes, which is how two dimensional paths stayed ou
 boundary unnoticed); the source ratchet covers every path whether a fixture walks it or not.
 It is shrink-only, and each survivor carries a written reason.
 
+**Designating an owner means DELETING the alternative reader, not redirecting the caller.**
+The rule that most reliably held while this ADR was implemented, and the one that explains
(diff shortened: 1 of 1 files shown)


pzfreo:
Adversarial review of `5c3549c`

I found one documentation-accuracy issue before I think this is ready:

**The new paragraph classifies `_SAME_PATH_AS_ENVELOPE` as named pending debt, but it is not pending migration work.** In `tests/test_sheet_identity_invariant.py` it is an intentionally maintained, behaviorally verified equivalence roster: those verbs are listed because they return `_Params` through exactly the same route as `envelope`, and `test_the_same_path_verbs_really_share_the_route` checks that invariant. Grouping it with `_PENDING_VALUE_CARRYING` and the shrink-only `_FMT_BUDGET` conflates two useful but different mechanisms.

I suggest splitting the claim:

- unresolved migration/debt must be explicitly inventoried and shrink-only (for example `_PENDING_VALUE_CARRYING` and `_FMT_BUDGET`);
- intentional shared routes or exceptions must be explicitly listed **and behaviorally verified** (for example `_SAME_PATH_AS_ENVELOPE`).

I would also scope the headline rule to *competing readers of the same semantic fact*. As written, “designating an owner means deleting the alternative reader, not redirecting the caller” is absolute enough to discourage legitimate compatibility forwarding at a public boundary. The historical examples support the narrower and stronger rule: once an internal semantic owner is designated, parallel inference/interpretation paths must be deleted.

Everything else in this ADR addition is consistent with the referenced cleanup history, and `git diff --check` is clean.

pzfreo:
## Both fixed — and both were overreach in my prose, not in the idea

### The headline was absolute enough to condemn legitimate shims

You are right, and the narrower form is also the **stronger** one. As written it ruled against `make_drawing`'s facade, the `sheet_dsl` alias and `dimension`'s transitional overload — all deliberate, all fine.

> **Once a semantic owner is designated, parallel INFERENCE paths must be deleted, not redirected.**

with the distinction stated: **two ways to *ask* for an answer is an API convenience; two ways to *decide* it is the defect.** A forwarding shim computes nothing, so it cannot disagree. A second inference path exists precisely to conclude, and will — `location_role` and `plan_locations` and the off-axis pass each independently deciding whether a hole has a position.

Every historical example is the second kind, so the scoping costs the rule nothing and stops it being quoted against the first.

### `_SAME_PATH_AS_ENVELOPE` is not pending debt

Correct, and I should have caught it — I wrote the verification test. Listing it beside the shrink-only inventories conflated two mechanisms that differ in **what makes them trustworthy**, so they are now separated on exactly that axis:

- **unresolved debt** — inventoried *and* **shrink-only**, so it can only get smaller (`_PENDING_VALUE_CARRYING`, `_FMT_BUDGET`);
- **an intentional shared route** — inventoried *and* **behaviourally verified**, so the listing stays a checked claim rather than an assertion (`_SAME_PATH_AS_ENVELOPE`, proven by `test_the_same_path_verbs_really_share_the_route`).

That distinction is worth more than the original sentence was: it says what a reader should *do* with each list.

Docs only; the ADR-inventory tests still pass. #934 is merged.

pzfreo:
Re-review of `b7d1c2f`: the two previous findings are fixed cleanly, but the tighter wording exposes one remaining factual inconsistency.

**`_PENDING_VALUE_CARRYING` is not wholly “unresolved debt” or wholly shrink-only.** The tuple contains both:

- `hc_`, explicitly described as a pending migration; and
- `pmi_`, explicitly described in the same test as “the one documented permanent exception.”

ADR 0016 itself reinforces that distinction earlier: `render_pmi` takes the model “on purpose and permanently” because it renders author-supplied text rather than inferring a generated measurement. Therefore citing the entire `_PENDING_VALUE_CARRYING` tuple as an example of unresolved debt that “can only ever get smaller” contradicts both the test commentary and the ADR’s source-of-truth table. It also puts a permanent non-inference exception under the lead-in “When a parallel path cannot yet be deleted”.

I suggest naming the `hc_` member specifically as the unresolved item, and treating `pmi_` as a permanent, behaviorally constrained boundary exception—or changing the inventory so pending migrations and permanent exceptions are represented separately. The former is enough for this docs-only PR; the latter looks like follow-up cleanup because the test tuple/comment already conflates those categories.

Apart from that, this revision is much stronger: the inference/forwarding boundary is clear, `_SAME_PATH_AS_ENVELOPE` is now characterized accurately, and `git diff --check` remains clean. Lint is green; the test matrix is still running.

pzfreo:
## Fixed — and the error was inside the paragraph warning about it

You are right: `_PENDING_VALUE_CARRYING` is `("hc_", "pmi_")`, one pending migration and one **permanent** exception, so citing the whole tuple as debt that "can only ever get smaller" contradicts both the test's own commentary and this ADR's contract table, where `render_pmi` takes the model *"on purpose and permanently"*.

The fix is a third category, because the missing one is exactly what I collapsed:

| survives because | trustworthy because |
|---|---|
| **unresolved debt** | inventoried **and shrink-only** — `_FMT_BUDGET`, `hc_` |
| **permanent exception** | inventoried **and argued** — `pmi_`, `render_pmi`/`render_gdt`: author-supplied text, so there is no inference path to delete |
| **intentional shared route** | inventoried **and behaviourally verified** — `_SAME_PATH_AS_ENVELOPE` |

A permanent exception needs an *argument*, not a shrink target, and a reader has to be able to tell which they are looking at without running `git log`. Writing a two-category rule while the code in front of me had three is the same conflation the paragraph exists to warn about, committed inside it.

## The code smell, folded into #926

Took your first suggestion for this PR (name the members separately) and recorded the second where the work lands rather than filing a fourth issue. #926 now covers splitting the tuple:

```python
_PENDING_MIGRATION = ("hc_",)        # shrink-only; empty when #926 closes
_PERMANENT_EXCEPTIONS = ("pmi_",)    # argued, never shrinks
```

That also repairs #926's own acceptance criterion, which I had written as the awkward *"the `_PENDING_VALUE_CARRYING` list is `("pmi_",)`"* — a pending list whose finished state is non-empty. It is now "`_PENDING_MIGRATION` is `()`", which is a checkable end state.

Docs only; the ADR-inventory tests still pass.

Issue #943, opened 2026-07-30
Emitted Sheet script for a turned part fails to rebuild: declared steps don't span the profile


pzfreo:
Found while implementing #938 (the dimension mirror). **Pre-existing** — verified identical on `main` — and independent of that change.

## Reproduction

```python
from build123d import Cylinder, Pos
from draftwright.builder import detect_part_model
from draftwright.sheet_emit import emit_sheet_script

part = Cylinder(15, 20) + Pos(0, 0, 20) * Cylinder(10, 15)   # a two-diameter turned shaft
src = emit_sheet_script(detect_part_model(part), "part", "s", title="T", number="N")
# run the emitted script →
```

```text
ValueError: step() declares a segment of a turned profile, but the declared steps don't
span this part's full height — it is not a turned (rotational) body. For a boss (an
external cylinder on a prismatic part) use .boss() — it renders its own ø and height.
```

The generated script **does not run at all** for this part. The detected model is `{rotational, step}` with `orientation="z"`, so the emitter writes `sheet.step(...)` lines — and `_assemble`'s validation then rejects them because the declared steps do not tile the full height.

## Why it matters

A generated script that raises is worse than one that diverges: `--script --style sheet` is the documented editable-drawing entry point, and for turned parts it produces a file that cannot be executed. The #707 class, at its most severe.

## Where to look

Either the emitter is not emitting every segment of the profile (so the declared steps genuinely do not span it — e.g. the base cylinder is represented by the `rotational` feature, which has no declarative verb and emits only a comment), or the validation is too strict for a model that also carries a `rotational` feature.

The first looks more likely: `rotational` is the kind with no declarative verb, so the part of the profile it represents is emitted as a comment and the remaining `step` lines cannot span the height on their own. If so this is the same root as #938's fallback — a feature kind with no verb makes the script incomplete — and both are relieved by giving `rotational` a declarative surface.

## Acceptance

- The emitted script for a turned part **executes**.
- Annotation-set parity with the automatic drawing, or a stated and tested divergence.
- A turned fixture in the emit corpus, since none currently exercises this path end to end.

Refs #938, #942, ADR 0011 (recognise + emit + declare round-trip).

pzfreo:
## Sharpened while implementing #945 — the root is not the missing verb

[...]

Issue #942, opened 2026-07-30
Epic: finish the code-generation surface — an editable script whose dimensions are editable


pzfreo:
The remaining ADR 0016 work on generated scripts, plus the emitter consolidation it enables. Successor to #867 (implementation epic, closed), which deliberately scoped the emitter out.

## The state that motivates this

**#922 built the capability and nothing uses it.** Three PRs shipped — features have handles (#931), the emitter binds names (#932), an authored set emits `dimension(...)` declarations (#933). But `--script` emits from `detect_part_model`, which never carries an authored set, so every generated script still says `sheet.auto_dimensions()` and **not one automatic dimension is commentable**.

So the product promise — *comment a line out to drop it* — holds for **features** and not for **dimensions**. Closing that is what this epic is for.

## Children

| # | Child | Gate / oracle |
|---|---|---|
| 1 | #938 — mirror the planner's set as `dimension(...)` lines | regenerated annotation set **equals** the automatic drawing's; commenting one line drops exactly that dimension |
| 2 | #939 — comment floor for what the mirror cannot express, with a named-gap ratchet | every annotation is either emitted or named; no third state |
| 3 | #940 — retire `--style imperative` | one emitter, no capability lost, proven per kind |
| 4 | #941 — redundancy lint | settle "lint or planner rule?" first |

Ordering: **#938 → #939 → #940**, with #941 independent. #940's gate *is* the definition of "#938 is done".

## Decisions already taken

- **Imperative is being retired** (2026-07-30). That closes **#935** (envelope dims replay as generic `dim_length*`) **by deletion** rather than by fixing it — verified imperative-only; the Sheet script round-trips the same fixture exactly. #935 is therefore blocked on #940, not scheduled.
- **Emit from intent, never from placed annotations.** Walking the drawing would let solver pressure rewrite version-controlled source, and a dropped dimension could never be recovered (ADR 0016, "the script records intent").
- **One line per addressable UNIT.** Correlated sets emit one line; there is no member line to mislead.

## Out of scope

- `RelationDimensionId` and its selector spelling — inter-feature spans stay on the comment floor until that lands.
- Per-**member** location identity (#883).
- The recognition/completeness gaps that make a drawing under-dimensioned regardless of surface (#916, #917, #918, #909).

## Why it matters beyond tidiness

[...]

Issue #939, opened 2026-07-30
Comment floor for dimensions the mirror cannot express, with a named-gap ratchet


pzfreo:
Child of #942.

Child of the code-generation epic; pairs with #938.

## Why

Once #938 mirrors the planner's set as `dimension(...)` lines, "absence of a line means suppressed" becomes the script's contract. That claim is only safe over the set the emitter can actually express — so everything it cannot must be **visible in the file**, not missing from it.

ADR 4 (was 0016) already inventories what is not mirrorable:

| Not mirrored | Why |
|---|---|
| Inter-feature spans and angles | no `(feature, role)` form — needs `RelationDimensionId`, selector spelling still open |
| Imported AP242 PMI | materialised (`ref_pts`/`ref_bbox`/`at`), nothing to reference → `measured_dimension(...)` |
| Low-level furniture | centre marks, section arrows, hatching, NTS caption — no editable intent, engine-automatic by decision |
| Anything the emitter cannot re-solve | the fidelity floor `emit_sheet_script` already holds for features |

## Scope

- Emit a **self-describing comment** for each drawn thing the mirror cannot express, naming what it is and why there is no line — the floor `emit_sheet_script` already applies to feature kinds with no declarative verb.
- A **ratchet** that the set of unexpressible dimensions is exactly the ADR's list. Without it the comment floor silently absorbs new gaps, which is how `_PENDING_VALUE_CARRYING` came to mean "furniture" (#925 review round 4).
- Distinguish the three reasons a thing is unexpressible, per ADR 4 (was 0016)'s own three-category rule: **pending** (shrink-only), **permanent** (argued — PMI, furniture), **blocked on a decision** (relations, on `RelationDimensionId`).

## Acceptance

- Every annotation in the automatic drawing is either produced by an emitted line or named by a comment. No third state.
- The test enumerates it rather than sampling: a new dimension kind that is neither mirrored nor commented fails.

Refs #938, ADR 4 (was 0016 phase 4).

pzfreo:
**Not closed — I over-claimed, and the review was right to stop it.**

I proposed closing this on the grounds that its acceptance was already met by
`TestTheScriptAccountsForEveryAnnotation` plus the `_PLAN_DIMENSIONAL_FIELDS` ratchet. That is
wrong on two counts:

1. **The comments are not in the generated file.** This issue asks that every drawn thing the
   mirror cannot express be *"visible in the file, not missing from it"* — a self-describing
[...]

PR #944, opened 2026-07-30
feat(#938): mirror the planner's dimension set as commentable dimension(...) lines

### LINKED ISSUE #938: Mirror the planner's dimension set as commentable dimension(...) lines
Child of #942.

ADR 0016 **phase 4**, the core. Child of the code-generation epic.

## The gap

#922 built the capability across three PRs — features have handles (#931), the emitter binds names (#932), an authored set emits `dimension(...)` declarations (#933). **Nothing uses it.** `--script` emits from `detect_part_model`, which never carries an authored set, so every generated script still says:

```python
sheet.auto_dimensions()
```

and **not one automatic dimension is commentable.** The plumbing is finished; the CLI path does not walk it.

That is the product promise of an "editable drawing": comment out a line to drop that dimension. It holds for features today and not for dimensions.

## Scope

Emit one `sheet.dimension(feature, role)` line per planned `AddressableDimension`, instead of `auto_dimensions()`.

- **One line per addressable UNIT, never per member.** A `step_height` / `step_position` ladder and a rotational body's bores are one `AddressableDimension` holding N members — emit one line; commenting it drops the set. (ADR 0016 identity tier 3.)
- **Locations get a line**, per feature — possible since #925 gave them a role; `dimension(hole1, "location")`.
- **Emit from INTENT, never from placed annotations.** Walking the drawing is the obvious way to build a mirror and it is wrong: a dimension the solver dropped would vanish from the regenerated script and could never be recovered, and unrelated layout churn would rewrite version-controlled source. ADR 0016 states this; it is the easiest thing to get wrong here.
- The script leads with `authored_dimensions()` (#933), so its source stays explicit.

## Acceptance

- The regenerated script's **annotation set equals** the automatic drawing's — asserted on drawn annotations, not on emitted text. A text check passes on a script naming the wrong feature, which is what positional addressing would have produced (the #889 / #934 assertion shape).
- Commenting out one emitted `dimension(...)` line drops **exactly** that dimension.
- Commenting out a correlated-set line drops the whole set, and no member line exists to mislead.
- A fixture corpus covering every feature kind, not one part — the lesson of #925/#934, where single-ladder and no-hole fixtures hid whole paths.

Refs #867 (closed epic), #922, #933, ADR 0016 phase 4.

pzfreo:
Closes #938. First child of #942.

## The gap

**#922 built the capability and nothing used it.** `--script` emits from `detect_part_model`, which never carries an authored set, so every generated script said `sheet.auto_dimensions()` and **not one automatic dimension was commentable**.

So the editable-drawing promise held for *features* and not for *dimensions*: the only way to drop one dimension was to comment out its whole feature — losing that feature's callout, centre marks and location with it.

```python
sheet.authored_dimensions()
sheet.dimension(hole1, "bore.diameter")
sheet.dimension(hole1, "location")
sheet.dimension(envelope1, "width.length")
```

## Three things the implementation had to settle

**The overall height on a part with no `EnvelopeFeature`.** A mirrored set *is* an authored set, and `_compile_overall_height` deliberately refuses the bounding-box fallback under one — a script "must not acquire a measurement its author had no way to ask for" (#925). Mirroring would have silently dropped it: **#889's bug at the authored level**. The script now declares an envelope and names only its `height`. Honest rather than a workaround — the engine was already dimensioning the bounding box, and the declaration writes down what it was doing implicitly. Naming only `height` keeps width and depth omitted exactly as the planner left them (verified by set equality, not by inspection).

**Features with no declarative verb** (`rotational`) emit a comment, so they bind no name and cannot be referenced. Those models keep `auto_dimensions()` **and say why** — a mirrored set that silently omitted their dimensions would claim a completeness it does not have. Derived from the emitter's own fallback (`_feature_line` returns a comment) rather than a second table of kinds.

**Locations come from the compiled set**, not `plan_locations`. The two differ by exactly the positions compiled elsewhere — a slot's near-end offset, a side-drilled hole's — and walking the planner lost them on pad/slot parts.

## Two pre-existing bugs it surfaced

The mirror turns silent divergence into a raise, which is how both appeared:

- **The emitted boss line dropped the boss HEIGHT.** The declared boss carried only `boss.diameter` while the detected one also carried `boss_height.length` — an ADR 0011 round-trip violation (recognise, emit and declare must agree about a feature's parameters). Fixed here.
- **#943 filed**: a turned part's emitted script fails to *execute* (`declared steps don't span this part's full height`). Verified identical on `main`, independent of this change, and not fixable without giving `rotational` a declarative surface.

## Tests

**Annotation-set equality over a nine-part corpus** — plate, no-envelope flange, stepped, slot, pocket, two-hole, pad, bare block, side-drilled. On drawn annotations, not emitted text: a text assertion passes on a script naming the *wrong feature*, which is what positional addressing produced before #932. A corpus rather than one part, because single-fixture guards are how whole paths stayed uncovered twice in this series.

Plus: commenting out one line drops **exactly** one dimension; a correlated set emits exactly one line; the unnameable-feature fallback; the synthesised envelope names only its height.

Three existing tests kept their property and changed their expectation — the script states its source (either verb), no MEASURED overload (`dimension(kind=` rather than the bare substring), feature lines bind names (scoped to the feature block).

Lint, smoke and `test_sheet_emit.py` (**121 passed**) green; CI has the full matrix.

### DIFF
--- a/src/draftwright/sheet_emit.py
+++ b/src/draftwright/sheet_emit.py
@@ -203,8 +203,14 @@ def _feature_line(f) -> str:
         thr = (
             f", thread={f.thread!r}" if getattr(f, "thread", None) else ""
         )  # external thread (#859)
+        # The HEIGHT round-trips too (#938). Dropping it made the declared boss carry only
+        # `boss.diameter` while the detected one also carries `boss_height.length`, so the
+        # regenerated model could not express a dimension its source had — silently before
+        # the mirror named it, and as a raise afterwards. ADR 0011's round-trip rule is that
+        # recognise, emit and declare agree about a feature's parameters.
+        height = f", height={_n(f.height)}" if getattr(f, "height", None) else ""
         return (
-            f"sheet.diameter(diameter={_n(f.diameter)}, at={_pt(f.frame.origin)}, "
+            f"sheet.diameter(diameter={_n(f.diameter)}{height}, at={_pt(f.frame.origin)}, "
             f'axis="{f.frame.axis}"{thr})'
         )
     if k == "step":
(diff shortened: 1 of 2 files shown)


pzfreo:
## Codex adversarial review — round 2: no correctness blocker

Two rounds run in an isolated worktree, open-ended prompt (explicit licence to reject the approach).

### Round 1 found three, all correct, all fixed in `95b66dc`

| | finding | fix |
|---|---|---|
| Medium | the all-or-nothing fallback made a partial implementation look complete — *"adding one rotational feature changes every dimension declaration from explicit to implicit"* | filed **#945** (the missing ADR 0011 declarative verb for `RotationalFeature`, also #943's root); the emitted comment now **names the kind and cites the issue**; the test pins *unfinished work* rather than a settled limitation, with a second test that fails if the reference goes stale |
| Low | `object.__setattr__` on a frozen public IR dataclass — *"emitter bookkeeping masquerading as model state"* | `mirror_model` returns `(model, envelope)`; marker and rediscover-by-shape helper both deleted |
| Low | name-set equality too weak — a request rebuilt with the wrong **value** keeps the same registry name | corpus compares `_annotation_signature` (labels, specs, leader coverage, boxes) — which immediately caught a fixture artefact where the automatic build carried no title |

### Round 2 verdict

> No correctness blocker found in the current revision.

with a Medium on architecture that it explicitly does **not** consider blocking:

> I do **not** think deferring this redesign is the wrong call for PR #944. The current code reuses canonical planner/compiler functions rather than duplicating selection rules, and the remaining architectural debt is bounded and visible. **I would require a tracked follow-up, not block this PR.**

Filed as **#946** — a compiler-level export result carrying every addressable approved intent, consumed by both rendering and emission, so completeness stops being maintained by memory. It would subsume both the synthesised envelope and #939's comment floor.

Its assessment on the axes you asked about:

- **Correctness** — *"no uncovered concrete divergence"*; parity includes values, spans, sides, labels, leader coverage and furniture boxes.
- **Technical debt** — *"bounded and visible"*; the one item is #946.
- **API** — *"maintainable and safely editable. Commenting a referenced feature fails loudly instead of retargeting another feature."*
- **ADR fit** — *"Good fit with ADR 0016 and ADR 0011"*; the only tension is #946's missing abstraction, and no second rendering path is created.
- **Tests** — *"tests the property rather than merely the implementation's textual shape… a credible regression gate rather than a proxy."*

Waiting on CI before merging.

Issue #946, opened 2026-07-30
Compiler-level export representation: one addressable result for both rendering and emission


pzfreo:
Child of #942. Required follow-up from the Codex adversarial review of #944 (#938), which raised it twice and judged it a tracked-debt item rather than a blocker.

## The observation

`sheet_emit._mirrored_requests` assembles the emitted dimension set from **three** representations:

| source | for |
|---|---|
| a synthesised `EnvelopeFeature` | the compiler's bounding-box overall height |
| `plan_dimensions()` | feature parameters |
| `compile_dimensions().locations` | positions |

The split is already encoding compiler knowledge — the code carries comments explaining *why* `plan_locations()` is insufficient (it misses slot and side-drilled positions, which are compiled elsewhere) and *why* the overall height needs reconstructing (`_compile_overall_height` refuses the bbox fallback under an authored set).

> A future compiler-owned dimension category can render automatically yet disappear from generated scripts unless someone remembers to extend `_mirrored_requests`.

That is the failure mode, stated precisely. It is the same shape as the defects the #925 series kept finding — a fact the compiler owns, re-derived by a consumer — one level up from where ADR 0016 Amendment 1 fixed it for renderers.

## Why it is debt rather than a bug

Nothing is currently wrong: #938's corpus proves annotation-signature parity across nine parts, and the emitter reuses the canonical planner/compiler functions rather than duplicating selection rules. The debt is that **completeness is maintained by memory** instead of by construction.

It is also weaker than ADR 0015's compiler-waist direction: the compiler is meant to be the one place that decides what the drawing carries, and here a façade re-assembles that decision.

## Scope

A compiler-level export result carrying **every addressable approved intent in order**, each with either a declarative target (the feature and role a script would name) or an explicit *not-representable* result and its reason. Both rendering and script emission consume that one result.

It need not expose placement coordinates or placed annotations — the point is the addressable *intent*, not the page.

Two things it would subsume:
- the synthesised envelope (the overall height would carry a model-level target, so no feature needs inventing);
- `#939`'s comment floor (the *not-representable* results are exactly what the floor prints).

## Acceptance

- `_mirrored_requests` is a serialisation of one compiler result, not an assembly of three.
- A new compiler-owned dimension category appears in generated scripts without editing `sheet_emit.py`, and a test enforces that rather than a reviewer noticing.
- The synthesised-envelope mechanism is deleted.

Refs #938, #939, #944, ADR 0015, ADR 0016.

[...]

Issue #945, opened 2026-07-30
RotationalFeature has no declarative verb — the missing ADR 0011 round-trip leg


pzfreo:
Child of #942. Raised by the Codex adversarial review of #944 (#938's implementation), which identified it as the thing that makes the dimension mirror partial.

## The gap

`RotationalFeature` is **recognised** and **rendered** but cannot be **declared** or **emitted**. ADR 0011's round-trip rule is that every recognised feature supports all three surfaces (epic #574 closed on that basis; this kind was never covered).

Concretely, `sheet_emit._feature_line` has no branch for it, so a turned part emits:

```python
# rotational @ (0, 0, 0) — no declarative verb yet; drawn by the auto-pass
```

## What it costs, now that #938 has landed

The dimension mirror is **all-or-nothing per part**. A rotational feature carries planned dimensions (`od`, `bore`) that no emitted line can name, so a mirrored set would silently omit them and claim a completeness it does not have. #938 therefore falls back to `sheet.auto_dimensions()` for the whole part.

The review's criticism is fair and worth recording verbatim:

> Adding one rotational feature changes every dimension declaration in the generated script from explicit to implicit. Removing that feature later causes the generator to rewrite the entire dimension section. Users cannot explicitly edit supported dimensions merely because one unrelated dimension is unsupported.

So a turned part — a large fraction of real machined parts — gets none of #938's benefit, and the generated script gives the user no name for the dimension that caused it.

## It is also the root of #943

A turned part's emitted script **fails to execute** (`declared steps don't span this part's full height`), because the profile segment the `rotational` feature represents emits only a comment, leaving the `step` lines unable to span the height. Same missing verb, more severe symptom.

## Scope

- A declarative verb (`sheet.rotational(...)` or an extension of an existing one) that rebuilds a `RotationalFeature` — OD, axis, bores.
- The emit line, so `_feature_line` binds a name for it.
- Then #938's `_is_mirrorable` fallback can be deleted and its test converted from "this limitation is locked in" to "every part mirrors", and #943 re-tested.

## Acceptance

- A turned part's emitted script executes (**closes #943**).
- A turned part mirrors its dimensions rather than falling back (**removes #938's fallback**).
- Annotation-signature parity for a turned corpus.

Refs #938, #943, #942, ADR 0011, epic #574.

PR #947, opened 2026-07-30
test(#939): account for every annotation a generated script does not declare

pzfreo:
Second child of #942. Follows #938 (#944).

## The claim that needs guarding

#938 made a generated script mirror the planner's set as `dimension(...)` lines, so **absence of a line now means suppressed**. That claim is only safe over the set the emitter can actually express.

So every annotation the drawing carries must be one of two things:
- produced by a line the script contains, or
- named as furniture, with an argument.

**No third state.** A dimensional mark the script cannot express would sit in the drawing with nothing to comment out and nothing saying why — and the reader would conclude the engine chose it, when in fact the script could not say anything about it.

## Checked behaviourally, not by name

Comment out **every** `sheet.dimension(...)` line and assert what survives is exactly the named furniture:

```
plate+hole   ['m_cm0', 'note_iso_nts', 'title_block']
flange       ['centerline_plan', 'centerline_side', 'm_cm0'…'m_cm4', 'note_iso_nts', 'title_block']
stepped      ['note_iso_nts', 'title_block']
slot         ['note_iso_nts', 'title_block']
pocket       ['note_iso_nts', 'title_block']
pad          ['note_iso_nts', 'title_block']
side hole    ['note_iso_nts', 'title_block']
```

Every dimensional annotation disappears. A prefix taxonomy was the obvious implementation and is a **proxy** — it asserts what the author believed about the names, and that shortcut has been the bug twice in this series (`_VALUE_FREE` absorbing the pitch dim; `_PENDING_VALUE_CARRYING` read as "furniture").

## The furniture list is argued, not just listed

Per ADR 0016's three-category rule (recorded in #937), these are **permanent exceptions**, so each carries a reason it can *never* be a dimension line. Two guards enforce it:

- every entry needs a real argument — which immediately caught my own one-word `"section hatching"`;
- every entry must be **reachable**, because a stale prefix silently widens what the accounting forgives, letting the next unexpressible dimension through under a name nobody produces any more.

## Scope: deliberately half of #939

The other half — *emitting* comments for unexpressible dimensions — is subsumed by **#946**, the compiler export result Codex asked for on #944: its "not-representable" entries are exactly what such a floor would print. Building it now would be work #946 deletes.

That is the deletion discipline (#937) applied *before* creating the debt rather than after. #939 stays open for that half; this PR closes the part that is durable regardless of how #946 lands.

Lint, smoke and `test_sheet_emit.py` (**133 passed**) green; CI has the full matrix.

### DIFF
--- a/src/draftwright/sheet_emit.py
+++ b/src/draftwright/sheet_emit.py
@@ -209,9 +209,16 @@ def _feature_line(f) -> str:
         # the mirror named it, and as a raise afterwards. ADR 0011's round-trip rule is that
         # recognise, emit and declare agree about a feature's parameters.
         height = f", height={_n(f.height)}" if getattr(f, "height", None) else ""
+        # The SPAN too, not just the height. Detection reports `frame.origin` as the boss TOP
+        # while `declare.boss` reads `at` as its CENTRE, so round-tripping the origin alone
+        # shifted the height dimension by half the boss — same value, wrong witness lines
+        # (#947 review, found by a boss fixture the corpus previously lacked). Emitting the
+        # span states the geometry outright instead of relying on the two ends agreeing about
+        # a coordinate convention.
+        span = f", span=({_pt(f.span[0])}, {_pt(f.span[1])})" if getattr(f, "span", None) else ""
         return (
-            f"sheet.diameter(diameter={_n(f.diameter)}{height}, at={_pt(f.frame.origin)}, "
-            f'axis="{f.frame.axis}"{thr})'
+            f"sheet.diameter(diameter={_n(f.diameter)}{height}{span}, "
+            f'at={_pt(f.frame.origin)}, axis="{f.frame.axis}"{thr})'
         )
     if k == "step":
         thr = (
--- a/src/draftwright/sheet_emit.py
+++ b/src/draftwright/sheet_emit.py
@@ -687,17 +687,33 @@ def _dimension_block(model, names: dict[int, str], synthesised_envelope=None) ->
     if model.authored_dimensions is None and not _is_mirrorable(model):
         # A feature with no declarative verb carries planned dimensions, so a mirrored set
         # would silently omit them and claim completeness it does not have (#938).
+        # WHY, specifically. `_is_mirrorable` now fails for any dimension the compiler
+        # approved and no line can name — not only the no-declarative-verb case — so blaming
+        # #945 unconditionally would misdirect a reader whenever the cause is something else,
+        # and could print an empty kind list (#947 review).
+        missing = unmirrored_dimensions(model)
         unnameable = sorted(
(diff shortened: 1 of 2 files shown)


pzfreo:
## Codex adversarial review — 4 rounds, clean

> **No blocking findings. I would merge this.**
>
> The structural guard now uses the production emitter's actual request set, checks every current compiled-plan dimension collection, requires nameable features, and falls back to `auto_dimensions()` when completeness cannot be claimed. That adequately closes the third state; #946 is now diagnostic/disclosure improvement, not a correctness prerequisite.

Trajectory: **3 findings (1 HIGH) → 3 (2 HIGH) → 3 (0 HIGH) → 0.**

### What the rounds found, and what they were worth

| round | finding | why it mattered |
|---|---|---|
| 1 | the guard was *"a smoke test over the current happy-path corpus"* | the `"side-drilled"` fixture detected only an `envelope` — the path it was named for was untested. `"slot"` detected a pocket. |
| 1 | `_SCRIPT_FURNITURE` was a prefix exemption | `"section_"` would forgive a future `section_depth` measurement — the exact blind spot the test's own commentary claimed to supersede |
| 2 | **`_is_mirrorable()` was a test-only approximation** | a live production bug **in this PR**: a compiled dimension outside `plan_dimensions` on a nameable feature left the script printing *"THIS IS THE COMPLETE SET"* while omitting it |
| 2 | no exhaustiveness over `RenderableDimensionPlan` | a new `plan.ordinate_dimensions` could be rendered-and-unmirrored with every test green |
| 2 | `_is_value_bearing` was **dead code** | I introduced it as "the classification that matters" and never called it; the list was validated by the *length of its prose* |
| 3 | the furniture list was written **from memory** | `section_label`/`section_line` do not exist — the renderer emits `section_caption`. Three dead entries, and it would have flagged a real sectioned drawing. |
| 3 | a test whose **name lied** | it claimed to construct a synthetic model per kind; it compared two tables |

### Real bugs the rounds surfaced

Adding the fixtures Codex demanded found **two recognise↔declare disagreements, both in the boss** — the emitted line dropped its **height** (fixed in #944), and `declare.boss` read `at` as the boss **centre** while detection reports its **top**, shifting the height dimension by half the boss (fixed here). Neither was visible without a fixture that built the feature and compared signatures.

### Tracked, not absorbed

**#943** turned script won't execute · **#945** rotational declarative verb (root of #943 *and* the mirror's fallback) · **#946** compiler export representation · **#948** eight IR kinds with no round-trip fixture.

On #948 specifically — those eight are named rather than silently claimed, and Codex's argument for why that matters is concrete: a kind can be consistent across IR, `_FACTS` and `_feature_line` and still have a lossy declaration. Two such defects have already been found this epic, both in one feature.

**161 passed**, all four lint checks clean. Waiting on CI.

Issue #958, opened 2026-07-31
Multi-solid recognition combines faces into phantom milled features


pzfreo:
## Outcome

Recognition must never construct one machined feature from faces belonging to different solids. A multi-solid assembly should preserve each body's own slots and pads without inventing a slot in the air gap or treating a detached body as a pad on another body.

## Corrected evidence

The original fixture was described as an attached raised pad:

```python
part = Box(80, 60, 10) + Pos(0, 0, 10) * Box(30, 20, 4)
```

It is actually two disconnected solids:

| body | Z span |
|---|---|
| base | -5 to 5 |
| smaller box | 8 to 12 |

The 3 mm air gap is exactly the reported phantom slot:

```
Slot(width_axis='z', long_axis='x', width=3.0,
     w_center=6.5, lo=-15.0, hi=15.0, d_lo=-10.0, d_hi=10.0)
```

The same run also reports the detached smaller solid as a `RaisedPad`, then emits both `m_pad0_length = 30` and `m_slot0_length = 30`.

A genuinely attached control, `Pos(0, 0, 7) * Box(30, 20, 4)`, is one solid. It recognises one pad, no slot, and emits the 30 mm span only once.

## Root cause

This is recognition, not planner de-duplication:

- `recognise_slots` scans `part.faces()` as one pool and pairs facing walls from different solids;
- `recognise_rectangular_pads` also scans the aggregate face set and global bounding box, so a detached smaller body can satisfy the raised-pad predicate;
- suppressing one duplicate later would retain two false physical records and conceal the ownership defect.

The likely implementation shape is one public orchestration call with per-solid internal scans. That preserves ADR 0017's once-per-family ownership while enforcing ADR 0013's rule that recognised records come from coherent physical evidence.

## Acceptance

- [ ] The original two-solid fixture produces neither a cross-body `Slot` nor a detached `RaisedPad`.
- [ ] A genuinely attached raised pad remains recognised, dimensioned once, and lint-clean.
- [ ] A real through-slot contained within one body of a multi-solid assembly remains recognised.
- [ ] Genuine slots, pockets, channels, and pads on separate bodies remain independently recognised; derived cross-body patterns are tracked by #1073.
- [ ] `build_recognition_result` still invokes each public recogniser family once; declared builds still invoke none.
- [ ] No planner or renderer rule suppresses the symptom by matching equal numeric spans.
- [ ] Mutating the scan back to the aggregate face pool makes the cross-body regression fail.

## Scope

[...]

Issue #955, opened 2026-07-31
Z-turned overall height is suppressed unconditionally, so a dropped step chain leaves the height dimensioned by nothing


pzfreo:
## Summary

For a Z-turned part, `_compile_overall_height` (`model/compiled.py`) suppresses the overall
height **unconditionally**, with the reason `"Z-turned (the step chain tiles the height)"`.

That premise is settled at compile time, but the step chain's fate is settled much later, at
**placement**: `_draw_step_chain` (`annotations/from_model.py`) drops the whole chain when the
shoulders are too closely spaced to dimension at the chosen scale. When that happens, the
height is conveyed by neither the chain (dropped) nor the envelope (suppressed).

`axial_length_missing` fires, so the drawing is not silently wrong — but the user gets a lint
warning instead of a dimension, on a drawing where the correct answer (one overall height) was
available all along.

## Repro

```python
from build123d import Align, Cylinder, Pos
from draftwright import build_drawing

shaft = Cylinder(12, 60, align=(Align.CENTER, Align.CENTER, Align.MIN))
shaft -= Pos(0, 0, 30) * (
    Cylinder(12, 4, align=(Align.CENTER, Align.CENTER, Align.MIN))
    - Cylinder(9, 4, align=(Align.CENTER, Align.CENTER, Align.MIN))
)
d = build_drawing(shaft, number="X")
print([lab for _, o in d.iter_annotations() if (lab := getattr(o, "label", None))])
print(d.lint_summary()["by_code"])
```

```
['ø24', '4 WIDE × ø18', 'DRAWING', 'ISO VIEW (NTS)']
{'leader_crosses_silhouette': 1, 'axial_length_missing': 1, 'step_dim_dropped': 1}
```

No 30, no 26, no 60. The 4 mm groove puts two shoulders 4 mm apart, which fails the legibility
gate, which drops the **entire** chain rather than the crowded pair.

Affects the detected and declared paths identically (found while reviewing #953).

## Why it isn't a one-line fix

The two decisions sit on opposite sides of the ADR 0016 compiled-plan boundary:

- The compiler cannot know that placement will fail — legibility depends on the page and scale.
- The renderer must not invent dimensional content it did not receive; that is precisely what
  Amendment 1 forbids. It cannot simply draw a height when it drops the chain.

So this needs a design decision, roughly one of:

1. **Contingent rung** — the plan carries the overall height marked "draw only if the chain
   does not survive", making the fallback content the compiler approved rather than content
[...]

PR #959, opened 2026-07-31
feat(#941): report two dimensions that state the same measurement

### LINKED ISSUE #941: Redundancy lint: report over-dimensioning that is neither identical nor coincident
Child of #942.

ADR 4 (was 0016) **phase 5**. Child of the code-generation epic.

## Why

The engine has two duplicate-protection layers today — value/identity dedup, and the corridor solve's coincident-span dedup. Neither catches a drawing that is **over-dimensioned without duplicating anything**: a pattern given both its per-hole locations *and* its pitch says the same thing twice in different words, and every mark is distinct.

It becomes more visible once #938 mirrors the planner's set as explicit lines: a reader looking at 40 `dimension(...)` lines needs to know which are redundant, and "the engine chose them" stops being the answer when the script is the source of truth.

## Scope

- Report over-dimensioning as lint, at a severity that keeps a valid sheet clean (the #133 precedent: a completeness shortfall is `info`, a defect is `warning`).
- Name the pair and the reason, so the fix is a line to comment out rather than a puzzle.
- Do not auto-resolve it. Which of two redundant dimensions to keep is a drafting judgement — ISO 129 prefers the functional one — and guessing produces a plausible wrong drawing, which this codebase ranks below a visible message (#630/#631).

## Open question to settle first

**Is redundancy a lint or a planner rule?** If the planner should not emit both in the first place, this is a suppression rule and belongs in `plan_dimensions`. If a script may legitimately declare both (a manufacturing preference), it is lint. The answer likely differs per pair — a pattern's pitch-vs-locations is a planner choice; two authored lines are the author's.

Settle that before implementing; it decides where the code lives.

Refs #938, ADR 4 (was 0016 phase 5).

pzfreo:
Closes #941. Child of #942 (ADR 0016 phase 5) — the **last** child of the epic.

## The open question, settled with measurements

The issue said to settle "is redundancy a lint or a planner rule?" before implementing. I
measured the planner's actual output. The answer is neither option as posed, and it rescoped
the work — full evidence in [this comment](https://github.com/pzfreo/draftwright/issues/941#issuecomment-5139291120).

**The planner cannot over-dimension.** Every span it approves along an axis is measured from
a **common datum**:

```
step_height     z -20 → 10      overall_height  z -20 → 20
step_position   x -30 → 10      m_env_width     x -30 → 30
```

Common-datum spans cannot close a chain — which is the thing ISO 129 warns about. The derived
remainders (10 and 20) are left unstated, which is correct datum dimensioning. A suppression
rule in `plan_dimensions` would have nothing to suppress.

**The motivating example does not occur.** A 3-hole linear pattern draws `2× 40` **plus one
origin location** — not per-hole locations *and* pitch. Pitch gives spacing, location gives
where the array sits. Complementary, not redundant.

**The referential authored path can't create redundancy either** — `dimension(feature, role)`
picks a *subset* of the addressable set, and a subset of a non-redundant set is non-redundant.

**The materialised path can, and did.** `measured_dimension(...)` (and imported AP242 PMI, the
same IR kind) carries its own value and ref points and is checked against nothing:

```
m_env_width  '40'
pmi_x_0      '40'      <-- the same measurement, twice
lint: {}               <-- reported by nothing
```

## What this adds

`redundant_dimension`, severity `warning`, in `linting/structural.py`. It reports only what it
can prove — same orientation, same measured length, coincident extent along the measuring axis.

No view mapping is needed or guessed at: the orthographic views are aligned, so a plan-view
width and a front-view width share the page's x band by construction.

It **names the pair and picks no winner**. ISO 129 keeps the functionally significant
dimension, which is a judgement about the part; guessing produces a plausible wrong drawing,
which this codebase ranks below a visible message (#630/#631). A `_suggest_fix` snippet says so.

## It found a real defect on its first sweep

One true positive across 20 parts: `Box(80,60,10) + Pos(0,0,10)*Box(30,20,4)` states the same
30 mm span twice — as the pad's length and as a mis-recognised slot's. Identical `long_axis`,
`lo`, `hi`. Filed as **#958**, and excluded from the no-false-positive corpus **by name with the
reason**, so the exclusion is visible rather than a silent gap.

## Deliberately NOT done

No new planner suppression rule. That half of phase 5 **already exists**, ad hoc, in
`_compile_overall_height` ("Z-turned (the step chain tiles the height)"). #955 records that it
is *unsound* rather than missing — it fires on a premise placement can invalidate, so when the
chain drops nothing conveys the height. Fixing that beats adding a second one.

## Tests

- 6 unit tests on the check's discriminators: same length at a *different* position is not
  redundant (value coincidence ≠ measurement coincidence), different lengths over overlapping
  spans, perpendicular dims, three-over-one-span reports every pair.
- End-to-end: the restatement is reported; the same part without it is clean (the control);
  and `test_the_planner_alone_does_not_over_dimension` makes the "lint, not a planner rule"
  conclusion **executable** — if a planner change starts producing redundancy, it fails and the
  conclusion gets revisited.

## Verification

- `ruff check` / `ruff format --check` / `mypy` / `codespell` clean
- smoke 54 passed
- lint suites + emitter + layout-hypothesis: 309 passed
- `test_make_drawing.py` + `test_e2e_standards.py`: 717 passed (no fallout from the new code)

### DIFF
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -710,8 +710,8 @@ one exists today:
    different features measuring the same span.
 3. **A redundancy lint** — for over-dimensioning that is neither identical nor coincident,
    e.g. carrying both a pattern's per-hole locations *and* the pitch that determines them.
-   **This does not exist**: the current codes are `*_dropped` and `feature_not_dimensioned`;
-   nothing reports duplicate or redundant dimensioning. Adding it is part of this work.
+   **Landed as `redundant_dimension` (#941, 2026-07-31)**, but narrower than written here,
+   because the motivating example turned out not to occur — see the amendment below.
 
 ### Dimensions are sheet-level; the view is derived placement
 
@@ -1089,7 +1089,8 @@ second with the single-source-of-truth of the first — over the identified set,
   editable line, so the intent-mirror property holds; the two verbs stay distinct precisely
   because one references and the other restates.
 - **A duplicate/redundancy lint code joins `linting/structural.py`** — the third protection
-  layer; nothing reports redundant dimensioning today.
+  layer; nothing reports redundant dimensioning today. **Done** (#941): `redundant_dimension`,
+  severity `warning`, reporting the pair rather than picking a winner.
 - **`--style imperative` retires once the declarative mirror reaches its coverage**, leaving
   one generated output. The low-level `Drawing` verbs stay a hand-use API but stop being a
   generated surface (ADR 0001 §3), and the generated file deliberately loses its
@@ -1149,6 +1150,7 @@ second with the single-source-of-truth of the first — over the identified set,
    not yet mirrorable.
 5. **Redundancy lint.** The third duplicate-protection layer — report over-dimensioning that
    is neither identical nor coincident (a pattern's per-hole locations *and* its pitch).
+   **Landed** (#941), rescoped — the pattern example does not occur; see the amendment.
 6. **Retire `--style imperative`** once the mirror reaches its reconstruction coverage
    (rotational, the ladders, off-axis `locate`, machined callouts, pocket / slot patterns),
    leaving the declarative script as the single generated output. **Landed** (#940). The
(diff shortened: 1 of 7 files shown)


pzfreo:
## Blocked on #961 — marking draft rather than merging

Adversarial review found the check's **measurement representation** unsound, twice, on the
same underlying issue. I chased it and confirmed the reviewer is right, so this should not
merge as a warning-severity lint.

**What's proven and worth keeping:** the check's value. On its first corpus sweep it found
#958 — a real duplicate in an *automatic* drawing, where a pad and a mis-recognised slot carry
the identical x span and both get dimensioned. It also correctly reports the
`measured_dimension` restatement it was written for, and does not fire on 26 other parts
including cubes and square plates.

**What's not sound:** deciding which world axis a dimension measures. Review demonstrated a
missed duplicate (a 4 mm width on `Box(4,20,100)`), and I could not fix it:

| annotation | measured | bbox w | bbox h |
|---|---:|---:|---:|
| `dim_height` | 100.0 | 8.0 | 100.1 |
| `m_env_width` (measures **X**) | 4.0 | 14.8 | 7.05 |
| `pmi_x_0` (measures **X**) | 4.0 | 14.8 | 60.9 |

`max(w, h)` misreads `pmi_x_0`. "Closest to `measured_length`" misreads `m_env_width`. I tried
the second heuristic and it broke a case the first got right — the page bounding box just does
not encode the axis, because it is dominated by arrows, extension lines and the label offset.

The model layer does not answer it either: `ApprovedDimension.span` is `None` for feature
dimensions, so `pad_length.length` and `slot_length.length` — the two in #958 — carry no world
span to compare.

**So the prerequisite is #961**: populate `span` on every `ApprovedDimension`, making the
compiled plan the single answer to "what does this dimension measure", which is what ADR 0016
Amendment 1 says the plan is for. Once that lands, this check becomes a comparison of exact
world spans: no orientation inference, no view→axis table, no scale or detail-view blind spot,
and it can be re-pointed in a few lines.

Also worth noting the review corrected a claim of mine that outlived the code: I had written
"the planner cannot over-dimension", and #958 refutes it — the planner accepted two roles on
two features naming the same span and emitted both. ADR 0016 Amendment 2 now claims only what
was measured (ordinary datum ladders do not close chains) and records that a planner-side rule
is not ruled out.

Two other real findings from review are filed and stand on their own: **#960** (slot patterns
emit pitch but no origin location, so the array is unlocated, and nothing lints it) and the
`_suggest_fix` bug (it called `drop()`, which takes a feature, where the by-name verb is
`remove()`) — fixed in this branch.

pzfreo:
**Parked for 0.4.0** (maintainer, 2026-08-02) — not abandoned. Deliberately out of the release; pick up first after the tag.

Recording the state so the red checks are legible rather than looking like rot:

**The two failures are real, not stale.** From run 30607152261:

1. `test_drawing_privates_set_is_current` — `_redundancy_entries` was added to `Drawing` without a roster entry. Worth more than a one-line fix now: **#986 (merged today) deleted seven `Drawing` compat aliases** precisely to stop state accumulating on the state bus, and `test_the_expired_compat_aliases_stay_deleted` guards their absence. A new `Drawing` private arriving the same week wants a justification — or an owner other than `Drawing` — not just a roster line.
2. `test_auto_drawing_defines_pad_footprints_and_pocket_locations` — the coverage score drops `1.0 → 0.85`. That is a recognition-completeness regression, plausibly the same pad/mis-recognised-slot interaction this PR itself filed as #958.

**Also 16 commits behind main**, and that gap matters: #720 removed the bare dimension-role spelling at 0.4.0 (#991), so any `dimension(f, "width")` on this branch now raises. Rebase before re-running, don't just re-trigger CI.

**What survives regardless of the code**: the measured answer to the issue's open question — the planner *cannot* over-dimension (common-datum spans can't close a chain), the motivating pattern example doesn't occur, and only the materialised `measured_dimension` / AP242 PMI path can produce redundancy. That analysis is in the PR body and is the reusable part.

To resume: rebase onto main → decide where `_redundancy_entries` belongs → investigate the 0.85 coverage drop (probably #958) → re-run.

pzfreo:
Closing this stale, conflicting draft rather than carrying its implementation into the new delivery plan. The product need remains open in #941 and the provenance prerequisite remains in #961; restart from current `main` when that outcome is promoted.

Issue #964, opened 2026-07-31
Epic: make generated Sheet scripts semantically faithful, complete, and documented


pzfreo:
Successor/finalisation epic for #942. The editable dimension surface and single Sheet emitter now exist (#938 and #940), but “the script runs and draws the same annotation names” is not yet a strong enough round-trip contract. #962 demonstrates the gap: the emitted pocket is reconstructed in the wrong place while annotation parity still passes.

## Outcome

A generated Sheet script is a faithful, editable representation of the detected drawing intent—not merely a program that happens to produce the same annotation-name set on today’s fixtures.

## Main lane — proposed order

| Order | Issue | Why / completion gate |
|---:|---|---|
| 1 | ~~#962 — preserve pocket depth-axis position~~ | ✅ **Merged 2026-07-31** (PR #967). The `at=` fix was the small part: it also delivered the **model-fidelity oracle** item 2 asked for, and found six further emitter defects plus #969 on the way — slot and both pattern-member templates dropping `at=`, through-hole depth dropped in two places, pattern `direction`, a falsy-zero `angle` guard at three sites, and PMI routed as having no round-trip obligation while `_raw_pmi_line` serialises it. |
| 1b | #968 — expose the built Drawing before export | No ordering dependency; can land any time. Every consumer of a generated script currently reaches around `sheet.export(...)` — the e2e round-trip regex-rewrites the tail, the fidelity oracle slices the source at it. Landing this simplifies #948's fixtures. |
| 2 | #948 — complete round-trip coverage | **The oracle now exists** (#962): two corpora, route-driven obligations derived from their enforcement, fail-closed against the IR, one shared structural comparator, lint parity on `(code, severity)`. Fill the fixtures against it. Original note: **strengthen the oracle before filling the eight fixtures**, or they are written against one already known to be insufficient. On #962 annotation-set parity passed, the drawing was identical, and the only signal was an incidental lint divergence. The oracle needs *declared model vs detected model*, field by field, plus asserted lint parity — not just the drawings they produce. |
| 2b | #969 — one grid convention | **Next after #962 merges.** Detection reports `grid[0]` as the X pitch with `rows` along X; `declare._pattern_members` lays `cols` across X using `grid[1]`, so a declared grid array is transposed. The hole verb masks it by emitting explicit `members=`; pocket/slot recompute and expose it. Direction: make detection conform to declaration. #967 records it as a strict xfail rather than translating at the emitter, so this PR chooses the convention with nothing to unpick. |
[...]

PR #967, opened 2026-07-31
fix(#962): carry at= into emitted recesses, and check the model not just the drawing

### LINKED ISSUE #962: The emitted sheet.pocket(...) line drops at=, so the declared pocket loses its depth-axis position and lint diverges from the direct build
## Summary

`emit_sheet_script` writes a pocket without `at=`:

```python
pocket1 = sheet.pocket(width=20, length=30, depth=5, long_axis="x", width_axis="y",
                       lo=-50, hi=-20, w_center=0)   # pocket 20 × 30 × 5 deep
```

`lo`/`hi`/`w_center` fix the pocket in its two in-plane axes, but nothing carries its
position along the **depth** axis, so the declared feature falls back to the origin:

```
direct   pocket origin = (-35.0, 0.0, -5.5)
scripted pocket origin = (-35.0, 0,   0.0)     <-- z lost
```

Every other field round-trips exactly (`field diffs: {}`); only the frame origin differs.

## Symptom: the same part lints differently through the two paths

```
direct   lint: {}
scripted lint: {'unrecognised_defining_geometry': 1}
           warning — dimension-relevant source geometry is absent from recognised IR:
                     1 bounded blind recess(es)
```

**Annotation-set parity holds** (`annotation parity: True`, no names on either side only), which
is why the existing pocket round-trip tests pass. The drawing is the same; the *lint* is not.

## Root cause

`linting/coverage.py::pocket_owner` matches a recognised pocket to a model pocket on its
dimensions *and its location*. With the origin defaulted, no model pocket matches the
recognised one, so `missing_ir += 1` and the recess is reported as absent from the IR.

Confirmed by patching the emitted line by hand:

```
as emitted           origin=(-35.0, 0, 0.0)     lint={'unrecognised_defining_geometry': 1}
with at= restored    origin=(-35.0, 0.0, -5.5)  lint={}
```

## Reproduce

```python
from build123d import Align, Box, Cylinder, Pos, export_step
part = Box(120, 80, 16)
part += Pos(0, 0, 8) * Cylinder(12, 10, align=(Align.CENTER, Align.CENTER, Align.MIN))
part -= Pos(0, 0, 0) * Cylinder(5, 80)
part -= Pos(-35, 0, -8) * Box(30, 20, 10)      # blind pocket in the underside
export_step(part, "bracket.step")
```

Then compare `build_drawing("bracket.step").lint_summary()` against running
`generate_sheet_script("bracket.step")`'s output.

## Why it matters beyond the lint

The lint divergence is the visible symptom; the real defect is that the emitted script declares
a pocket **in a different place** from the one detected. It happens not to change this drawing
because the pocket callout is placed from the projected geometry, but the declared model is not
a faithful round trip, and a consumer of `sheet.model()` gets the wrong z.

It also means the round-trip guarantee is weaker than the tests assert: annotation-set parity
is necessary but not sufficient, and this is a case where it passes while the model diverges.
Worth considering whether `TestRoundTripParity` should compare the declared model against the
detected one as well as the drawings.

## Fix

Emit `at=` on the pocket line (the emitter already does this for other kinds — `sheet.hole`,
`sheet.groove`, `sheet.step_level` all carry a position). Check `sheet.slot` for the same gap:
it shares the lo/hi/w_center shape.

## Found by

Rendering a sample bracket by both routes while demonstrating the #940 single-emitter output.

## Related

- #940 — the Sheet script is now the only emitter, so its fidelity is the only fidelity
- #707 / #810 — the script↔direct parity family
- #898 — the `unrecognised_defining_geometry` lint that surfaces it

pzfreo:
Closes #962. First item of epic #964.

## The bug

`sheet.pocket(...)` was emitted without `at=`, so `declare.pocket` synthesised an origin from
`lo`/`hi`/`w_center` with the **depth-axis** component left at zero:

```
direct   pocket origin = (-35.0, 0.0, -5.5)
scripted pocket origin = (-35.0, 0,   0.0)     <-- z lost
```

Every other field round-tripped exactly. The drawing was identical, because a pocket callout is
placed from the projected geometry rather than from the declared frame.

`sheet.slot` had the same gap, and so did **both pattern member templates**, whose docstrings
claimed "its own position is irrelevant (the pattern's `at=` fixes the array centre)". True for
the two in-plane axes; not true for the depth axis. Adding `at=` there cannot move an array —
`declare.pocket_pattern` takes its centre from the pattern's own `at=` and falls back to the
member's only when that is absent — and I verified the computed member layouts are unchanged.

## The oracle matters more than the fix (epic #964, item 2)

`TestRoundTripParity` compares the **drawings**, and it passed throughout this bug. The only
signal was an incidental lint divergence, which fired because `coverage.py::pocket_owner`
happens to match on location. Luck, not coverage.

So this adds `TestTheDeclaredModelMatchesTheDetectedOne`: emit a detected model, run the script,
compare the model it declares **field by field** against the one it came from, across eight
fixtures. The tolerance is sized to the emitter's 3-dp rounding and nothing more — smaller fails
on every float, larger stops noticing drift.

**Canary: 5 of the 8 fail on `main`.**

## It found a second divergence immediately

A **through** hole's measured depth was dropped, leaving `depth=None` where detection had `8.0`.
Now emitted. It changes no annotation (verified: identical annotation sets either way), and it
deliberately does not go through `.depth()`, which would also set `through=False` and invert the
very fact #878 exists to preserve.

## One exemption, argued and proven

A lone hole's `members` is exactly `(frame.origin,)`, so omitting it loses nothing and emitting
it would add noise to every hole line. Rather than a skip list, the test **asserts the
equivalence holds** for each fixture that uses it — so the exemption cannot quietly start
covering a pattern, where `members` carries positions the frame cannot.

## ADR fit

Serves ADR 0011's round-trip contract (recognise + emit + declare) and ADR 0016's declared-intent
surface: the artefact must be a faithful representation of the model, not merely a program that
reproduces today's annotation names.

## Verification

- `ruff check` / `ruff format --check` / `mypy` / `codespell` clean
- smoke: 51 passed
- `test_sheet_emit` + `test_sheet_identity_invariant` + `test_pocket_pattern` +
  `test_slot_pattern` + `test_script_detail_parity`: 388 passed
- CI owns the full matrix

### DIFF
--- a/src/draftwright/sheet_emit.py
+++ b/src/draftwright/sheet_emit.py
@@ -118,6 +118,12 @@ def _member_hole_str(m) -> str:
         if m.depth is not None:
             kw.append(f"depth={_n(m.depth)}")
         kw.append("through=False")
+    elif m.depth is not None:
+        # A THROUGH member's measured depth is a fact too, exactly as on the standalone verb
+        # — and this template was the sibling that still dropped it after `_hole_line` was
+        # fixed, which the model-fidelity oracle missed because its corpus carried no hole
+        # pattern (#967 review). The two templates diverging is the recurring shape here.
+        kw.append(f"depth={_n(m.depth)}")
     return f"hole({', '.join(kw)})"
 
 
@@ -298,8 +304,16 @@ def _feature_line(f) -> str:
             parts.append(f"bcd={_n(f.bcd)}")
         elif f.pattern == "linear" and f.pitch:
             parts.append(f"pitch={_n(f.pitch)}")
+            if f.direction:
+                # Redundant for the LAYOUT, since `members=` below is spelled out — but it is
+                # a field on the feature, and leaving it None made the declared pattern differ
+                # from the detected one (#967 review). The emitted script is a representation
+                # of the model, not only a program that reproduces its positions.
+                parts.append(f"direction={_pt(f.direction)}")
         elif f.pattern == "grid" and f.grid:
             parts.append(f"grid=({_n(f.grid[0])}, {_n(f.grid[1])}), rows={f.rows}, cols={f.cols}")
+            if f.angle:
+                parts.append(f"angle={_n(f.angle)}")
         if f.members:
             parts.append("members=[" + ", ".join(_pt(p) for p in f.members) + "]")
         return f"sheet.pattern({_member_hole_str(f.member)}, " + ", ".join(parts) + ")"
--- a/src/draftwright/sheet_emit.py
+++ b/src/draftwright/sheet_emit.py
@@ -86,6 +86,13 @@ def _hole_line(f) -> str:
     # an emitter) may not infer an engineering fact from a dimension's presence.
     if not f.through and f.depth is None:
         kw.append("through=False")
+    elif f.through and f.depth is not None:
+        # A THROUGH hole's measured depth is a fact too — it is the bore length through the
+        # material — and dropping it left the declared feature with `depth=None` where
+        # detection had 8.0 (#962, epic #964's model-fidelity oracle). It changes no
+        # annotation: the callout reads THRU either way, checked. `.depth()` is not usable
+        # here because it also sets `through=False`, which would invert the very fact above.
+        kw.append(f"depth={_n(f.depth)}")
     line = f"sheet.hole({', '.join(kw)})"
     if not f.through and f.depth is not None:
         line += f".depth({_n(f.depth)})"  # sets through=False too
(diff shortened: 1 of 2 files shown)


pzfreo:
Adversarial review of `9aac6d2`

The product fixes look coherent and `git diff --check` is clean, but I found two gaps in the new fidelity oracle before this is ready:

1. **[P1] Raw `PmiFeature` is incorrectly routed as having no round-trip obligation.** `_feature_line` does not merely comment or drop this kind: `_raw_pmi_line` actively serializes every raw PMI record as `sheet.add(PmiFeature(...))`. That makes it an emitted declaration with exactly the same fidelity obligation as `authored_dimension`. Yet `_FIDELITY_ROUTE` classifies it as `unnameable`, whose handler positively requires that neither corpus produce it. The existing AP242 test only checks that the source contains `sheet.add(PmiFeature(`; it never executes that line and structurally compares the rebuilt feature. A lossy change to `_raw_pmi_line` therefore remains green while the new machinery claims every emitted kind is covered. Add a synthetic raw `PmiFeature` to the declared corpus, route `pmi` through `declared`, and compare it with `_structurally_equal`. “Raw fallback” describes the API quality, not an exemption from serialization fidelity.

2. **[P2] The lint-parity oracle compares only code counts, not lint parity.** `test_the_script_lints_the_same_as_the_direct_build` compares `lint_summary()["by_code"]`. Two paths can have identical codes/counts while disagreeing on severity, affected annotation, message, or suggested repair. In particular epic #964 explicitly requires code/severity parity, and the test docstring currently overstates what is asserted. At minimum compare a multiset of `(code, severity)` (and the aggregate error/warning/info counts); ideally normalize and compare the stable semantic issue fields while deliberately excluding presentation-only text if necessary.

CI’s platform test matrix and lint are green. `codecov/patch` is still failing at 60% branch coverage, reporting four partial branches in `sheet_emit.py`; that also needs to be green before merge. I started the targeted `test_sheet_emit.py` suite locally and observed no failure in the portion run, but stopped rather than duplicate the already-green GitHub matrix.

pzfreo:
On splitting out #969: **yes, the implementation belongs in its own PR**, because it requires choosing one public grid convention and has a wider blast radius than #962. But I would change how #967 records the gap.

The current test positively asserts broken behavior:

```python
assert sorted(rebuilt.members) != sorted(original.members)
```

That is not a fidelity oracle: it stays green if a later change produces a completely different wrong layout. Instead, write the desired equality assertion and mark those two grid cases `xfail(strict=True, reason="#969")`. That gives the right outcomes:

- the known defect remains visible rather than silently excluded;
- a different wrong transpose still fails for the same precise reason;
- fixing #969 produces an XPASS failure that forces removal of the marker and promotion into the normal corpus.

Recommended sequence:

1. Keep the #969 implementation separate.
2. In #967, record it as strict expected failure, not a passing assertion that the models differ.
3. Resolve the raw-PMI and lint-parity review findings and get Codecov green.
4. Land #969 immediately afterwards and remove the xfails.

For #969 itself I lean toward making detection conform to declaration: rows vary along Y, columns along X, and `grid=(row_pitch, column_pitch)`. That matches ordinary row/column terminology and the declaration implementation’s documented vocabulary. Given the alpha status, deleting the competing detector convention is preferable to permanent emitter translation.

Also please update #969’s body: it currently says PR #967 takes emitter translation as a stopgap, while the latest commit correctly says that translation was reverted and the defect is only recorded.

Issue #1002, opened 2026-08-02
A placed annotation carries no measurement identity, so two builds cannot be compared reliably


pzfreo:
Found reviewing #1001 (the differential audit harness, #996 WP1 step 2).

## The gap

A placed annotation carries **no measurement identity**. Its name — `m_locx0`, `m_env_width` — is an engine-assigned registry slot, not a handle on what it measures.

Demonstrated: build the same plate twice with one hole moved from X=20 to X=40.

```
before: m_locx0 = 70
after : m_locx0 = 90
```

The `70` measurement is gone and a **different** measurement has reused its slot. Nothing in the drawing records that, so a differential can only report "the label changed" — which is true, and weaker than the fact.

## Why it matters beyond the harness

Two consequences, both currently unfixable at the consumer:

1. **A replaced measurement hides as a changed one.** Any tool comparing two builds — an audit, a regression harness, the maquetto eval scoreboard — cannot tell "this measurement moved" from "this measurement went and another arrived".
2. **A loss cannot be attributed to a suppression.** `Drawing.suppressions()` names the *feature* a rule declined; the annotation does not name its feature. So linking "this dimension vanished" to "this rule removed it" is guesswork. #1001 initially matched them by parameter-name substring, which silently excused a lost slot width with an unrelated envelope suppression — an alarm cancelled by a coincidence of strings. That match is now a non-binding hint precisely because it cannot be made reliable.

## What would close it

ADR 0010's provenance seam already records `intent → [annotation names]` at the render seam. What a consumer needs is the inverse and a little more: given a placed annotation, which `DimensionId` / feature / parameter produced it — exposed on the public read surface, as plain data, alongside the label.

`ApprovedDimension` already carries `id: DimensionId | None` and `ref: FeatureRef | None`. The information exists at compile time; it does not survive to the sheet.

## Not urgent, but it caps what the audit can be

#1001 ships as a **triage aid** rather than a detector, and says so in its own docstring, because of this. Closing this issue is what would let it become a reliable one — and would also let the completeness check (#996 WP1 step 3) key on measurement identity rather than on annotation names.

Refs #996, #999, #1001, ADR 0010.

pzfreo:
Narrowing this, from the #1001 review.

**Feature provenance already exists — partially.** `AnnotationRegistry.feature_of(name)` returns the owning feature for annotations the render passes tagged. Measured on a real build:

```
m_cm0        -> HoleFeature
hc_plan0     -> HoleFeature
[...]

Issue #996, opened 2026-08-02
Epic: manufacturing-completeness — a drawing that omits a defining dimension must not lint clean


pzfreo:
## Status — seventh artifact complete 2026-08-06

`wheel_rh.step` exposed one more false-confidence mode after the six-case closeout: automatic output contained only the envelope and scored 1.0 despite a proved double-D through-bore and a 292-face toothed body.

Delivered:

- #1058 pins the exact STEP evidence and names the bounded contract.
- #1059 adds a physical principal-boundary critique for internal profiles outside the current circular/rectangular/obround/rounded-rectangular vocabulary. The wheel now emits `unrecognised_defining_geometry` and scores below 1.0.
- #1060 covers every classifier rejection decision after delayed Codecov results exposed an incomplete test matrix. Lint, Codecov patch/project, and all 15 OS/Python jobs are green.

This is deliberately a **fail-loud completeness fix**, not a claim of feature support. The model still contains only `envelope`; gear-tooth and double-D recognition remain separate recognition work if a drawing needs them represented. They do not keep this manufacturing-confidence epic open because the engine no longer certifies the omission as complete.
---

## Status — complete 2026-08-06

All six supplied case studies are closed against pinned or directly verified artifacts.

| case | outcome |
|---|---|
| #918 square envelope | independent square size is planned and deletion is reported |
| #914 flat A/F | real artifact retains its only A/F definition with provenance |
| #915 dense prismatic | shared callout solve plus truthful per-level support recovers the full A2 detail |
| #916 blind pocket | callout exits from the opening rim without crossing the silhouette |
| #917 U-channel | recognised as an open channel with a non-redundant defining chain |
| #909 sloped/padded wedge | pad footprint is recognised and crowded height recovery is automatic, disjoint and complete |

The implementation did not validate the original generic WP sequence. The cases required distinct small contracts: planner selection, boundary-aware placement, thin recognition records, family-specific correspondence and detail policy. ADR 0017 therefore remains accepted only for recognition ownership; its proposed global identity types, shared requirements module, outcome ledger and named reconciliation stage were not built. Scoped follow-ups such as #1012, #1032 and #1037 remain independent and do not keep this case-study epic open.

---

## Original hypothesis and work plan

Five real-part case studies filed 2026-07-28 (#909, #914, #916, #917, #918), plus #915, share one signature:

> **Four of them produce clean lint on a drawing you could not manufacture from.**

[...]

PR #1007, opened 2026-08-03
feat(#996): per-feature completeness check — BLOCKED on #1009

pzfreo:
**Draft — do not merge. Blocked on #1009.**

WP1 step 3 of #996: for every recognised feature, is a defining size drawn and a location? Requirements from the recognisers (never the IR — the ADR 0015 carve-out), coverage from the #1002 measurement identity, three verdicts (`covered` / `uncovered` / `unchecked`).

## Why it is blocked

Two adversarial rounds found **six distinct fabricated-defect classes** — the checker inventing defects rather than finding them. Full table on #1009. The two P0s were each reproduced on independent real parts:

- **patterns**: the compiler collapses 3+ members into one `pocket_pattern`/`slot_pattern`, while the checker lists members and requires an exact kind match — so a fully dimensioned array of three slots yields **three fabricated defects**;
- **off-centre slots**: a lone `SlotFeature.frame` sits at the part bbox centre on the transverse axes by design, while the checker anchors on the true centroid — so **no off-centre slot can ever match**. The slot test in this branch used a *centred* subtraction, so it passed by construction.

There is also a false negative that is **inherent, not a gap**: a pocket's X and Y location dims share one identity, so deleting either still reads `location='covered'`. That needs directional identity (#883).

## The root cause

The check hand-reconstructs a correspondence between recogniser records and compiler identities. That correspondence is genuinely non-trivial — grouping, deliberate frame conventions, cardinality — and every approximation of it has invented defects. **#1009** specifies what is actually needed.

## What is sound and worth keeping

- coverage read from measurement identity rather than witness geometry (the first cut matched witnesses and called a fully-dimensioned pocket undefined — its size is a leader, which has no witnesses);
- the three-verdict polarity: `unchecked` is never folded into either side;
- the fail-closed id ratchet, which catches a required id the compiler never mints;
- per-record requirements, so an edge-anchored pocket is not asked for location dims it correctly lacks.

## Also withdrawn

The blast-radius measurement reported on #996 ("12 judged, 0 undefined") **is withdrawn** — it was produced by this instrument, and neither a slot nor an edge-anchored pocket was in the sample. #996's "measure before choosing severity" step is still open.

## Related

#1008 (the kind inventory covers 9 of ~17), #1009 (the correspondence model), #883 (per-member location identity).

### DIFF
--- a/src/draftwright/linting/__init__.py
+++ b/src/draftwright/linting/__init__.py
@@ -16,6 +16,11 @@ Import the public surface from here, not the submodules.
 
 from __future__ import annotations
 
+from draftwright.linting.completeness import (
+    FeatureVerdict,
+    completeness_summary,
+    feature_completeness,
+)
 from draftwright.linting.coverage import (
     CoverageState,
     lint_axial_coverage,
@@ -31,8 +36,11 @@ from draftwright.linting.suggest import _suggest_fix
 
 __all__ = [
     "CoverageState",
+    "FeatureVerdict",
     "LintIssue",
     "_suggest_fix",
+    "completeness_summary",
+    "feature_completeness",
     "lint_axial_coverage",
     "lint_boss_height_coverage",
     "lint_declaration_reconciliation",
--- a/src/draftwright/linting/completeness.py
+++ b/src/draftwright/linting/completeness.py
@@ -51,23 +51,52 @@ COVERED = "covered"
 UNCOVERED = "uncovered"
 UNCHECKED = "unchecked"
 
-#: kind -> (size parameter ids, location parameter ids). A recognised kind absent from here
-#: answers ``unchecked``, by design — see the module docstring.
+#: kind -> (size parameter ids, location parameter ids) for a feature with nothing special
+#: about it. A recognised kind absent from here answers ``unchecked``, by design.
 #:
-#: The ids are the compiler's own `parameter_id`s, read off real builds rather than guessed,
-#: because a typo here would silently mean "never covered" and read as a real finding.
-_REQUIRED: dict[str, tuple[frozenset, frozenset]] = {
+#: The ids are the compiler's own `parameter_id`s. Every one of them is asserted to appear in
+#: a real build's ledger by `test_every_required_id_is_one_the_compiler_actually_emits` —
+#: because the first cut wrote `location_slot.location` where the compiler emits
+#: `location_slot.length`, which made EVERY slot report its location missing, and nothing
+#: caught it. A hand-maintained string table without a ratchet is an unenforced claim.
+#:
+#: (The two location ids disagree in form — `location_pocket.location` against
+#: `location_slot.length` — which looks like an engine inconsistency rather than a deliberate
+#: distinction. Recorded as a side issue rather than papered over here.)
+_BASE_REQUIRED: dict[str, tuple[frozenset, frozenset]] = {
     "pocket": (
         frozenset({"pocket_width.length", "pocket_length.length", "pocket_depth.length"}),
         frozenset({"location_pocket.location"}),
     ),
     "slot": (
         frozenset({"slot_width.length", "slot_length.length"}),
-        frozenset({"location_slot.location"}),
+        frozenset({"location_slot.length"}),
     ),
     "chamfer": (frozenset({"chamfer.length"}), frozenset()),
     "fillet": (frozenset({"fillet.radius"}), frozenset()),
 }
+
+
+def _requirements(kind: str, feat):
+    """What *this* feature needs — a function of the RECORD, not of its kind.
+
+    Kind alone cannot express a per-feature requirement, and that is not academic: a pocket
+    whose walls are flush with the stock edges is located by those edges, so the planner
+    correctly emits no location dimension for it. A kind-only table demanded one anyway and
+    reported a correct drawing as defective (Codex #1007 r1) — the same fabricated-defect
+    class this module has now produced four ways.
+
+    ``edge_anchored`` is a **recogniser** fact, sitting on the record, so consulting it is not
+    the plan-sourcing the ADR 0015 carve-out forbids. That is the distinction: reading what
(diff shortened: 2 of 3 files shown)


pzfreo:
## Codex review: do not merge. Confirmed — converting to draft.

Six findings, two of them false defects Codex **reproduced on real parts**. I verified both rather than taking them on trust; both are worse than they look.

### 1. Every slot reports its location missing

```
slot ids drawn : ['location_slot.length', 'slot_length.length', 'slot_width.length']
_REQUIRED says : location_slot.location
verdict        : size='covered'  location='uncovered'
```

The table has the wrong id, so **no slot can ever read covered**. The docstring above that table says the ids were *"read off real builds rather than guessed, because a typo here would silently mean 'never covered' and read as a real finding"*. I wrote that warning and then committed exactly that typo — my vocabulary probe used a part whose slot was recognised as a **pocket**, so I never saw a slot's ids at all and wrote the row by analogy.

### 2. Edge-anchored pockets are required to have location dims they correctly lack

```
recognise_pockets(...)[0].edge_anchored = True
verdict = size='covered'  location='uncovered'
```

The planner deliberately omits location dimensions for a pocket the stock edges already locate. `_REQUIRED` keys on **kind alone**, so it demands them anyway. This is the structural finding: my claim that requirements *"come from recognition semantics"* is false — they come from a kind-only table that cannot express a per-feature requirement, while `Pocket.edge_anchored` is sitting right there on the record.

### 3–6, all accepted

- the join reduces `feature_key` to kind + origin and rejoins within a scale-independent 0.6 mm cube, so two nearby same-kind features share the union of their ids — the anchor-collapse defect moved from key generation into matching;
- `_inventory` claims "every recognised feature kind" and omits patterns, countersinks, plates, face levels, step shoulders and turned steps, so the *unchecked* population is not real either — and it reuses only five `Analysis` inventories, so the one-inventory claim is overstated too;
- `_by_kind` counts size and location separately while the API describes feature counts, and the test asserts `summary["by_kind"]` against itself — tautological, can never fail;
- the tests cover only the two fixtures I had already fixed, not the advertised contract.

### Why I am not patching this

That is **four** distinct false-defect classes in one module: witness geometry, anchor collapse, a wrong parameter id, and kind-only requirements. Each fix has been sound and each time the next probe found the same shape underneath. Per the repo's own rule — the same class of bug twice means stop and reconsider the approach — patching finding 1 and 2 would just move it again.

The real design fault is that I hand-wrote a mapping that must be **derived**. Two things follow, and both fit inside the ADR 0015 carve-out:

1. **Requirements are a function of the recogniser RECORD, not of its kind.** `edge_anchored` is a recogniser fact, not a planner decision, so reading it is not the circularity the carve-out forbids.
2. **The id vocabulary must be fail-closed.** A test that asserts every id in `_REQUIRED` appears in some real build's ledger would have caught `location_slot.location` before it was written. A hand-maintained string table with no ratchet is the same unenforced-claim pattern this epic keeps punishing.

Plus: an exact join on the full `feature_key`, and an inventory sourced authoritatively rather than by hand.

Draft until that rework lands. The blast-radius number I reported on #996 ("12 judged, 0 undefined") is **withdrawn** — it was produced by this instrument, and findings 1 and 2 mean it was measuring the wrong thing for at least slots and edge-anchored pockets.

pzfreo:
## Codex round 2: do not merge. Stopping here rather than attempting a third design.

Two P0s, each reproduced on independent real parts, and both invalidate the approach rather than an instance:

**Patterns.** The compiler collapses 3+ members into one `pocket_pattern`/`slot_pattern` feature. `_inventory` lists individual members and the join requires an exact kind match, so a fully dimensioned array of three slots yields **three fabricated defects**. Worse than the omission the docstring admits to.

**Off-centre slots.** A lone `SlotFeature.frame` is deliberately placed at the part bbox centre on the transverse axes (`model/detect.py`); the checker anchors on `Slot.location`, the true centroid. So **no off-centre slot can ever match** — verified at 5, 8 and 10 mm offsets. The slot test I added last round used a **centred** subtraction, so it passed by construction and could not have caught this. That is the same self-confirming shape I have been criticising all along, in my own test.

**And a false negative that is inherent, not a gap:** a pocket's X and Y location dims share one identity, so deleting either still reads `location='covered'`. The pocket is under-located and the check says fine. That needs directional identity (#883), not a patch.

Plus P1–P3 accepted: the kind list is still partial; "unchecked when identity is unavailable" is not implemented at parameter level; the verdict key is weaker than the ledger key it summarises; and `test_unchecked_is_never_folded_into_either_side` still contains the self-referential `by_kind` comparison — I added a real test beside it instead of removing the misleading assertion, which was the wrong call.

## Why I am stopping

That is **six** distinct fabricated-defect classes across two review rounds. Each fix was sound; each time the next probe found the same shape underneath. The repo's rule is that the same class twice means reconsider the approach, and the cause is now clear: I am hand-reconstructing a recogniser↔compiler correspondence that is genuinely non-trivial — grouping, deliberate frame conventions, cardinality — and approximating it keeps inventing defects.

Raised as **#1009** with the design requirement. A third attempt at the same shape in the same session would very likely produce a seventh class.

This PR stays in **draft**. What is sound and worth keeping: identity-based coverage rather than witness geometry, the three-verdict polarity, the fail-closed id ratchet, and the per-record requirements. What is not: the correspondence underneath all of it.

**#996's blast-radius number remains withdrawn.**

pzfreo:
## Superseded after ADR 0017 Gates 1-3

Closing this draft without merge. Its adversarial findings were valuable and directly shaped #1018, but the implementation's universal inventory/join is the approach those findings invalidated.

#1042 landed exact flat correspondence and #1044 landed exact lone-slot/pattern correspondence, including the P0 off-centre and N:1 cases plus directional deletion detection. Both preserve the sound parts of this draft: semantic measurement provenance, fail-closed unverifiable states, and distinct outcomes.

Gate 3 found no evidence for a general checker, global source id or shared requirements policy. User-visible completeness work continues as scoped #996 slices.

Issue #1005, opened 2026-08-03
Side-hole location candidates record no measurement identity


pzfreo:
Split out of #1002; narrowed after #1044.

## Completed

- [x] Pattern pitch/grid placement now records directional `DimensionId` provenance, including corridor, legibility and drop paths (#1044).
- [x] The corresponding exemption was removed from `tests/test_audit_differential.py`; stale exemption removal is ratcheted.

## Remaining

`annotations/holes.py::_off_axis_queue` still draws side-hole location corridor candidates without a `measurements=` map threaded from its callers.

Consequence: a lost or same-name-substituted side-hole location reads as “nothing claims it” in `draftwright.audit`.

## Acceptance

- Thread the existing semantic measurement identities into `_off_axis_queue` candidates.
- Retain those identities on every placement/drop path.
- Remove the remaining `_off_axis_queue` exemption from `test_every_direct_placement_records_identity_or_says_why_not`.
- Prove with a targeted loss/substitution mutation; do not infer from labels or coordinates.

Related outcome epic: #996. #1044 is the completed pitch/grid half.

Issue #966, opened 2026-07-31
The dimension vocabulary has no canonical registry: a Literal kept in step by AST-scraping the DimParameter call sites


pzfreo:
## Summary

#963 typed `Sheet.dimension(feature, role)` with `DimensionRole = Literal[...]`, and keeps that
list honest with `tests/test_dimension_role_vocabulary.py`, which walks `model/ir.py`'s AST and
collects every `DimParameter(kind, role, …)` construction.

That works today and is bidirectional (nothing missing, nothing stale), but it is **test-time
source scraping standing in for a registry**. It assumes every vocabulary member appears as a
syntactically simple `DimParameter(...)` call, in one particular file. Harmless refactors break
it for reasons unrelated to the API:

- extracting a helper or factory that builds parameters;
- building them in a loop or comprehension;
- moving a feature type to another module;
- naming the arguments (`DimParameter(kind=…, role=…)` is not matched by the positional walk).

The `len(pairs) > 20` guard-the-guard assertion catches a *total* extraction failure, but not a
partial one — the failure mode is a vocabulary member silently going unlisted, whose symptom is
a caller told a valid measurement is invalid.

## Options

1. **A data-driven vocabulary in production.** Have the IR expose its parameter inventory as
   data (a module-level table or a registry the feature types populate), so both the runtime and
   the `Literal` derive from one place instead of one scraping the other. This is the same shape
   as `model/detect.py`'s typed record→`Feature` converter registry (ADR 3 (was 0013) roadmap 1c),
   which is fail-closed by `tests/test_detect_registry.py` — a precedent worth copying.
2. **Generate the `Literal` from an explicit schema** as a build/check step, so the artefact is
   derived rather than hand-maintained-and-verified.
3. **Keep the hand-maintained list, drop the AST.** A plain table of supported public ids plus a
   runtime test that every listed id resolves on some fixture is clearer than pretending AST
   extraction is a durable source of truth.

(3) is the cheapest honest option; (1) is the right end state and fits how this codebase handles
its other closed rosters.

## Also worth settling: the naming

The review of #963 (PR #965) noted that the change leaves terminology debt:

- `DimensionRole` contains parameter **ids**, not roles;
- `roles()` returns parameter **ids**, not roles;
- the parameter is still called `role`.

The suggested end state was `DimensionParameterId` and `handle.dimension_ids()`. Both names are
new in #963, so renaming is free *now* and a compat burden later — but `DimensionRole` is the
[...]

Issue #741, opened 2026-07-19
Test reach-through, round 2: ~149 dwg._* attribute reads outside the #720 alias set


pzfreo:
Follow-up in the #720 family, from the #733/#734 review discussion. #721 migrated the seven aliased privates (361 sites) to the public read surface, but ~149 `dwg._*` attribute reads remain in tests, coupled to internals with **no public read equivalent yet**:

- `dwg._analysis` — 54 sites (the build-state namespace; widest surface, needs a deliberate public projection or accessor, not a blanket expose)
- `dwg._defer_intents` — 30 sites (deferred-mode toggle; tests drive it directly instead of `deferred()`)
- `dwg._intents` — 22 sites (recorded-intent inspection)
- `dwg._coords` — 7, `dwg._registry` — 5 (public `registry` exists — trivial), `dwg._record_build_issue` — 5, `dwg._build` — 4, `dwg._part_model` — 3 (public `model()` exists — trivial), plus stragglers.

This is the one unpoliced quadrant left: the encapsulation guards police src (whole-engine since #722), and `test_private_test_imports` polices test *imports* from annotations/ — test-side **attribute** reach-through has no ratchet, so these can silently multiply.

Plan (the #721 pattern):
1. Triage which reads need NEW public surface (`_analysis` projection, intent inspection, a `deferred`-state query) vs which map to existing reads (`registry`, `model()`).
2. Design the `_analysis` read surface deliberately — it is the big one and a blanket property would just rename the coupling.
3. Migrate mechanically per group; zero-revert bar as in #721.
4. Add the missing ratchet: a test-side attribute-read guard (shrink-only allowlist, like the src engine guard from #722).

Not release-blocking; sized meaningfully larger than #721 because of the `_analysis` design work.

pzfreo:
Done. Delivered in two PRs:

- **#814 (PR 1)** — the missing **test-side attribute-read ratchet** (`tests/test_private_test_attr_reads.py`): an AST scanner pinning reads of `Drawing` privates to a shrink-only per-name ceiling, fail-closed in all directions (new name / grown count / un-lowered ceiling / new Drawing private), plus the `_registry` → `dwg.registry` migration. Hardened over 3 Codex rounds (whole-class scope-aware completeness scan, AugAssign counting, no self-exemption).
[...]

PR #1010, opened 2026-08-03
feat(#966): a feature declares its location stem (slot/pocket/hole paths)

pzfreo:
First slice of #966. Each feature declares the stem its position is minted under; the compiler that mints the id and the renderers that read it back both derive it from that one declaration.

## What this establishes

Declaration ownership of the location **stem**, proven by renaming the declaration against real builds and asserting both the minted ids and the drawn annotations follow. Every mint site and reader, each verified by mutation:

| site | canaried by |
|---|---|
| slot mint (`model/compiled`) + `from_model` slot reader | `[slot]` |
| Z-normal pocket mint | `[pocket-z]` |
| non-Z pocket mint + `from_model` discriminated-pocket reader | `[pocket-x]` |
| Z-hole and Z-pattern mint (via `planner.location_role`) | `[hole-z]`, `[pattern-z]` |
| side-drilled mint + `holes.py` off-axis filter | `test_the_side_drilled_path_mints_from_its_own_declaration` |
| `render_locations` and `Drawing.locate()` on-axis-bore skip | `test_the_on_axis_bore_skip_reads_the_declaration_on_both_surfaces` |
| exact-type eligibility, all six locatable kinds | `test_a_subclass_is_not_silently_locatable` |

Every row was checked by editing the protected code back to a string literal and confirming the named test goes red. **Four were green against broken code before that check** — which is the point of doing it:

- the discriminated-pocket branch (a Z-normal fixture never reaches it);
- the on-axis skip, on *both* surfaces (a plate fixture never reaches either);
- `_compile_slot_positions` (found by review, not by the guard — see below);
- and the subclass guard itself, whose first rewrite built a `PartModel` with no `datum_xy`, so five of six kinds minted nothing for the **parent** either and every case passed for free.

## What it does not establish

- **The suffixes still differ** — `location_pocket.location` vs `location_slot.length`. The stem is owned; the suffix is still chosen at each mint site. That is the remaining half of #966, and `ir.py` now says so rather than claiming "one declaration site, one spelling".
- The off-axis canary observes compiled ids and drawn annotations, not `measurement_keys`: those dims record no measurement identity yet (**#1005**), so a ledger assertion would pass vacuously.

## The review defects, closed

1. **Undocumented behaviour change (r3).** `isinstance(feature, _LOCATABLE)` accepted subclasses where the `dict[type, str]` it replaced did not. Exact membership restored and documented as deliberate: a subclass inherits `LOCATION_STEM`, so accepting it would mint its position under its parent's name.
2. **Off-axis compilation (r3)** now comes under the declaration, via a second stem (`LOCATION_OFF_AXIS_STEM`) rather than a suffix rule — the bbox-datum, per-axis positions are a different measurement from the Z-normal ladder, so a separate declaration is the honest shape.
3. **The canary's overclaim (r3)** — fixed by widening the evidence rather than the prose, then stating the scope as narrowly as what was mutated.
4. **The `ir.py` comment (r3)** — corrected.
5. **`_compile_slot_positions` bypassed the eligibility rule (r4).** It selected slots with a bare `isinstance`, so a `SlotFeature` subclass the planner refused still minted `location_slot.length` — the exact collision fix 1 claimed to prevent, reached by the one compiler that never asked. Eligibility now comes from `location_datum`, matching `_compile_off_axis_hole_locations`; the guard is parametrised over all six kinds and asserts on the compiled plan.

Two further fixes came from CI rather than review: the on-axis canary read `dwg._analysis` (tripping the #741 private-read ratchet) and now asserts the compiler approved a position instead — a stronger precondition; and `location_datum`'s trailing `isinstance` arm plus `return None`, which exact-type membership made unreachable, is gone. It was the branch's one uncovered patch line, and the narrowing it claimed to give mypy was never needed (the parameter is untyped).

Codex rounds 5 and 6: no findings.

## Verification

Full fast tier under coverage: **2672 passed, 1 skipped, 2 xfailed**; no uncovered line remains in this branch's diff. ruff check, ruff format, mypy, codespell clean.

### DIFF
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -91,6 +91,7 @@ from draftwright.model.ir import (
     AUTHORED_DIMENSION_KINDS,
     HoleFeature,
     PatternFeature,
+    PocketFeature,
     SlotFeature,
 )
 
@@ -195,7 +196,7 @@ def render_slots(dwg, plan, a, *, ctx, only=None) -> int:
     pocket_locations = {
         (loc.ref, loc.discriminator): loc
         for loc in plan.locations
-        if loc.role == "location_pocket" and loc.discriminator is not None
+        if loc.role == PocketFeature.LOCATION_STEM and loc.discriminator is not None
     }
     # A slot's own position dim — datum→near-end along its long axis. Compiled, not
     # computed from `a.bb`: it prints a number, so an authored set that does not name the
@@ -605,7 +606,11 @@ def render_locations(dwg, plan, a, *, ctx, only=None, pinned=None) -> int:
         # filter). A pattern ref (role "location_pattern" — e.g. a bolt-circle
         # centre) is NOT filtered, even on the axis. A renderer-side filter only ever
         # REMOVES an approved entry (a drop), so it does not breach the boundary.
-        if loc.role == "location" and a.is_rotational and _concentric_with_axis(a, rx, ry):
+        if (
+            loc.role == HoleFeature.LOCATION_STEM
+            and a.is_rotational
+            and _concentric_with_axis(a, rx, ry)
+        ):
             continue
         # Provenance (ADR 0010): the located feature. `resolve_feature` is the sanctioned
         # seam for exactly this — the corridor's feature map keys drop()/annotations_of().
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -87,7 +87,12 @@ from draftwright.layout import StripCandidate, plan_strip
 from draftwright.model.callout import _first as _first
 from draftwright.model.callout import hole_callout_spec as hole_callout_spec
 from draftwright.model.compiled import FeatureRef, resolve_feature
-from draftwright.model.ir import AUTHORED_DIMENSION_KINDS, HoleFeature, PatternFeature
+from draftwright.model.ir import (
+    AUTHORED_DIMENSION_KINDS,
+    HoleFeature,
+    PatternFeature,
+    SlotFeature,
+)
 
 
 def callout_from_spec(spec, draft, count) -> HoleCallout | None:
@@ -195,7 +200,12 @@ def render_slots(dwg, plan, a, *, ctx, only=None) -> int:
     # A slot's own position dim — datum→near-end along its long axis. Compiled, not
     # computed from `a.bb`: it prints a number, so an authored set that does not name the
     # slot's location must not get one (#925).
-    slot_positions = {loc.ref: loc for loc in plan.locations if loc.role == "location_slot"}
+    # Derived, not spelled: the name is a CONTRACT between the compiler that mints it and
+    # this renderer that reads it, and until #966 neither end derived it — so renaming the
+    # declaration silently dropped every slot location dim rather than renaming it.
+    slot_positions = {
+        loc.ref: loc for loc in plan.locations if loc.role == SlotFeature.LOCATION_STEM
+    }
     draft = dwg.draft
     tier = draft.font_size + 2 * draft.pad_around_text
     views = {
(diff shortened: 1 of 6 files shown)


pzfreo:
## Codex round 2: do not merge. Draft, and I am stopping here for now.

Four findings, all correct. The two that matter both say the same thing: **my guards do not measure what they claim.**

### The "fail-closed" completeness guard is circular

`test_every_locatable_feature_declares_its_stem` discovers locatable classes by iterating `_LOCATION_ROLE` — the very table under test. The reverse check discovers dataclasses mechanically but then keeps only those *already carrying* `LOCATION_STEM`.

So a new feature routed by a new compiler path, but absent from both the declaration and the table, is **invisible to both tests**. My claim that "adding a locatable feature without a declaration fails here" is false. The mapping's *values* are projected; its *membership* is still a hand-maintained list.

I canaried the uniqueness check and the projection check. I never canaried the completeness claim — which is the one I described as fail-closed.

### The behaviour test proves less than its prose

Mutating `SlotFeature.LOCATION_STEM` after import does not exercise the planner projection at all — `_LOCATION_ROLE` is an import-time snapshot. It passes because the slot's compiler and renderer happen to read the class attribute dynamically. The other five declarations get no equivalent compiled-and-rendered assertion, and finding 2 notes that competing renderer-side spellings for those kinds remain, where the same rename would silently drop a dimension.

So: I fixed the slot path, wrote a test that only exercises the slot path, and described it as establishing canonical ownership generally.

### What it would actually take

1. **Non-circular discovery** — enumerate locatable features from the compiler/renderer paths that mint or read a location, not from the table that is supposed to describe them.
2. **Per-kind behavioural assertions** — the rename canary for all six declarations, compiled and rendered, not slots alone.
3. **Remove the remaining parallel readers** for the other five kinds, which is the substance of finding 2.

That is a real piece of work, not a patch, and I would rather do it fresh than iterate again tonight — this is the third PR today where the guards overclaimed, and every one of those was caught by review rather than by me.

Staying in **draft**. The two commits on the branch are sound in themselves — the slot path genuinely derives its name at both ends now, and that removed a real silent-drop failure mode — but the PR's central claim of canonical ownership is not yet earned.

pzfreo:
## Codex round 3: do not merge. Recording accurately and stopping here.

Four findings. Two are new defects of mine, two are overclaims in prose I wrote.

### 1. I introduced an undocumented behaviour change while calling the PR behaviour-neutral

`type(feature) not in _LOCATION_ROLE` → `isinstance(feature, _LOCATABLE)` is **not** equivalent: isinstance accepts **subclasses**, the dict membership did not. On main a subclass of a locatable feature received no location; on this branch it does, and can compile new dimensions for caller-supplied `PartModel`s.

I changed a semantic and described the change as neutral. Exact membership should be preserved unless subclass support is a deliberate, documented decision — and it isn't one I made.

### 2. A production compiler still bypasses the declared owner

Off-axis hole locations are minted as `location_off_axis.{axis}` in `compiled.py`, independent of `HoleFeature.LOCATION_STEM`. So the declaration is not the owner for that path, and my canary does not reach it.

### 3. The canary still asserts less than its docstring claims

It covers automatic Z-normal builds for hole, pocket and slot only. It does **not** exercise off-axis compilation (which is how finding 2 hides), `Drawing.locate()` through the reader I changed in `holes.py`, or pattern/pad/pocket_pattern. So its "both ends follow the declaration" and the module's "fail-closed" framing remain false as written.

I narrowed the scope claim in the **commit message** last round but left the **test docstring** overclaiming. Fixing the description in one place and not the other is the same error in miniature.

### 4. `ir.py`'s "One declaration site, one spelling" overstates what landed

The suffixes still differ — `location_pocket.location` against `location_slot.length` — because slot suffixes are minted separately while ordinary planned locations take theirs from `DimParameter.kind`. The motivating inconsistency is still there. Acceptable for a first slice; not acceptable as a comment that tells future readers it is solved.

### Status

Staying in draft. Three rounds, and each has found something real — the process is working, but the branch is not ready and I am not going to attempt a fourth rework in this session's state.

What is genuinely sound and worth keeping: the slot path derives its name at both ends (which removed a real silent-drop failure), the snapshot dict is gone in favour of a live read, and the parametrised canary is the right shape even though its coverage is narrower than its prose.

Next: restore exact type membership, bring off-axis under the declaration, and either widen the canary to the paths named above or cut the docstring back to what it actually proves.

Issue #1008, opened 2026-08-03
Completeness check enumerates 9 of ~17 recognised feature kinds


pzfreo:
Split out of #996 WP1 / PR #1007 (Codex r1 finding 4).

`linting/completeness.py::_inventory` enumerates hole, slot, pocket, pad, boss, chamfer, fillet, flat, groove. It omits public recognisers for:

- hole patterns, pocket patterns, slot patterns
- countersinks
- plates
- face levels
- step shoulders
- turned steps

Some of those aggregate into IR features rather than mapping 1:1, so the right treatment differs per kind — but silently excluding them means a part dominated by them produces a **sparse report that looks authoritative**. The first version's docstring claimed "every recognised feature kind"; that claim is now corrected rather than met.

Also: only five of the nine reuse the `Analysis` inventory; the rest re-recognise, so the ADR 0015 one-inventory claim is only partly true.

Wanted: an authoritative, fail-closed source for the kind list — ideally a test that enumerates the `recognise_*` surface and fails when a recogniser exists with no entry here (present or explicitly excluded with a reason), so the omission set can never drift silently again.

pzfreo:
## Gate 3 disposition

Closing this as **not planned**. This issue accurately audited the hand-maintained inventory on draft PR #1007; it does not describe a shipped authoritative completeness surface on `main`.

Phase 1 of #1018 separately made recognition ownership fail-closed for all 17 public families. That does not imply all 17 should be migrated into one semantic completeness checker. The flat and slot/pattern slices proved that physical requirements and correspondence remain family-specific.

Future completeness adapters now attach to user-visible #996 children when a red fixture requires them. A sparse all-family programme is no longer retained as assumed work.

Issue #914, opened 2026-07-28
Flat-across-cylinder drawing can drop the only A/F definition


pzfreo:
## Case study

Current `main` (`a3f3c5f`) was run against `IF_step_flat_across_cylinder.step`. Recognition succeeds:

```text
boss: 1, envelope: 1, step_level: 1, flat: 1, rotational: 1
```

but the defining flat callout is dropped:

```text
flat_dropped: flat callout 25000 A/F not placed (no clear room)
```

The drawing retains the cylindrical diameter and step dimensions but no longer defines the flat. This is a manufacturing-completeness failure, not merely cosmetic placement loss: A/F is the feature's only size parameter.

The source file is unusually large (`40000 × 60000 × 40000 mm`, A0 at 1:200), but the placement invariant must hold at any supported page/scale: a defining feature dimension must either survive, trigger a layout escalation, or produce explicit incomplete-drawing diagnostics.

## Synthetic reproduction

Add a normal-unit build123d fixture so the regression is independent of the real file and its unit convention:

```python
from build123d import Align, Box, Cylinder, Rot
from draftwright import build_drawing

shaft = Rot(90, 0, 0) * Cylinder(
    20, 60,
    align=(Align.CENTER, Align.CENTER, Align.CENTER),
)
# Two opposed flats: true width across flats = 25 mm.
part = shaft & Box(
    25, 60, 40,
    align=(Align.CENTER, Align.CENTER, Align.CENTER),
)

dwg = build_drawing(part, page="A4")
assert any(f.kind == "flat" for f in dwg.model().features)
```

If this compact version does not stress the strip enough, add an adjacent turned step and a small number of competing defining annotations; do not make the test depend on 40-metre coordinates.

## Expected architecture

The flat renderer should emit a first-class defining candidate through the existing collect-then-solve pipeline. A failed preferred corridor should escalate through ranked alternatives: opposite corridor / supporting view / larger page or reduced scale / explicit unresolved coverage. Do not hand-place the callout or special-case this fixture.

## Acceptance criteria

- The synthetic flat is recognised with the correct `25 A/F` value.
- Its A/F definition is present in the final drawing under representative placement pressure.
- It does not overlap the view, other annotations, or sheet furniture.
- If no legal placement exists, lint states that the drawing is manufacturing-incomplete and names the affected flat; a bare placement warning is insufficient.
- The real case study no longer silently loses its only flat definition.
- Existing diameter and step dimensions remain correct.

[...]

Issue #574, opened 2026-07-12
Epic: every recognised feature must round-trip all three ADR-0011 surfaces (recognise + emit + declare)


pzfreo:
## Principle

Per **ADR 0008** (one feature inventory) and **ADR 0011** (three authoring modes: *auto* / *auto+tweak* / *generate*), a feature is only **done** when it round-trips through **all three** surfaces — not detection alone:

1. **Recognise** — detect from geometry → IR `Feature` → auto-pass callout (`make_drawing`/`build_drawing`).
2. **Emit** — `sheet_emit` writes the feature into the generated editable Sheet script (`--script --style sheet`).
3. **Declare** — `sheet_dsl.Sheet` + `model/declare.py` let a caller *author* the feature from scratch / read it off a build123d object.

**Adding recognition alone is tech debt:** the feature is detectable but neither emittable nor declarable, so the *generate* and *declare* modes silently drop it. Going forward, **no recognition-only features** — each feature is completed across all three surfaces in the same PR.

## Current gap

Detection-only today (recognise ✅, emit ❌, declare ❌): **chamfer** (#560), **plate/wall** (#559), **step-position** (#555), **countersink** (#558). `sheet_dsl` declares hole/boss/step/slot/pattern/envelope + GD&T/finish; `sheet_emit` emits hole/pattern/envelope/step_level/pmi.

## Back-fill (recognition-only features already merged / in flight)

- [x] **chamfer** — #576 ✅ (first feature complete across all three surfaces)
- [x] **plate/wall** — #577 ✅
- [x] **step-position** — #578 ✅ (epic back-fill complete — all four features round-trip all three surfaces)
- [x] **countersink** — #575 ✅

## Going forward

The remaining dimensioning issues are updated to require **recognise + emit + declare in the same PR**: #561 (fillet), #148 (blind-pocket), #562 (authored `sheet.dimension()`).

## Notes

- **Emit tier is the cheap win** — these features are already in the detected model; `sheet_emit` just doesn't handle the kinds yet.
- **Declare tier** — a `sheet.<verb>()` on the fluent surface + a `declare.py` object-reader + an explicit-value flavour.

## Lessons learnt (from #576 chamfer — apply to #575/#577/#578)

The chamfer back-fill went through three review cycles; the reusable patterns:

[...]

Issue #1009, opened 2026-08-03
Completeness check needs a recognition↔compiler correspondence model


pzfreo:
Blocks #996 WP1 step 3. Raised from two adversarial reviews of PR #1007, which found **six distinct fabricated-defect classes** across two rounds.

## The recurring cause

The check reconstructs, by hand, a correspondence between **recogniser records** (its ground truth, per the ADR 0015 carve-out) and **compiler measurement identities** (its coverage). That correspondence is not a simple kind+position match, and every attempt to approximate it has produced defects that do not exist:

| # | reproduced defect |
|---|---|
| 1 | witness-geometry matching called a fully-dimensioned pocket undefined (its size is a leader, which has no witnesses) |
| 2 | anchor fallback to `(0,0,0)` collapsed four chamfers onto one key → four fabricated alarms |
| 3 | `location_slot.location` vs the emitted `location_slot.length` → every slot's location reported missing |
| 4 | kind-only requirements demanded location dims from edge-anchored pockets that correctly have none |
| 5 | **patterns**: the compiler collapses 3+ members into one `pocket_pattern`/`slot_pattern`; the checker lists members and requires exact kind match, so every member of every pattern reads fully uncovered |
| 6 | **off-centre slots**: a lone `SlotFeature.frame` is deliberately placed at the part bbox centre on the transverse axes (`model/detect.py`), while the checker anchors on `Slot.location`, the true centroid — so no off-centre slot can ever match |

5 and 6 are P0 and were each reproduced on independent real parts. 6 is especially instructive: the slot test I wrote used a **centred** subtraction, so it passed by construction and could not have caught it.

## And one false negative that is inherent, not a gap

A pocket's X and Y location dimensions carry the **same** identity, `location_pocket.location`. Coverage is a set membership test, so deleting either one still reads `location='covered'` — the pocket is manufacturably under-located and the check says it is fine. Fixing this needs directional identity or counts; per-member location identity is #883.

## What is actually needed

1. A correspondence model owned by recognition and consistent with the compiler — covering **grouping** (member → pattern), the compiler's deliberate frame conventions (a lone slot's transverse origin), and cardinality.
2. Location identity with enough granularity to answer "located in both required directions" (#883).
3. An authoritative, fail-closed kind inventory (#1008).

Until 1 and 2 exist, a completeness verdict cannot be trusted, and an untrustworthy completeness check is worse than none: #996 exists to remove false confidence, and this would manufacture it.

[...]

PR #1011, opened 2026-08-03
fix(#914): a lost A/F callout is a completeness failure, not just a placement one

pzfreo:
Closes part of #914.

## What was wrong

A machined flat truncating round stock has exactly **one** size parameter — its
across-flats. Every other dimension on such a drawing describes the *stock*. So when the
A/F callout goes, the feature's entire definition goes with it and the part cannot be made
from the drawing.

The engine was not silent: `render_flats` raises `flat_dropped` when no leader lands clear.
But that is a report about the **layout** — "no clear room" — raised from inside the
placement pass. Nothing in it says the drawing has stopped defining the part, and it says
nothing at all when the callout is lost by another route, because it only fires on the one
code path that places it.

## What this adds

`lint_flat_coverage`, beside the other drawing-derived coverage checks.

```text
warning  flat_dropped             flat callout 25 A/F not placed (no clear room)
warning  flat_not_dimensioned     machined flat 25 A/F on the Z stock has no across-flats
                                  callout on the sheet — the flat's only size definition
warning  flat_callout_mismatched  the Z stock's flat measures 25 A/F but the sheet calls
                                  it out as 24 A/F
```

Both of the first two are emitted deliberately: one names the cause, the other the
consequence. That is what the issue asks for — *"a bare placement warning is insufficient."*

**A flat is covered by an A/F leader tipped at that flat's own page position.** Three
conditions, each earned by a review round:

- *A leader*, because a size defines a feature only when something points at it. Text alone
  let the drawing title and any free-form note mentioning the size satisfy a flat.
- *Tipped at the flat*, because association by value is not association. `render_flats`
  leads from `FlatFeature.frame.origin`, which **is** the recogniser's `Flat.at`.
- *An A/F callout*, anchored — `25 A/F`, `25 ±0.2 A/F`, `2× 25 A/F`, `A/F 25`, but not
  prose containing a size.

A leader at the flat stating a size the geometry contradicts is a *wrong* dimension, not a
missing one, and gets its own code naming both numbers. Ground truth stays the geometry:
ADR 0015 is explicit that coverage does not trust the part model, because a stale or
mistaken declaration is exactly what it must catch.

## How it got here

Eight adversarial review rounds; nine findings. Worth reading as a record, because the
interesting part is not the individual bugs:

| round | finding | fix |
|---|---|---|
| r1 | a title block's label satisfied the check | skip non-leaders |
| r1 | `geometry_issues` did not count it | register the code |
| r2 | one label covered two flats; one `25.05` covered 25.0 and 25.1 | consume each callout |
| r3 | greedy pairing reported a *defined* flat as missing | maximum matching |
| r4 | `dwg.note("USE 25 A/F SPANNER")` disabled the check | leaders only |
| r4 | "first number wins" rejected `2× 25 A/F` | anchored pattern |
| r5 | a declared value the geometry contradicts read as "no callout" | `flat_callout_mismatched` |
| r6 | two leaders on one flat silenced a second, undefined flat | **associate by position** |
| r7 | two same-sized flats on parallel stock group as one | inherited — filed as #1013 |

r2, r3 and r6 are one class: matching labels to features by *value* is not association at
all, and no amount of cleverness in the matching rescues an association never made. r6
prompted a rewrite rather than a seventh patch. The check got **smaller** — Kuhn's matching,
the orphan-pairing heuristic and the global label pool are all deleted, because with
positional association there is nothing to mis-assign. That whole class is now structural,
not something tested for.

## Verification

Every test was canaried by breaking what it protects and confirming the named test goes red:

| mutation | caught by |
|---|---|
| unwire the check from `Drawing.lint()` | drop test, removal test |
| accept any annotation, not just leaders | note test, title test |
| unanchor the callout pattern | appended-prose case |
| drop the quantity-prefix alternation | `2×` case |
| ignore the tip | 6 tests |
| widen the tip window to the page | the same 6 |
| restrict a group to its first face | 11 tests |
| report every unmatched group as a gap | both mismatch tests |
| report every unmatched group as a mismatch | 16 tests |
| drop the new code from `_GEOMETRY_AWARE_CODES` | the register assertion |

**Three of my own guards were fake, and only mutation found them:**

1. the `locate()`-style note test used Codex's "USE 25 A/F SPANNER" wording, which the
   anchored pattern rejects *before* the leader rule is reached — so it stayed green with
   the leader rule deleted;
2. `geometry_issues >= 1` was satisfied by an unrelated `feature_not_dimensioned` in the
   same drawing, so it stayed green with the new code unregistered;
3. after the r6 rewrite, the leader rule was protected only by `Note` incidentally having
   no `tip` — accidental, not intent — so it now has a dedicated guard.

One comment I wrote was also wrong in a way worth naming: it claimed greedy matching could
only fail on labels within `tol` of each other. The r3 counterexample's labels are 0.2 mm
apart, outside the 0.15 mm window I said was required. I had written it from the shape of
the algorithm rather than from anything tested.

Also run: the four lint gates, and the full fast tier locally at each stage (**2704 passed**)
since `Drawing.lint()` is shared engine code.

## Architectural fit

- **ADR 0007** — draftwright owns linting; the check lives in `linting/`.
- **ADR 0015** — the lint/coverage carve-out has coverage re-recognise from geometry rather
  than trust the part model, because the defect being detected *is* what the model omitted.
  This follows it, and r5 was resolved by keeping to it rather than departing from it.
- **ADR 0014** — placement is untouched; this reads the finished sheet.
- **DAG** — `linting/coverage.py` already imports `recognition/`; one name added, nothing
  upward.
- **ADR 0011 / #574 round-trip** — `flat` already recognises, emits and declares, so no
  recognition-only debt is added. Both declared paths are tested.
- **#1008** — closes one of the recognised kinds the completeness check did not cover.

## What this does NOT do

**The placement escalation ladder** #914 also asks for (opposite corridor → supporting view
→ larger page) is left open on the issue. No synthetic fixture drops the callout at any
auto-chosen page and scale: I swept A4/A3 × {auto, 1:1, 2:1, 4:1} over Z- and Y-axis flatted
stock, with and without a turned step, with and without a crowding bolt pattern, and at the
case study's 40-metre scale. `_corner_candidates` already tries each member's corner and one
lands every time. The drop needs a caller-*forced* `scale=2.0` on A4, which the engine warns
about up front and which already reports `view_out_of_bounds`. The machinery has no
demonstrated need; if the real `IF_step_flat_across_cylinder.step` still drops at A0 1:200,
that reproduction would justify it, but I cannot run that file.

## Issues raised

- **#1012** — none of the drawing-derived coverage checks knows about ADR 0016 authored
  suppression, so a deliberate omission reads as a completeness gap. This PR inherits the
  behaviour rather than deciding it; six checks is enough silent precedent that the policy
  should be stated once.
- **#1013** — `Flat` records no stock identity, so the renderer draws one callout for two
  same-sized flats on parallel stock and this check cannot see that. The fix is the record
  (ADR 0013). Documented in the docstring and pinned by a test that asserts current
  behaviour and says it is not endorsing it.

### DIFF
--- a/src/draftwright/_core.py
+++ b/src/draftwright/_core.py
@@ -837,7 +837,6 @@ class Analysis:
     pockets: list
     pocket_patterns: list
     pads: list  # geometry-derived rectangular-pad coverage inventory (#885)
-    flats: list  # machined flats (recognise_flats), detected once — the one inventory (#244)
     z_diams: list[float]
     cross_diams: list[float]
     cyls: tuple[list, list]
--- a/src/draftwright/_core.py
+++ b/src/draftwright/_core.py
@@ -837,6 +837,7 @@ class Analysis:
     pockets: list
     pocket_patterns: list
     pads: list  # geometry-derived rectangular-pad coverage inventory (#885)
+    flats: list  # machined flats (recognise_flats), detected once — the one inventory (#244)
     z_diams: list[float]
     cross_diams: list[float]
     cyls: tuple[list, list]
--- a/src/draftwright/analysis.py
+++ b/src/draftwright/analysis.py
@@ -52,6 +52,7 @@ from draftwright.recognition import (
     full_cylinders,
     recognise_bosses,
     recognise_countersinks,
+    recognise_flats,
     recognise_hole_patterns,
     recognise_holes,
     recognise_pocket_patterns,
@@ -546,6 +547,10 @@ def _analyse(
     pockets = recognise_pockets(part)
     pocket_patterns = recognise_pocket_patterns(pockets)
     pads = recognise_rectangular_pads(part)
+    # Detected here rather than inside the lint that reads them, so a repeated
+    # `lint()` does not re-run the O(faces × stock) adjacency scan (#244 / ADR 0015,
+    # Codex #1011 r23). `cyls` is already in hand, so this costs the face loop once.
+    flats = recognise_flats(part, cyls=(z_cyls, cross_cyls))
     # Build the IR once, up front, so page/scale selection sizes from the SAME feature
     # model the renderers use — detected and declared parts share one sizing path and no
     # recogniser record reaches the sheet estimators (ADR 0008; #584 WP1 A). A declared
@@ -728,6 +733,7 @@ def _analyse(
         pockets=pockets,
         pocket_patterns=pocket_patterns,
         pads=pads,
+        flats=flats,
         z_diams=z_diams,
         cross_diams=cross_diams,
         cyls=(z_cyls, cross_cyls),
--- a/src/draftwright/analysis.py
+++ b/src/draftwright/analysis.py
@@ -52,7 +52,6 @@ from draftwright.recognition import (
     full_cylinders,
     recognise_bosses,
     recognise_countersinks,
-    recognise_flats,
     recognise_hole_patterns,
     recognise_holes,
     recognise_pocket_patterns,
@@ -547,10 +546,6 @@ def _analyse(
     pockets = recognise_pockets(part)
     pocket_patterns = recognise_pocket_patterns(pockets)
     pads = recognise_rectangular_pads(part)
-    # Detected here rather than inside the lint that reads them, so a repeated
-    # `lint()` does not re-run the O(faces × stock) adjacency scan (#244 / ADR 0015,
-    # Codex #1011 r23). `cyls` is already in hand, so this costs the face loop once.
-    flats = recognise_flats(part, cyls=(z_cyls, cross_cyls))
     # Build the IR once, up front, so page/scale selection sizes from the SAME feature
     # model the renderers use — detected and declared parts share one sizing path and no
     # recogniser record reaches the sheet estimators (ADR 0008; #584 WP1 A). A declared
@@ -733,7 +728,6 @@ def _analyse(
         pockets=pockets,
         pocket_patterns=pocket_patterns,
         pads=pads,
-        flats=flats,
         z_diams=z_diams,
         cross_diams=cross_diams,
         cyls=(z_cyls, cross_cyls),
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -2702,6 +2702,7 @@ class Drawing:
                 self.part,
                 self,
                 cyls=cyls,
+                flats=a.flats if a is not None else None,
                 assembly=self.assembly,
             )
             model = self._part_model
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -73,6 +73,7 @@ from draftwright.linting import (
     lint_declaration_reconciliation,
     lint_drawing,
     lint_feature_coverage,
+    lint_flat_coverage,
     lint_location_coverage,
     lint_prismatic_coverage,
 )
@@ -2692,6 +2693,14 @@ class Drawing:
                 self,
                 assembly=self.assembly,
                 **prof_kw,
+            )
+            # A flat's A/F is its only size parameter, so a lost callout is a completeness
+            # failure, not just a placement one — reported alongside `flat_dropped` (#914).
+            issues += lint_flat_coverage(
+                self.part,
+                self,
+                cyls=cyls,
+                assembly=self.assembly,
             )
             model = self._part_model
             # Turned bosses/bands remain in the OD + axial-step policy; this check is
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -97,6 +97,7 @@ _GEOMETRY_AWARE_CODES = frozenset(
         "pocket_not_located",
         "unrecognised_defining_geometry",
         "axial_length_missing",
+        "flat_not_dimensioned",
         "missing_principal_dimension",
         "label_vs_measured",
         "dim_inside_part",
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -2702,7 +2702,6 @@ class Drawing:
                 self.part,
                 self,
                 cyls=cyls,
-                flats=a.flats if a is not None else None,
                 assembly=self.assembly,
             )
             model = self._part_model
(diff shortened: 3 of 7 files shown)


pzfreo:
## High-level review: this PR has reached an architectural stop condition

The individual findings have mostly been real, and the issue spillover is useful discovery. But after 24 adversarial rounds, the shape of the work is now stronger evidence than the latest green iteration:

- 1,115 additions for one completeness check;
- 798 lines of new tests and roughly 275 lines of coverage logic;
- several discarded association strategies (shared values, consumption, maximum matching, point proximity, chord proximity, successive stock keys);
- an optional caching change added late and reverted because it violated the declared-model contract;
- this closes only part of #914 while surfacing #1012, #1013, #1014 and #1015.

The recurring problem is not insufficient edge-case testing. `lint_flat_coverage` is trying to reconstruct the semantic question **“does this placed annotation define this particular flat?”** from presentation artifacts: parsed label grammar, leader type, tip position, projected face geometry, cylinder radius, axis line and axial extent. Each round finds another case where those clues are ambiguous because the association was lost upstream. More adversarial review is likely to keep producing valid counterexamples.

I recommend stopping this PR, marking it draft, and treating the current branch as a discovery/prototype rather than a merge candidate. The recovery sequence should be:

1. Fix the flat inventory first: #1013/#1015 need stable stock identity and correct recognition before completeness can judge it.
2. Settle the recognition↔compiler correspondence in #1009 and the placed-measurement identity work around #1002/#1004/#1005.
3. Decide authored suppression once in #1012.
4. Rebuild flat completeness as an explicit chain:

   ```text
   recognised flat requirement
           ↓ stable feature/stock identity
   approved A/F measurement
           ↓ measurement identity
   placed or dropped annotation
   ```

5. Keep label/tip inference only as a conservative fallback for genuinely external/manual annotations. Where association cannot be proven, report it as unverifiable rather than guessing.

That would replace the current reverse path:

```text
scan labels → parse tolerances → inspect leader tips → project geometry
→ infer stock → solve association
```

The fixtures and discovered stock cases are worth salvaging. The downstream matching machinery should not be preserved merely because it took many rounds to obtain. The fact that the PR keeps raising more foundational issues than it closes is not a criticism of the review quality; it is the signal that this feature sits above contracts the codebase has not established yet.

My recommendation is therefore **do not continue another local review/fix round on this head**. Split the prerequisites, keep #1011 draft or supersede it, and return once semantic identity reaches the placed annotation.

pzfreo:
Agreed, and converted to draft. Stopping here — no round 26.

## The part of your framing I had missed

I reached "the association was lost upstream" myself at round 6, then walked it back for what
still looks like a real reason: a check that reads the pipeline's own provenance cannot see the
pipeline get it wrong, and that independence is what let this branch catch #1013 and #1015. So
I chose the reverse path deliberately and then spent nineteen rounds paying for it.

Your point 5 dissolves that trade-off in a way I did not consider. The defect was not choosing
inference over provenance — it was that `lint_flat_coverage` has only **two** answers, covered
and not-covered, so every ambiguous case had to be resolved as one or the other. That is what
made each round's counterexample possible: not a missing clue, but a forced verdict. A third
state — *unverifiable* — keeps the independence without the guessing, because "I cannot prove
this annotation defines this flat" is a true and useful thing to report, and it is what most of
the twenty-three findings actually were.

## What is worth salvaging, concretely

**Already written and tested, and it is your step 1:**

- `Flat.axis_at` — the OD the recogniser matched by edge adjacency, surviving into the record
  instead of consumers re-deriving it from proximity (which picks the wrong axis whenever
  another cylinder sits nearer a face than its own does).
- `Flat.stock` — `(solid, axial start, axial end)`. Stock identity is the axis LINE **and** its
  EXTENT: parallel lobes share a span and differ in line; coaxial regions share a line and
  differ in span; one double-D matches on both. Neither half alone is sufficient, which cost
  three rounds to establish.
- **The #1015 fix** (a8e4114) — the opponent search now requires the same `span`, so two lone
  D-flats on separate coaxial regions stop being paired as each other's opposite face. That was
  a wrong number on the feature's only size parameter, and it is fixed here.

That trio is a small, self-contained PR against `recognition/flats.py` plus its converter, with
no dependency on any of the coverage machinery. I would lift it out as-is.

**The fixtures**, which are the discovery: two parallel lobes, two coaxial regions, a stepped
shaft with different diameters per region, a four-flat section, and two lone D-flats on
separate coaxial stock. Each was built to expose a specific identity confusion and each still
does.

**Not worth preserving:** the label grammar, leader-type filter, tip/chord association,
nearest-wins assignment, tie-breaking and the stock-keyed radius lookup — the whole
`scan → parse → project → infer` chain. It should not survive on the grounds that it was
expensive to obtain.

One note for whoever picks up #1012: the authored-suppression question is not specific to
flats. All six drawing-derived coverage checks share it, and none of them documents a position.

## Method note

Worth recording separately from the design: the recurring error in my own work here was taking
ground truth from the thing under test. Seven guards were green against broken code; two
comments asserted invariants a counterexample falsified within one round; and one test constant
was copied from buggy recogniser output, so the expectation inherited the defect it should have
caught. Mutation testing found all of them, and nothing else would have.

pzfreo:
Superseded by ADR 0017 Amendment 1 (#1040) and the evidence-gated Gate 1 plan in #1018.

## Fixture inventory

Keep as independent geometry/behaviour fixtures:

- one double-D/hex stock region: two recognised faces, one physical A/F requirement;
- two parallel lobes: equal A/F and axial span, distinct axis lines, two requirements;
- two disjoint coaxial regions: equal A/F and axis line, distinct stock spans, two requirements;
- stepped coaxial regions with different diameters: each region owns its radius/size;
- lone D-flats on separate coaxial stock: they are not opposing faces.

The stock-identity and opponent-search discoveries are already represented on `main` by `tests/test_flat_stock_identity.py` and the landed #1013/#1015 fixes. Gate 1 will reuse the first three as correspondence fixtures and keep the latter two as recogniser correctness guards.

Discard from this head:

- A/F label grammar;
- leader/non-leader classification as semantic proof;
- tip/chord/page-position matching;
- nearest-wins/tie-breaking assignment;
- stock-radius reconstruction from presentation;
- any covered/not-covered verdict that guesses through missing provenance.

The replacement starts from the shared `RecognitionResult`, exact recognition-to-IR correspondence, compiled `DimensionId`, registry measurement provenance, compiler omissions, and an explicit unverifiable state. This draft remains useful discovery history, but its implementation should not be revived or cherry-picked.

Issue #1015, opened 2026-08-03
recognise_flats pairs opposed faces across an axis line without checking they are the same stock, giving a wrong A/F


pzfreo:
`recognise_flats` phase 2 pairs a flat with an "opponent" — an antiparallel face across the
same axis — to decide whether `across` is flat-to-flat (a double-D) or flat-to-opposite-OD (a
lone D). The opponent search requires the same axis *line* (`_same_axis_line`) but **not the
same piece of stock**.

So two lone D-flats on two different coaxial regions pair with each other, and `across` comes
out as the sum of two unrelated chord offsets.

## Reproduction

Two coaxial Z cylinders, ⌀40 with a lone flat at `x=12` and ⌀30 with a lone flat at `x=-9`:

```python
flats = recognise_flats(lower + upper)
# both records read 21 A/F  (12 + 9)
# the true lone-flat sizes are 32 A/F and 24 A/F
```

The records already carry distinct `stock` values — `(0, 0, 10.5)` and `(1, 20, 30)` — because
`Flat.stock` was added in #1011. The pairing loop just never compares them.

## Why it matters

`across` is the flat's ONLY size parameter (#914). A wrong one is not a placement or coverage
problem; it is a wrong number on the drawing, and the part is made to it.

`render_flats` and `lint_flat_coverage` both consume the recogniser's value, so they agree
with each other and neither reports anything. This is invisible to the completeness check by
construction — coverage reconciles *whether* a size is stated, not whether the recogniser
derived it correctly.

## The fix

Require the same stock identity in the opponent search, not merely the same infinite axis
line — `_same_axis_line(...) and other["span"] == cand["span"]`. Phase 1 already has the
matched cylinder's extent in hand; the pairing loop is the one place that ignores it.

A fixture needs two coaxial regions of different diameters, each with a single flat.

Related: #1013 (the renderer collapsing same-sized flats on separate stock — the same "which
piece of stock" question, in the grouping rather than the pairing).

Found by the twenty-fourth adversarial review round on #1011.

Issue #1013, opened 2026-08-03
render_flats collapses same-sized flats on parallel stock into one callout


pzfreo:
`recognition.flats.Flat` records `axis` (the letter "x"/"y"/"z"), `across`, and `at` (the
face centre). It does **not** record which piece of stock the flat belongs to, and that
turns out to be the one thing two consumers need.

## The ambiguity

Two faces at `(-12.5, 0, 0)` and `(12.5, 0, 0)`, axis `z`, across 25 — the two faces of one
double-D, **one** A/F definition.

Two faces at `(0, 0, 0)` and `(100, 0, 0)`, axis `z`, across 25 — two D-flats on two
parallel lobes, **two** independent A/F definitions.

From the record alone these are the same shape of data. Nothing in `Flat` says whether two
flats share an axis line.

## What it breaks

**The renderer collapses them.** `render_flats` groups by `(g.facts.axis, round(pd.value, 3))`
(`annotations/from_model.py:1713`) and emits one job per group. A part with two parallel
25 mm-A/F lobes gets **one** callout, leaving the second lobe undefined on the sheet.

**Coverage cannot catch that**, because it groups the same way (`linting/coverage.py`,
`lint_flat_coverage`, added in #1011). It was written to mirror the renderer so a double-D's
two faces count as one requirement — correct for a double-D, wrong for two lobes. The check
therefore reports the two-lobe drawing as complete when it is not.

Reproduction (against `lint_flat_coverage` directly, from the #1011 r7 review):

```python
flats = [Flat("z", 25, (0, 0, 0)), Flat("z", 25, (100, 0, 0))]
# sheet carries one "25 A/F" leader tipped at the first flat
lint_flat_coverage(part, dwg, flats=flats, assembly=False)   # [] — the second is unreported
```

Making coverage stricter on its own would only report a drawing the engine cannot currently
produce correctly, so both ends want the same fix.

## The fix is the record

ADR 0013 is explicit about this case: *"where a record first looks too thin (a face level, a
turned step) the fix is the record"* — `recognise_face_levels -> list[FaceLevel]`,
`recognise_turned_steps -> list[TurnedStep]` each carrying its own `axis`. `Flat` needs the
same treatment: carry the stock's axis **line** (the in-plane position of the axis it is
coaxial about), not just the axis letter.

With that, grouping becomes `(axis_line, across)` in both places, and:

- two faces of one double-D share an axis line → one callout, one requirement;
[...]

PR #1016, opened 2026-08-03
docs: propose recognition correspondence and provenance ADR

pzfreo:
## What changed

Adds proposed ADR 0017, defining a first-class recognition result, an explicit reconciliation stage for competing geometric claims, and deterministic feature/measurement-requirement identity carried through compilation and placement.

Updates the ADR index and reading path.

## Why

PR #1011 and issues #1002/#1004/#1005/#1006/#1009/#1013/#1015 exposed the same missing contract: downstream consumers repeatedly reconstruct which placed annotation corresponds to which physical manufacturing requirement using labels, page geometry and local heuristics.

ADR 0013 standardizes individual recognizers and ADR 0015 defines the compiler waist, but neither owns aggregate recognition inventory, conflict reconciliation, or end-to-end measurement provenance. This draft proposes that missing architectural seam before more feature-specific completeness matchers are added.

## Scope

This is a design proposal only. It deliberately does not choose concrete identifier schemas or begin a large-bang migration. It preserves:

- ADR 0013’s geometry-only recognizer contract;
- ADR 0015’s independent physical-completeness principle;
- ADR 0011’s no-detection declared build path;
- ADR 0010’s single annotation-provenance registration seam;
- ADR 0014 placement ownership.

The ADR includes consequences, rejected alternatives, staged migration constraints, and the guards required before acceptance.

## Validation

- `git diff --check origin/main...HEAD`
- `uv run codespell docs/adr/0017-recognition-inventory-correspondence-and-measurement-provenance.md docs/adr/README.md`

### DIFF
--- a/docs/adr/0017-recognition-inventory-correspondence-and-measurement-provenance.md
+++ b/docs/adr/0017-recognition-inventory-correspondence-and-measurement-provenance.md
@@ -152,6 +152,13 @@ measurables when forming requirements; a correlated requirement may render as se
 Therefore feature identity, measurable identity, requirement identity and annotation identity
 remain distinct types rather than one overloaded key.
 
+They are also distinct **Python types** — provisionally `FeatureId`, `MeasurableId`,
+`RequirementId` and `AnnotationId`, implemented as frozen value types rather than four aliases
+for a bare tuple or string. Passing a measurable identity where a requirement identity is
+expected must be a static/runtime type error, not a valid lookup that happens to return the
+wrong result. Their internal fields may reuse smaller typed geometry keys; consumers do not
+assemble or unpack those fields independently.
+
 This is not a flat-only correction. `Flat`, `Chamfer`, `Fillet`, `Groove` and `CounterSink`
 currently carry an axis letter plus a point but cannot by themselves answer “which material
 region am I on?”. `TurnedStep`, `Plate` and `Slot` already carry useful extents. Migration must
@@ -170,6 +177,14 @@ For example, several step measurables may become one correlated length-chain req
 a groove-owned diameter may make a coincident boss measurable redundant. Recognition does not
 make those drafting decisions.
 
+That policy has one explicit, low-level home, provisionally `draftwright.requirements`: a pure
+measurable→requirement mapping below both `model/` and `linting/` in the dependency DAG. It takes
+geometry-only measurable data plus explicit drafting-policy inputs and returns typed
+requirements; it imports neither the compiler/planner nor placed annotations. `model/` consumes
+it when compiling intent, and `linting/` consumes the same mapping when forming its independent
+physical-completeness inventory. The two callers may select different output projections, but
+they may not reimplement applicability, correlation or redundancy rules.
+
 The compiler records the fate of each applicable requirement:
 
 - approved;
@@ -186,18 +201,19 @@ recognition choose strips and coordinates.
 
 ADR 0015's lint carve-out remains: completeness cannot take the compiled dimension set as its
 only inventory, because an omission before planning would then be invisible. Its independent
-ground truth is the recognised physical requirements.
+ground truth is requirements formed from recognised measurables by the shared requirements
+policy — independently of what the compiler claims to have approved.
 
 Independence from the plan does **not** require independence from the recognition result, nor
 does it justify recovering association from rendered text when provenance exists. Lint compares:
 
 ```text
-recognised applicable requirements ↔ compiler/placement outcomes
+requirements(recognised measurables, drafting policy) ↔ compiler/placement outcomes
 ```
 
 It reports at least these distinct conditions:
 
-- no intent was approved for a required measurement;
+- no intent was approved for a requirement;
 - intent was deliberately suppressed;
 - intent was approved but placement dropped it;
 - an annotation states a conflicting value or identity;
@@ -243,7 +259,9 @@ lint wording belong to draftwright's critique layer.
   is clarified to mean one explicit recognition result per recognition run, not an informal
   collection split across `_analyse()` and `build_part_model()`.
 - **ADR 0015's lint carve-out remains.** Lint is independent of the compiler's claimed
-  completeness, but may consume the same recognition result as independent physical evidence.
+  completeness. It may consume the same recognition result as independent physical evidence
+  and the shared requirements policy, but does not import `model/` or trust compiler output as
+  its inventory.
 - **ADR 0010 remains the annotation provenance seam.** This ADR specifies the semantic identity
   that provenance must carry; it does not create a second registration seam.
 - **ADR 0016 remains the owner of authored dimensioning intent.** Suppression is a recorded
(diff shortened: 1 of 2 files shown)


pzfreo:
Reviewed against the recognition subsystem as it stands and against what #1011 actually did.
**Support this, with two changes I would want before Accepted.** Both are specific; neither
touches the shape of the decision.

## What it gets right that I got wrong

Two clauses would have prevented most of #1011's twenty-three findings, and I did not have
either of them:

- **§5, "unverifiable is preferable to a plausible wrong match."** `lint_flat_coverage` has
  only two answers, so every ambiguous case had to be forced into one. That is what made each
  round's counterexample possible — not a missing clue, a forced verdict. The third state is
  the fix.
- **§7, "ambiguity is output, not absence."** An empty recogniser list currently conflates
  *absent*, *ambiguous*, *unsupported* and *lost the ownership contest*. I did not identify
  this and it is the more far-reaching of the two.

The rejected alternatives are real ones. "Make lint trust `PartModel`" is rejected for exactly
the reason I defended the harder design at r9, so §5 keeps the independence without keeping the
guessing — which is the tension I could not resolve myself.

## 1. `solid_idx` violates §3, and it is what migration step 2 would build on

§3: identity "must not depend on Python object identity, **traversal order**, annotation names,
formatted labels, or page coordinates."

Migration step 2: "introduce deterministic stock/region identity using the flat cases in
#1013/#1015."

Those flat cases are `Flat.stock = (solid_idx, s_lo, s_hi)` — the field I added on #1011. And
`solid_idx` is an enumeration index over `part.solids()` (`recognition/_features.py:91`,
`for solid_idx, face in faces_by_solid`). It is traversal-order derived, so the first concrete
identity the migration lays down would violate the principle the same ADR sets.

Practical exposure is multi-solid parts only — for a single solid it is always `0` — but that
is precisely the assembly case, and an index is exactly the kind of thing that looks stable
until a solid is added upstream and every identity after it shifts.

Suggest §3 name the positive form: a stock/region identity is derived from **geometry** — axis
direction, in-plane axis position, axial extent, diameter — and never from an index into a
traversal. That also makes the identity survive the import/redetection the guards ask about,
which an index does not.

## 2. "Requirement" is doing two jobs

- §1 gives the recognition result ownership of "physical measurement requirements recovered
  from those records."
- Rejected alternatives: "Recognition supplies geometry, evidence and physical measurement
  facts. **Drafting convention decides which requirements are applicable**, redundant or
  presentational."
- §4: "The compiler records the fate of each **applicable** requirement."

Read together these are consistent, but only if the reader notices that a *measurable physical
fact* and a *drafting obligation* are different things wearing one word. That matters for
placement: the facts are geometry and can live in `recognition/` (and eventually in the shared
geometry-only package), while applicability is policy and cannot. A future
`b123d-recognisers` split will land on this seam directly.

Suggest two nouns — e.g. **measurable** (geometry: this flat has an across-flats of 25) and
**requirement** (drafting: a complete drawing must state it) — with §1 owning the first and the
compiler owning the second. Everything else in the ADR then reads unchanged.

## Smaller notes

- **§6's lazy cached recognition needs a home.** The guards want "linting the same built
  drawing twice does not rerun recognition", which means cached state on `Drawing`. ADR 0005 /
  #639 reduced that to one typed `BuildState` with a single-writer guard; worth saying the
  cache goes there rather than becoming a new private attribute.
- **`Analysis`'s fate is unstated.** It carries sizing facts (bbox, zones, `is_rotational`)
  as well as inventories. Does `RecognitionResult` replace part of it or sit beside it? #639
  fought over exactly this ownership, so it is worth one sentence.
- **Baseline for the first guard.** "One recognition orchestration call per automatic build"
  is testable, but the starting point should be recorded so the guard has something to ratchet
  from. Measured today: of ~17 recognisers, **7** are carried on `Analysis`, **8** are
  injectable into `build_part_model`, and **8** are neither — countersinks, grooves, plates,
  chamfers, fillets, flats, step_shoulders, face_levels — so those re-run inside detect on
  every build regardless of what `_analyse` already did.
- **Record thinness is wider than flats.** §3 covers it in principle; worth naming the scope.
  `Flat`, `Chamfer`, `Fillet`, `Groove` and `CounterSink` carry an axis *letter* plus a
  *point*, so none of them can answer "which material am I on". `TurnedStep`, `Plate` and
  `Slot` already carry an extent and are fine. Five records need the §3 treatment, not one.

## On sequencing

Migration steps 1–2 are largely written and tested on #1011's branch (`Flat.axis_at`,
`Flat.stock`, the #1015 pairing fix, and the fixtures: parallel lobes, coaxial regions, stepped
shaft, four-flat section, lone-D pair). They can land as phase-0 evidence once the identity
drops `solid_idx` per point 1 — which is a reason to settle that wording before the first
implementation PR rather than after.

pzfreo:
Re-reviewed at 0a6666c. All four points from my earlier comment are addressed, and addressed
more thoroughly than I asked — the measurable/requirement split propagates through the identity
types, the pipeline diagram, the boundaries section and the rejected alternatives rather than
sitting in one paragraph. The `solid_idx` correction now names the failure mode ("adding or
reordering one solid can renumber every later record") and migration step 2 says to replace it
rather than promote it, which closes the trap I was about to walk into on the lift.

Two corrections to me that I accept: "neither" in the baseline does not imply "run twice", and
#1011's reusable evidence is its fixtures and counterexamples, not its matching machinery.

One new finding on this version, then two smaller notes.

## §5 puts lint's ground truth on the wrong side of a guarded boundary

§4 establishes that a **measurable** is geometry and a **requirement** is drafting policy —
"drafting convention decides applicability, correlation and redundancy." That split is right,
and it is what makes the shared-package boundary work.

§5 then says lint's independent ground truth is "the recognised physical **requirements**", and
that it compares:

```text
recognised applicable requirements ↔ compiler/placement outcomes
```

That phrase carries the tension in miniature: *recognised* is rank-0 geometry, *applicable* is
drafting policy. Today the only place that turns measurables into requirements is the planner,
in `draftwright.model` — and `linting/` may not import it. That is not merely a DAG rank
consequence; `tests/test_import_boundaries.py::test_linting_does_not_import_model` exists
specifically because the rank rule alone *would* permit it, and the test's own docstring cites
ADR 0015's carve-out as the reason.

So as written, §5 asks lint to consume something it is structurally forbidden from reaching.
Three ways out, and the ADR should pick one rather than leave it to surface as a failing
boundary test in phase 6:

1. **Requirements formation gets its own home below both** — a policy module that `linting/`
   and `model/` can each depend on. Cleanest, and it gives the "drafting convention" layer an
   actual address instead of being implicit in the planner.
2. **Lint reads measurables and applies applicability itself.** Rejected on the ADR's own
   opening argument — that is a second competing answer to one semantic question, which is the
   thing this ADR exists to stop.
3. **Relax the carve-out.** §5 explicitly keeps it, so this contradicts the decision.

I would say (1) explicitly in §5, because the difference between "lint reads requirements" and
"lint reads measurables" changes what the completeness check *is*, not just where a module sits.

## The four identity types need to be four types

§3 now distinguishes feature, measurable, requirement and annotation identity. That is right —
the 1:N:M:P chain is real — but four keys flowing through the same code is exactly the shape
that produced the errors this ADR is responding to. On #1011 I spent four review rounds getting
`axis_at` + `stock` combined correctly, and the fix was to stop making consumers combine
anything.

Suggest the ADR require each to be a **distinct Python type**, not a bare tuple or string, so
passing a measurable identity where a requirement identity is expected is a type error rather
than a silently wrong lookup. That is cheap, and it converts the most likely implementation
mistake into one a checker catches. Worth a line in the guards.

## Smaller

- **A fifth term appears.** §5's outcome list says "no intent was approved for a *required
  measurement*". After the careful measurable/requirement split, that reads as a third thing.
  Suggest "requirement".
- **The guards did not move with the split.** "Redetecting an unchanged part yields the same
  feature and requirement identities" predates the four-type model — it should cover all four,
  and there is no guard that the measurable→requirement mapping is deterministic, which is now
  a policy step that can vary.

Issue #1019, opened 2026-08-03
ADR 0017 phase 1: one cached RecognitionResult without changing build semantics


pzfreo:
Child of #1018. Implements ADR 0017 migration phases 0/1 without changing drawing semantics.

## Scope

- Introduce the smallest useful frozen `RecognitionResult` aggregate around recognition inventories already produced during analysis.
- Give the aggregate one explicit orchestration/fill site for the initial migrated families.
- Store the result in typed `BuildState`, preserving its single-writer builder contract.
- Let automatic model detection and geometry-based critique consume the same result or a documented projection.
- Ensure repeated lint on a built drawing reuses cached recognition rather than rescanning migrated families.
- Preserve the ADR 0011 declared-model contract: declared build/render performs no additional recognition. Explicit physical critique may lazily recognise once only if that policy is exposed and guarded.
- Establish a fail-closed inventory manifest: every public `recognise_*` family is migrated or explicitly excluded with a reason.

## Non-goals

- No stable feature/measurable/requirement identity yet (the next phase).
- No reconciliation rewrite.
- No measurable-to-requirement policy.
- No drawing-output changes.
- No universal rewrite of individual recognisers.

## Required guards

- [x] One orchestration call per automatic build. (#1021)
- [x] Shared migrated substrates/recognisers are not rescanned by model construction or repeated lint. (#1021, #1025 — repeated lint now runs zero recognisers)
- [x] Declared build/render invokes no automatic feature recogniser. (#1022)
- [x] Repeated lint returns equivalent results and does not rerun recognition. (#1025; the results-equivalence half was added in review of #1031, which found the guard had asserted only the call count)
- [x] Every public recogniser family appears in the manifest as migrated or explicitly deferred. (#1021; `DEFERRED` is now empty — #1026, #1028)
- [x] Existing representative drawing output remains semantically unchanged. (full suites green throughout; the one deliberate change is #1013's flat grouping)
- [x] A counter/counterexample test fails when either the cache or the declared-path gate is removed. (mutation-validated in every PR)

## Architectural constraints

- `RecognitionResult` owns only recognition state, not page/layout state.
- Individual recognisers retain the ADR 0013 geometry-only contract.
- Linting does not import `model/` and does not trust the compiled plan as its completeness inventory.
- The cache belongs to typed `BuildState`, not a new private attribute on `Drawing`.
- Prefer incremental family migration over a large-bang orchestration rewrite.

[...]

Issue #1018, opened 2026-08-03
Epic: recognition inventory and evidence-gated completeness correspondence


pzfreo:
## Decision update — 2026-08-05

Phase 1 is complete. The original phases 2–6 are **paused hypotheses, not an approved implementation sequence**.

This epic now follows an evidence-gated path:

1. preserve and accept the recognition-ownership contract that has landed;
2. prove one end-to-end completeness slice for flats without presentation inference;
3. prove a second, structurally different slice for slots/patterns;
4. only then decide which generic identity, requirements, outcome, or reconciliation abstractions have earned their cost.

The user-facing outcome remains owned by #996: a drawing that omits a defining requirement must not look complete. This epic owns only the cross-stage contracts that the vertical slices demonstrate are necessary.

## Phase 1 — complete

Every public `recognise_*` family is owned by one orchestration.

| measurement | result |
|---|---|
| `MIGRATED` families | 17 |
| `DEFERRED` | **empty** |
| automatic prismatic build | 17 families, once each |
| automatic turned build | 14 families; three are classification-gated by design |
| declared build/render | **0** recognition |
| first physical critique/export of a declared drawing | may obtain one aggregate |
| subsequent lint of the same drawing | **0 additional** recognition |

### Landed work

| PR | contribution |
|---|---|
| #1020 | `RecognitionResult` shell and typed build-state attachment |
| #1021 | fail-closed `MIGRATED` / `DEFERRED` manifest and one-orchestration guards |
| #1027 | count recogniser calls by code object rather than a patchable proxy |
| #1029 | declared build/render recognises nothing |
| #1030 | migrate the `BUILD_MODEL_ONLY` families |
| #1031 | split step shoulders into shared evidence plus pure per-consumer projection |
| #1033 | put classification-dependent gates inside the orchestration |
| #1035 | carry flat stock identity: `axis_line` plus `stock_span` |
| #1038 | correct the part-classification ownership claim |

### What phase 1 proves

- One recognition orchestration and result can own every current public family.
- Applicability gating can live inside that orchestration without running every family for every part.
- Declared build/render can remain recognition-free while physical critique obtains one cached aggregate.
- Repeated lint does not rescan recognition.
- Record-shape fixes should be attempted before adding reconciliation machinery.

It does **not** yet prove recognition-record → IR-feature correspondence, a shared requirements policy, a requirement outcome ledger, four global identity types, or a general reconciliation stage.

## Gate 0 — align the decision and tracker

[...]

PR #1020, opened 2026-08-03
Add the ADR 0017 recognition result shell

pzfreo:
## What changed

- adds the first frozen `RecognitionResult` aggregate from ADR 0017;
- moves the inventories already shared by analysis/model detection behind one orchestration function;
- preserves the aggregate on `Analysis` and typed `BuildState`;
- makes the cylinder substrate the exact shared object consumed downstream;
- updates the existing detect-once guard to observe the new orchestration seam;
- adds counterexample guards for injected recogniser dependencies, immutable inventory containers, and build-state reuse.

## Why

Recognition state was an informal group of locals on `Analysis`. ADR 0017 requires one explicit result that can grow into the shared feature/measurable/diagnostic inventory. This is the minimal shell: it moves only families already detected in analysis and deliberately leaves reconciliation, semantic identities, requirements, and the remaining recogniser families for later slices of #1019/#1018.

The orchestration function is named `build_recognition_result`, not `recognise_*`, because ADR 0013 reserves that shape for individual feature recognisers returning deterministic lists of records.

## Compatibility

No drawing policy or intended output changes. Existing `Analysis` projections remain while consumers migrate. Declared-path and full inventory/cache completion remain tracked in #1019; this PR does not claim to close it.

## Validation

- `uv run ruff check src/draftwright tests/test_recognition_result.py tests/test_part_model.py`
- `uv run mypy src/draftwright`
- `uv run pytest -q tests/test_recognition_result.py tests/test_part_model.py tests/test_recogniser_contract.py tests/test_score.py tests/test_detect_once.py` (37 passed)

Part of #1019 and #1018.

### DIFF
--- a/src/draftwright/_core.py
+++ b/src/draftwright/_core.py
@@ -26,7 +26,7 @@ from typing import TYPE_CHECKING, Literal
 
 if TYPE_CHECKING:
     from draftwright.compose import StripDepths
-    from draftwright.recognition import TurnedProfile
+    from draftwright.recognition import RecognitionResult, TurnedProfile
 
 from build123d import Align, BoundBox, Compound, Edge, Location, Mode, Shape, Text, Vector
 from build123d_drafting.helpers import (
@@ -822,6 +822,7 @@ class Analysis:
     """
 
     part: Shape
+    recognition: RecognitionResult
     bb: BoundBox
     x_size: float
     y_size: float
@@ -839,7 +840,7 @@ class Analysis:
     pads: list  # geometry-derived rectangular-pad coverage inventory (#885)
     z_diams: list[float]
     cross_diams: list[float]
-    cyls: tuple[list, list]
+    cyls: tuple[tuple, tuple]
     prof: TurnedProfile | None  # turned step profile (recognise_turned_steps), detected once
     od_diam: float | None
     is_rotational: bool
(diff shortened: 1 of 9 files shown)


pzfreo:
Adversarial review loop completed for this slice. Two substantive findings were fixed before publication: (1) the aggregate builder no longer uses the ADR 0013-reserved `recognise_*` naming shape, and (2) the cylinder substrate is now the exact object shared by Analysis/model construction/BuildState rather than an equal-but-parallel container. Dependency injection is mutation-guarded, existing detect-once coverage was moved to the new seam, and targeted validation is green (37 tests, Ruff, mypy, codespell). No substantive findings remain. Non-blocking scope note: this is intentionally the RecognitionResult shell; remaining families, declared-path/lazy-critique behavior, and the fail-closed manifest remain in #1019.

pzfreo:
CI failure addressed. The full matrix had 2 architecture-ratchet failures and 2,676 functional passes: the new BuildState writer was missing from the exact allowlist, and the new test read Drawing._build directly. Fixed without weakening either guard: the single-writer inventory now names `_build.recognition`, and `Drawing.recognition()` provides the experimental read-only surface. Post-fix adversarial review found no substantive issue; exact guards + focused suite pass (41 tests), with Ruff and mypy green.

PR #1021, opened 2026-08-03
feat(#1019): a fail-closed ADR 0017 recognition manifest

pzfreo:
Closes the manifest half of #1019 (ADR 0017 phase 1).

**Reduced after architectural review** ([comment](https://github.com/pzfreo/draftwright/pull/1021#issuecomment-5176027446)).
The previous head (`cd10e6f`, kept at `backup/1021-round10`) also memoised lint's
step-shoulder recogniser on `CoverageState` and amended ADR 0017 §6 to permit it. Review
was right that this made lint a second recognition owner, when ADR 0017's waist is one
authoritative physical result with independent *semantic* consumers. Both are gone. The
code-object call-counting helper is split out to #1027, which this branch is stacked on.

## What this does

**A fail-closed manifest.** `MIGRATED` / `DEFERRED` in `recognition/result.py` classify
every public `recognise_*` family, and the guard reads `draftwright.recognition.__all__`
rather than a hand-kept list — so adding a recogniser without deciding whether the
aggregate owns it fails CI, at the moment the author knows why.

Both halves are mechanical, not prose:

- a `MIGRATED` family must actually be called by the orchestration (checked by *running* it
  with the calls counted, not by reading the source);
- a `DEFERRED` family must be neither importable from it nor callable by it;
- the `CLASSIFICATION_GATED` families are additionally checked against a running build,
  because a detailed-but-wrong reason had already shipped once in this epic and was caught
  by human review rather than by a test.

**Reason codes, not paragraphs.** A deferral is a `Deferral` code plus the issue that
removes the constraint. The prose this replaces went stale inside one review cycle: it
carried a "governing principle" the `MIGRATED` list did not satisfy, and cold timings for a
migration that review reverted. The why-trail belongs in #1022 / #1025 / #1026, which is
where it now is.

**The second `build_part_model` call site.** `annotations.orchestrator.build_model` takes
`pockets`/`pocket_patterns`/`pads` off `a.recognition` instead of re-detecting them. It is
not reachable from `build_drawing` today — which is why it needed its own guard, since
nothing else would ever have noticed.

**A `BuildState` ratchet fix.** `test_build_state_has_a_single_construction_and_fill_site`
takes its watched field names from `__dataclass_fields__`. The hand-kept tuple had fallen
three fields behind, leaving `recognition`, `omissions` and `detail_view` writable by
setattr from any module with the single-writer guard none the wiser.

## What lint does now

It still rescans `recognise_step_shoulders` on every pass — as it did before this branch.
That rescan is **budgeted rather than hidden** (`_LINT_CALL_BUDGET`), so a *new* lint-side
scan fails the guard while the known one is recorded with its blocker.

Budgeting rather than memoising is the point of the reduction. The memo removed a real
rescan, but bought it with a second recognition owner. #1025 removes the same rescan by
splitting the recogniser into raw riser evidence on the aggregate plus a pure projection
per consumer — `levels=` is a *selector* over a scan that happens regardless, so there is
one physical answer for both consumers to project. That also closes the `step_zs=[]`
false-negative door, which the memo left open and merely documented.

## No path recognises more than it did on `main`

Detected prismatic 17 families / 18 calls (`recognise_step_shoulders` twice — the
deferral's two callers, pinned by `_BUILD_CALL_BUDGET`); declared `Box(80,60,20)` 11 / 11,
ratcheted by `test_a_declared_build_does_not_grow_new_recognition`. Identical before and
after.

## Sequence

| | |
|---|---|
| **#1027** | code-object call counting (this branch's base) |
| **this PR** | the fail-closed manifest |
| **#1022** | declared-path zero-recognition gate + the one lazy `BuildState` aggregate |
| **#1025** | raw riser evidence + pure model/lint projections; no caller-narrowable lint inventory |
| **#1026** | migrate what #1022 unblocks; shrink the manifest |

Base is `test/code-object-call-counting` (#1027) — the manifest guard counts families with
that helper. Retarget to `main` once #1027 merges.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -388,7 +388,7 @@ Current ADRs:
   deferral is a `Deferral` reason **code** plus the issue that removes it, not a paragraph:
   prose in a constant CI reads goes stale silently, and the first cut's did. The why-trail
   lives in the blocking issue.
-  Three things are still open, in this order. **#1022** — the ADR 0011 declared-path gate:
+  Four things are still open, in this order. **#1022** — the ADR 0011 declared-path gate:
   a declared build still recognises, and gating it is *not* just skipping the call, because
   page/scale selection reads `step_zs` and the turned profile off that recognition even
   when the model was declared, so sizing has to source them from the declaration first; the
@@ -398,8 +398,12 @@ Current ADRs:
   input differs between model construction (filtered by plate and pocket ownership) and
   critique (unfiltered — ADR 0015 forbids lint taking its inventory from the model); until
   then lint rescans on every pass, budgeted rather than memoised, since a lint-owned memo
-  would be a second recognition owner. **#1026** — migrate the families #1022 unblocks and
-  shrink the manifest. Beyond the epic's phase 1: stable feature identity, and the
+  would be a second recognition owner. **#1026** — migrate the three `BUILD_MODEL_ONLY`
+  families #1022 unblocks, and shrink the manifest. **#1028** — the classification gate is
+  a *different* blocker on the automatic path, untouched by #1022: `chamfers`/`fillets`/
+  `plates` migrate only once the orchestration itself carries the part classification, so
+  hoisting them stops meaning "scan every turned build for a discarded result". Beyond the
+  epic's phase 1: stable feature identity, and the
   measurable→requirement policy in the specified `draftwright.requirements` module — a
   pure mapping ranked below both `model/` and `linting/`.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -200,7 +200,12 @@ entry. Keep `_LAYERS` and this section in step.
   `pads.py` (bounded rectangular raised islands, #885)/
   `countersinks.py`, and `_record.py` (the shared frozen-`Record` mixin,
   `.to_dict()`). Bottom of the DAG: depends only on build123d/OCP. Import
-  via the package surface.
+  via the package surface. **`result.py`** sits one step above the individual
+  recognisers (still inside the package): the ADR 0017 `RecognitionResult`
+  aggregate, its one `build_recognition_result` orchestration — which computes
+  each shared substrate once and *injects* it downstream rather than letting a
+  recogniser rediscover it — and the `MIGRATED`/`DEFERRED` manifest that keeps
+  the aggregate's coverage of the package surface honest.
 - **`fonts.py`** — vendored, path-pinned IBM Plex fonts for deterministic
   cross-platform layout (ADR 0006).
 - **`export.py`** — SVG/DXF/PDF/PNG export + post-processing (page-size fix,
@@ -368,6 +373,35 @@ Current ADRs:
   **Not** shipped: the emitter dimension-mirror (phase 4). `emit_sheet_script` refuses a
   model with an authored set rather than silently writing `auto_dimensions()`, because
   naming a feature in a generated script would have to address it by position — #922.
+- **0017** — **Proposed; epic #1018 in progress** (the ADR lists the guards required
+  before it can be accepted): **the recognition inventory as a first-class result**, with
+  semantic feature correspondence and requirement provenance. Recognition stops being a
+  scatter of ad-hoc calls: one orchestration per build produces a frozen
+  `RecognitionResult` (`recognition/result.py`), carried in `BuildState` and reused by
+  model construction *and* by critique — lint reads its inventories off `Analysis`, which
+  ADR 0017 §5 explicitly permits (independence from the *plan* is not independence from the
+  *recognition*).
+  Phase 1 (#1019) landed the **fail-closed manifest** — `MIGRATED` / `DEFERRED` in
+  `recognition/result.py` classify every public `recognise_*` family, and
+  `tests/test_recognition_manifest.py` fails when a new one appears without that decision,
+  so a recogniser cannot be added and then quietly re-scanned from three call sites. A
+  deferral is a `Deferral` reason **code** plus the issue that removes it, not a paragraph:
+  prose in a constant CI reads goes stale silently, and the first cut's did. The why-trail
+  lives in the blocking issue.
+  Three things are still open, in this order. **#1022** — the ADR 0011 declared-path gate:
+  a declared build still recognises, and gating it is *not* just skipping the call, because
+  page/scale selection reads `step_zs` and the turned profile off that recognition even
+  when the model was declared, so sizing has to source them from the declaration first; the
+  same issue owns the one lazily-built `BuildState` aggregate a declared drawing needs when
+  physical critique *is* asked for. **#1025** — `recognise_step_shoulders` splits into raw
+  riser evidence in the aggregate plus a pure projection per consumer, because its `levels=`
+  input differs between model construction (filtered by plate and pocket ownership) and
+  critique (unfiltered — ADR 0015 forbids lint taking its inventory from the model); until
+  then lint rescans on every pass, budgeted rather than memoised, since a lint-owned memo
+  would be a second recognition owner. **#1026** — migrate the families #1022 unblocks and
+  shrink the manifest. Beyond the epic's phase 1: stable feature identity, and the
+  measurable→requirement policy in the specified `draftwright.requirements` module — a
+  pure mapping ranked below both `model/` and `linting/`.
 
 ## Dependencies
 
--- a/src/draftwright/annotations/orchestrator.py
+++ b/src/draftwright/annotations/orchestrator.py
@@ -202,6 +202,14 @@ def build_model(a: Analysis):
         patterns=a.patterns,
         bosses=a.bosses,
         slots=a.slots,
+        # The engine's SECOND build_part_model call site; unthreaded, these three were
+        # detected again here after _analyse had already recognised them (#1019). Read off
+        # the run's one RecognitionResult (ADR 0017); `a.pockets`/`a.pads`/
+        # `a.pocket_patterns` are list copies of the same records and would do as well —
+        # #1024 collapses that duplication.
+        pockets=a.recognition.pockets,
+        pocket_patterns=a.recognition.pocket_patterns,
+        pads=a.recognition.pads,
         prof=a.prof,
         step_zs=a.step_zs,
         rotational=(a.od_diam, _bores, a.od_axis) if a.is_rotational else None,
--- a/src/draftwright/recognition/result.py
+++ b/src/draftwright/recognition/result.py
@@ -50,7 +50,10 @@ class Deferral(Enum):
     list below did not satisfy, and cold timings for a migration that review reverted.
 
     ``CLASSIFICATION_GATED`` — ``build_part_model`` runs it only for one part class, so
-    hoisting it unconditionally scans the other class for a result that is discarded.
+    hoisting it unconditionally scans the other class for a result that is discarded.  Note
+    this constraint is on the AUTOMATIC path: #1022 removes recognition from the declared
+    path and leaves it entirely untouched.  It ends when the orchestration itself carries
+    the classification (#1028).
 
     ``BUILD_MODEL_ONLY`` — its sole engine consumer is ``build_part_model``, which the ADR
     0011 declared path skips.  The aggregate runs unconditionally, so migrating it removes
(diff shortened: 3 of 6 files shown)


pzfreo:
Architectural review against epic #1018 / ADR 0017: I don't think this should merge in its current shape, despite green CI.

The useful parts are real: a fail-closed family manifest is valuable, and the code-object call-count guard is much stronger than a module/binding patch list. But the PR is now validating and documenting an intermediate distributed architecture rather than implementing the epic's intended waist.

**Blocking: the declared-model boundary is in the wrong order.** #1019 requires declared build/render to perform no recognition. This PR explicitly leaves that path at 11 recogniser families and defers the gate to #1022. The manifest's migration choices and cost arguments are consequently shaped around a violation this phase is meant to remove.

**Blocking: `CoverageState.step_shoulders` becomes a second recognition owner.** ADR 0017's intended split is one authoritative physical recognition result with independent semantic consumers—not independent B-rep scans. The cleaner structure is:

```text
solid
  -> RecognitionResult containing raw face/level evidence
       |- pure model projection
       `- pure lint projection
```

Model construction and lint may select different shoulder sets without either rescanning the solid. The fact that `recognise_step_shoulders(..., levels=...)` answers caller-specific questions suggests that raw level/face evidence and the pure projection need separating; it does not imply that lint needs its own recognition cache.

**Blocking: the ADR amendment ratifies temporary debt.** There is one formal 28-line amendment, but it has needed repeated rationale corrections because it exists mainly to explain why a lint-owned cache does not contradict the original `BuildState` ownership rule. I think it does contradict the intended one-result ownership model. More prose makes that divergence explicit, not correct.

**Major: critique is still caller-narrowable.** The new comment correctly admits `step_zs=[]` can silence the check. Removing a `step_shoulders=` argument therefore does not establish independent physical evidence while another caller-controlled input can reduce it to nothing.

**Major: manifest reasons are too coupled to migration accidents.** The fail-closed classification should stay, but durable entries should be concise statuses/reason codes plus named blockers. Current call-site history, cold timings, and declared-path costs will become stale as soon as #1022 lands.

My recommended sequence:

1. Land the declared build/render zero-recognition gate first.
2. Put lazy physical critique behind the single typed `BuildState` recognition cache.
3. Add raw face/level evidence to `RecognitionResult`.
4. Derive model and lint shoulder inventories as pure projections.
5. Remove the `CoverageState` recognition memo and this ADR amendment.
6. Retain a smaller fail-closed manifest.
7. Keep the code-object call-count helper, ideally as an isolated test-quality improvement.

The vision remains recoverable, but the current PR has drifted from **one recognition run, one authoritative geometry result, multiple independent semantic consumers** toward **several recognition paths memoised by different owners**. The unusually high review count is a symptom of proving the latter ever more thoroughly rather than simplifying it back to the former.

pzfreo:
Accepted, and largely agreed. Taking the points in the order they matter.

**Blocking 2 (the second recognition owner) is the load-bearing one, and it is more
tractable than this PR assumed.** `recognise_step_shoulders(part, *, levels)` does a full
`part.faces()` scan and uses `levels` only to *select* which risers qualify. The scan is the
cost; the level set is a projection parameter. So the raw-evidence/pure-projection split you
sketch is a bounded refactor, not a redesign — and it dissolves this PR's stated blocker
directly. The manifest argues a keyed memo "has no home below both `model/` and `linting/`";
under the split the shared thing stops being a keyed cache and becomes evidence, which
already has one. The ownership filtering stays in `model/detect.py` as a projection choice,
so no drafting policy moves below the IR either.

Blocking 3 follows from 2 rather than standing alone: the amendment exists to license the
memo, so it should be withdrawn with the memo rather than re-amended. Agreed that more prose
makes the divergence explicit, not correct.

**Blocking 1 (ordering) — agreed, and it inverts most of this PR's reasoning.** Every cost
argument in `DEFERRED` has the form "migrating adds a scan to the declared path". Gate the
declared path first and that cost is zero, so the six deferred-for-cost families migrate
freely and the prose you (correctly) call stale never needs writing at all.

**Major 1 — conceded without qualification.** `step_zs=[]` is a live false-negative door in
a completeness check, and documenting it as tolerable was the wrong call. It is folded into
the split below, where lint's projection sources its level set from the evidence and the
door closes as a consequence rather than as a patch.

**Major 2 — agreed.** Reason codes plus named blockers; the why-trail belongs in the issue
and the ADR, not in a constant CI reads.

One qualification rather than a disagreement: none of this touches the fail-closed guard
itself, which is independent of how verbose the reasons are and of where the memo lives.
That is the part worth landing now.

## Sequence, filed

Your steps 1–7, mapped onto issues:

| your step | issue |
|---|---|
| 1. declared build/render zero-recognition gate | #1022 (already open, unchanged) |
| 3–5. raw evidence in `RecognitionResult`; model + lint as pure projections; remove the `CoverageState` memo and the ADR amendment | **#1025** (new) |
| 6. smaller manifest; migrate what #1022 unblocks | **#1026** (new) |
| 7. code-object call-count helper as an isolated test-quality change | to be split out of this PR |

Step 2 (lazy physical critique behind the single typed `BuildState` cache) is absorbed into
#1025: with the shoulder inventory a projection over the aggregate, lint has nothing left to
cache separately, so there is no second memo to relocate.

## Proposed disposition for this PR

Reduce it to the fail-closed manifest and its guard, with terse reasons — dropping the
`CoverageState` memo, the ADR 0017 §6 amendment, and the migration-history prose. The
`annotations.orchestrator.build_model` reuse stays (it is reuse, not a new owner). The
code-object spy work goes to its own PR, as the PR description already argued it should.

That leaves this PR doing one thing that is true independently of #1022, #1025 and #1026,
and lets the architecture land in the order you set out rather than being documented in the
order it happened to arrive.

pzfreo:
This is a strong response and recovers the intended architecture. In particular, the observation that `levels` is a projection parameter while `part.faces()` is the physical scan is the key: it gives us one evidence inventory with independent model/lint projections, rather than two recognition owners. Agreed that the `CoverageState` memo and its ADR amendment should both disappear, and that the manifest should retain terse, durable classifications rather than migration history.

Three details to tighten before implementing the split:

1. **The lazy aggregate cache is not fully absorbed by #1025.** Removing the shoulder-specific memo means lint has nothing *additional* to cache, but a declared drawing still needs to lazily construct the whole `RecognitionResult` when physical critique is explicitly requested, store/reuse it through typed `BuildState`, and run it at most once. Please make that explicit in #1022's acceptance criteria:
   - declared build/export without physical critique invokes zero recognisers;
   - first explicit physical lint obtains one aggregate;
   - repeated lint obtains none and returns equivalent results;
   - the cache is the typed `BuildState` owner, not another `Drawing`/lint private.

2. **#1022's description will become stale when this PR is reduced.** It currently says #1021 migrated flats, grooves and slot patterns. Please remove/update that claim as part of the rewrite.

3. **The current head is still the old architecture.** The plan resolves the review, but the diff does not yet. I'll re-review the reduced patch against these promises. In particular, `annotations.orchestrator.build_model` reuse should stay only if it consumes an inventory already owned by the aggregate and remains independently useful in the slim manifest PR.

Suggested landing order remains:

1. Reduce #1021 to the fail-closed manifest with terse reason codes/blockers; split the code-object spy into an isolated test-quality PR.
2. #1022: declared zero-recognition boundary plus the one lazy `BuildState` aggregate cache.
3. #1025: raw riser evidence and pure model/lint projections; no caller-narrowable lint inventory.
4. #1026: migrate the families unblocked by #1022 and shrink/remove obsolete deferrals.

With those refinements, I agree with the proposed recovery plan. The vision is back to **one recognition run, one authoritative physical result, multiple independent semantic consumers**.

pzfreo:
All three actioned, and the diff now matches the plan.

**1. The lazy aggregate cache — you're right, #1025 doesn't absorb it.** I conflated "lint
has nothing extra to cache" with "the declared path has its one aggregate", which are
different claims. It's now #1022's, with your four criteria as acceptance:

- declared build/export without physical critique invokes zero recognisers;
- first explicit physical lint obtains one aggregate;
- repeated lint obtains none, returning results equivalent to the first;
- typed `BuildState` owns it, not another `Drawing`/lint private (`Drawing._cyl_cache` is
  the existing violation, #1023).

Plus two the gate can't land without: declared turned parts keep axial critique
(`Analysis.prof` needs a declaration-sourced replacement, not a silent absence), and page
and scale selection unchanged with `step_zs` sourced from the declaration.

**2. #1022's stale claim removed.** The cost table now carries only the whole-aggregate
figures, with a note that the turned+flat number was measured at the revision that also
migrated the three families and loses its 3.1 ms component now that the migration is
reverted — re-measure before quoting. Which is the same discipline the earlier "~23 ms"
finding was about.

**3. The diff is reduced.** Force-pushed; previous head `cd10e6f` is preserved at
`backup/1021-round10`.

| | |
|---|---|
| files | 11 → 6 |
| `linting/coverage.py` | untouched vs `main` — memo, `critique_step_shoulders` and `test_lint_recognition_reuse.py` all gone |
| ADR 0017 | amendment withdrawn; §6 unchanged from `main` |
| `DEFERRED` | `Deferred(Deferral.CLASSIFICATION_GATED, blocker=1022)` — codes plus named blockers |
| fast tier | 2688 passed, 1 skipped, 2 xfailed; four lint gates clean |

Base is #1027 (the code-object helper, split out as you asked — the manifest guard counts
families with it). Retargets to `main` when that merges.

Two notes on judgment calls inside the reduction:

**`build_model` reuse kept.** It clears your bar: `pockets`/`pocket_patterns`/`pads` are
all `MIGRATED`, and unthreaded `build_part_model` re-detects them — so it's a real
de-duplication at that call site, not a preference between an inventory and its own list
copy. The 10-line comment hedging that point is cut to 5.

**"Lint runs no recogniser" could not survive the memo's removal**, so it isn't asserted.
The guard is now "lint runs no MIGRATED family", with the step-shoulder rescan pinned in
`_LINT_CALL_BUDGET` against #1025. Budgeted rather than hidden: a *new* lint-side scan
still fails, and the number can only go to zero. Asserting zero today would require the
memo you rejected, so recording the debt is the honest form.

One thing I'd flag rather than quietly keep: `_CLASSIFICATION_GATED` in the guard is now
derived from the reason codes instead of a hardcoded triple, so a family given that code
later is checked against a running build automatically — the reason-code change made the
guard stronger, not just shorter.

pzfreo:
Re-review of reduced head `3a98d99`: the architectural reset is real. The lint-owned memo and ADR amendment are gone, the manifest is now the focused deliverable, and #1022 has the missing lazy-aggregate acceptance criteria. Two substantive findings remain before I would call this clean.

### 1. Blocking — `CLASSIFICATION_GATED` names the wrong blocker

All three classification-gated families currently point at `blocker=1022`:

```python
recognise_chamfers / recognise_fillets / recognise_plates
    -> Deferred(Deferral.CLASSIFICATION_GATED, blocker=1022)
```

But #1022 only removes recognition from the **declared** path. It does not remove these families' automatic-build classification constraint: hoisting them unconditionally into `build_recognition_result` would still scan chamfers/fillets/plates on automatic turned builds where `build_part_model` currently skips them.

That is a different cost from `BUILD_MODEL_ONLY`. The old prose actually had this distinction right: these migrate when the orchestration carries/applies the rotational/turned classification. Either:

- make the blocker the issue that introduces classification-aware orchestration (possibly #1026, if its acceptance explicitly includes preserving these gates inside the aggregate), or
- split/file that prerequisite separately.

As written, the terse code has made the manifest more durable but less true: closing #1022 does not remove the constraint its blocker claims to remove. Please also correct #1026's statement that #1022 makes migration of all six “free”; it only makes the three `BUILD_MODEL_ONLY` migrations free on the declared path.

### 2. Major — the declared-path ratchet allows substitution

`test_a_declared_build_does_not_grow_new_recognition` asserts only:

```python
len(ran) <= 11
sum(ran.values()) <= 11
```

That allows one existing family to disappear and a new family to take its place: 11 families / 11 calls remains green. This contradicts the test's claim that a future migration cannot quietly add work to the declared path.

Make this a monotonic set/multiset ratchet rather than a cardinality ratchet, e.g. pin the current `{family: max_calls}` baseline and assert:

```python
set(ran) <= set(BASELINE)
all(ran[name] <= BASELINE[name] for name in ran)
```

That permits every intended downward move to zero while rejecting substitution and per-family growth. Mutation/counterexample: remove one currently-called family and add one currently-absent family; the present test stays green, while the corrected invariant must go red.

### Non-blocking observations

- The stacked base on #1027 is reasonable while the code-object helper is pending, but #1021 needs retargeting and one final diff/CI review after #1027 merges.
- `CLAUDE.md` says phase 1 “landed” within an unmerged PR. That reads correctly after merge, so I consider it a nit rather than a blocker.

Everything else I challenged in the previous review now matches the recovery plan. Fix these two invariants, let the stacked-base/retarget CI settle, and I expect the reduced PR to be reviewable cleanly.

pzfreo:
Both fixed in `e6d7b1d` (rebased on #1027's fixes).

**1. Blocking — the wrong blocker. Conceded; this is the sharper version of the finding.**
The reason code didn't just name the wrong issue, it lost a distinction the prose had right:
`BUILD_MODEL_ONLY` binds on the **declared** path and `CLASSIFICATION_GATED` binds on the
**automatic** one, and #1022 only touches the first. Shortening the reasons discarded that —
which is the failure mode the codes were introduced to prevent, appearing in the fix for it.

- **#1028** filed: carry the part classification in the orchestration so a family can be
  migrated *and* still not run for the class that discards it. The three now point there.
- The `Deferral` docstring says which path each constraint binds on, in as many words, so
  the next person cannot make the same substitution.
- **#1026** rewritten to the three `BUILD_MODEL_ONLY` families, with its earlier "all six"
  claim called out as wrong rather than quietly edited.

**2. Major — substitution.** Correct, and the cardinality form hid per-family growth too.
Now a per-family baseline: a family may leave, none may join, no count may rise.

Checked against your counterexample rather than asserted:

| mutation | `len<=11 / sum<=11` | per-family baseline |
|---|---|---|
| drop `recognise_slots`, add `recognise_grooves` | green | **red** (`added: ['recognise_grooves']`) |
| drop `recognise_slots`, `recognise_holes` 1 → 2 | green | **red** (`grew`) |

The baseline is measured, not assumed — the first draft budgeted `recognise_face_levels` at
2 where it actually runs once, which would have left exactly the slack this change is
supposed to remove.

**Nits.** CLAUDE.md's "landed" left as-is, reading post-merge as you suggested; it now lists
four open items, #1028 included. Retarget to `main` and final CI once #1027 goes in.

Fast tier 2692 passed, 1 skipped, 2 xfailed; four lint gates clean.

pzfreo:
## Re-review — no defects found

Both review findings are applied (`fe25241`), and I checked the guards by mutating the code rather than by re-reading them.

### The two fixes

**`blocker=1022` was wrong for `CLASSIFICATION_GATED`.** #1022 removes recognition from the *declared* path; the classification gate binds on the *automatic* path, where `build_part_model` skips chamfers/fillets/plates for turned parts. Hoisting them unconditionally would scan every automatic turned build for a discarded result, and #1022 does nothing about that. Filed **#1028** — the orchestration itself carries the classification — and the three families now point there. #1026's body was corrected too; it had the same error, claiming all six deferred families migrate once #1022 lands.

**The ratchet permitted substitution.** `len(ran) <= 11` / `sum(...) <= 11` stays green if one family stops running and another starts, or if two families each swap a call. `_DECLARED_BUILD_BASELINE` is now per-family, `added` is asserted before `grew` so a new family is named rather than folded into a total, and the numbers are measured rather than budgeted — my first draft had `recognise_face_levels: 2` where it actually runs once, which would have left exactly the slack the fix was for.

### Mutation results

| mutation | result |
|---|---|
| a MIGRATED family never called | `test_the_migrated_families_are_the_ones_the_orchestration_actually_runs` red |
| a DEFERRED family called by the orchestration | `test_no_deferred_family_is_reachable_from_the_orchestration` red |
| aggregate calls `recognise_pockets` but stores `()` | `test_the_aggregate_carries_what_its_recognisers_returned` red |
| `recognise_grooves` genuinely migrated into the aggregate | **5 red**, including `test_a_declared_build_does_not_grow_new_recognition` |

The last row is your substitution concern run for real rather than simulated in arithmetic: a genuine migration that loads the declared path trips the ratchet.

### Two things I confirmed rather than assumed

`_LINT_CALL_BUDGET`'s 1-per-lint is measured, not guessed — one lint gives `{recognise_step_shoulders: 1}`, three lints give 3. And the two blockers are coherent rather than arbitrary: `BUILD_MODEL_ONLY` → #1022 removes the cost and #1026 does the migration; `CLASSIFICATION_GATED` → #1028 removes the constraint and migrates in the same change.

Full fast tier green (2693 passed, 1 skipped, 2 xfailed); all four lint gates clean.

Issue #703, opened 2026-07-18
recognition: turned/grooves/flats re-run analyse_cylinders internally — add cyls= injection (>=4 scans per build)


pzfreo:
`recognise_turned_steps` (`turned.py:119`), `recognise_grooves` (`grooves.py:82`), and `recognise_flats` (`flats.py:108`) each open with `analyse_cylinders(part)` and offer no `cyls=` parameter — while `recognise_holes`/`recognise_bosses` take `cyls=` precisely "to avoid re-scanning the solid" (`_features.py:513`).

In the main build path `analyse_cylinders` therefore runs **at least 4 times per part**: `analysis.py:349` (injected into holes/bosses), `analysis.py:520` (inside recognise_turned_steps), `model/detect.py:307` (inside recognise_grooves), `model/detect.py:488` (inside recognise_flats). `build_part_model`'s kwargs also can't thread cyls down.

This recreates exactly the inconsistent-injection asymmetry ADR 0013 §2 was written to remove (holes/bosses injectable; turned/grooves/flats not) and cuts against ADR 0008 Am5's one-scan spirit. Not a literal contract breach — the package docstring classifies `analyse_cylinders` as substrate, not a recogniser — but the fix is mechanical: add a keyword-only `cyls=None` (compute if absent) to the three turned-stock recognisers, thread it from `analysis.py`/`detect.py`, and extend the contract test.

Related ADR 0013 hygiene (small, can ride along or go with the docs sweep #696):
- `recognise_hole_patterns` still declares bare `-> list` (`_features.py:1044`) against roadmap item 1a and the package's "never a bare list" promise — a `list[BoltCircle | LinearArray | RectGrid]` union annotation closes it.

From the 2026-07-18 architecture audit.

PR #1027, opened 2026-08-04
test: count recogniser calls by code object, closing the substrate guard's blind spot

pzfreo:
Split out of #1021 at review request: an isolated test-quality change, independent of
#1019's deliverable.

## The gap

`test_cylinder_scan_runs_once_per_build` (#703) guards that one build + lint scans the
solid for cylinders exactly once. It did that by monkeypatching `analyse_cylinders` across
a hand-kept list of modules — and the list had fallen behind the source. Neither
`recognition/result.py` nor `linting/coverage.py` was on it, and both bind the name **by
value** on live paths. So dropping the substrate injection from `build_recognition_result`
makes the aggregate re-scan every solid on every build, with the **entire fast tier green**.

Verified both directions on this branch:

| fixture | injection removed from `build_recognition_result` |
|---|---|
| old (module patch list) | **4 passed** |
| new (code object) | **1 failed**, 3 passed |

## Why not just add the two missing modules

Because that is a rung, not the fix. Five review rounds on #1021 climbed exactly this
ladder: modules nobody had listed → an *aliased* import (`as _rescan`) → a function held in
a **dict**. The next rungs are a class attribute, a closure cell, a default argument, a
`partial`, a late import. A patch list can only cover the binding forms someone thought of,
and every miss is silent — the call happens, the count does not move, the guard reports
success. CLAUDE.md §8: if the same class of bug appears twice, reconsider the approach.

A **code object** cannot be re-bound. Every call reaches the same `__code__` however the
caller got hold of the function, so counting with a profile hook closes container / class
attribute / closure / default arg / partial / alias / late import at once, and needs no
module list kept in step with the source.

The hook is thread-local and saves/restores any profiler already installed, so it composes
with a `--cov` run (coverage drives `sys.settrace`/`sys.monitoring`, a separate mechanism).
Verified under `--cov`.

`counting_calls` lands in `tests/conftest.py` rather than in the one suite because the ADR
0017 manifest guard (#1019) counts recogniser families the same way.

No source changes. Part of epic #1018.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,10 +33,21 @@ def counting_calls(functions: Mapping[str, Callable[..., object]]):
     blind spots must not open one.
 
     The hook is thread-local — only the calling thread is measured, which is what the tests
-    want. Any profiler already installed is **chained**, not suspended: `sys.setprofile`
-    holds one function, so an outer profiler (a nested ``counting_calls``, a plugin) would
-    otherwise go blind for the duration and never know. Coverage is unaffected either way —
-    it drives ``sys.settrace``/``sys.monitoring``, a separate mechanism.
+    want. A Python-level profiler already installed is **chained**, not suspended:
+    ``sys.setprofile`` holds one function, so an outer profiler (a nested ``counting_calls``,
+    a plugin) would otherwise go blind for the duration and never know.
+
+    A C-level profiler cannot be chained, and this refuses to run rather than pretend.
+    Under cProfile on Python <= 3.11, ``sys.getprofile()`` returns the ``Profile`` object,
+    which is not callable: chaining to it raises ``TypeError`` mid-context, and so does
+    restoring it on the way out. (On 3.12+ cProfile moved to ``sys.monitoring``, so
+    ``getprofile()`` is ``None`` and there is nothing to collide with.) Refusing on entry is
+    the same fail-closed choice as the duplicate-code-object check above: the alternatives
+    are a crash from inside a context manager, or counts taken with the outer profiler
+    silently switched off.
+
+    Coverage is unaffected throughout — it drives ``sys.settrace``/``sys.monitoring``, a
+    separate mechanism.
     """
     by_code: dict = {}
     for name, fn in functions.items():
@@ -50,6 +61,13 @@ def counting_calls(functions: Mapping[str, Callable[..., object]]):
         by_code[code] = name
     counts: dict[str, int] = {}
     previous = sys.getprofile()
+    if previous is not None and not callable(previous):
+        raise RuntimeError(
+            f"a non-callable profiler is installed ({type(previous).__name__}) — a C-level "
+            "profiler such as cProfile on Python <= 3.11. counting_calls can neither chain "
+            "to it nor restore it, so it refuses rather than crash mid-context or return "
+            "counts taken with the outer profiler silently switched off."
+        )
 
     def hook(frame, event, arg):
         if event == "call":
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,11 +6,12 @@ guard (#1019) counts recogniser families the same way.
 """
 
 import sys
+from collections.abc import Callable, Mapping
 from contextlib import contextmanager
 
 
 @contextmanager
-def counting_calls(functions: dict):
+def counting_calls(functions: Mapping[str, Callable[..., object]]):
     """Count invocations of *functions* (``{name: callable}``) by CODE OBJECT.
 
     Counted with a profile hook rather than by installing spies over the modules that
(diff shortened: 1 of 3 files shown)


pzfreo:
Adversarial review of `0780bcb`: the code-object approach is the right replacement for the binding patch list, and the demonstrated aggregate-rescan mutation is exactly the counterexample this guard needed. Full CI is green. Two focused findings remain in the shared helper before I would use it as the foundation for #1021's fail-closed manifest.

### 1. Major — an existing profiler is restored, but not composed while counting

`counting_calls` does:

```python
previous = sys.getprofile()
sys.setprofile(hook)
...
sys.setprofile(previous)
```

This preserves the profiler *after* the context, but suspends it *during* the context. That is observable for nested `counting_calls` contexts and for any pytest/plugin instrumentation using `sys.setprofile`: the outer profiler silently misses everything inside the inner guard. The docstring's “saves/restores any profiler already installed, so it composes” is therefore stronger than the implementation.

Please have `hook` delegate each event to `previous` when non-`None` (while still doing its own count), and add a counterexample test proving an installed/outer profiler sees the target call during the context and is restored afterward. Coverage compatibility remains separate—its trace/monitoring mechanism is already proven by CI.

### 2. Major — duplicate code objects silently overwrite one another

```python
by_code = {fn.__code__: name for name, fn in functions.items()}
```

Two distinct function objects can share one code object—for example, two closures returned by the same factory. The dict then silently keeps only the last name, and calls to either function are attributed to that one. A helper whose purpose is to eliminate silent blind spots should fail closed here.

Please either reject duplicate code objects with a clear `ValueError` (my preference; attribution is inherently ambiguous), or explicitly represent one code object mapping to all intended names if that is the desired contract. Add a small factory-produced counterexample so the failure mode cannot return.

### Nits

- A typed signature such as `Mapping[str, Callable[..., object]]` would make the helper contract clearer, but this is non-blocking.
- Directly importing from `conftest` works under this suite; I don't think moving the helper solely for style is warranted.

With profiler chaining and duplicate-code fail-closed validation, I consider the PR's approach clean and appropriately isolated from #1021.

pzfreo:
Both fixed in `9f68202`, both mutation-validated.

**1. Profiler chaining.** You're right that the docstring's "composes" was stronger than the
code — `sys.setprofile` holds one function, so the previous profiler was preserved *after*
and suspended *during*. `hook` now delegates every event to `previous` when there is one.

`test_an_outer_profiler_still_sees_calls_made_inside` installs an outer profiler, asserts it
records the target call made inside the context, that the hook was actually installed
(`inside is not outer`), and that `sys.getprofile() is outer` afterwards. Dropping the
delegation turns it red.

**2. Duplicate code objects.** `ValueError` naming both entries and the shared
`co_name`/file/line. Raised during validation, *before* `sys.setprofile` — so a refusal
cannot leave the interpreter profiling into a context that has already unwound, which
`test_the_refusal_leaves_the_profiler_untouched` pins separately. Counterexample is a
closure factory, with `first.__code__ is second.__code__` asserted in the test so the
fixture cannot silently stop demonstrating the case.

Restoring the `{fn.__code__: name}` comprehension reddens both duplicate tests.

Took the typing nit too — `Mapping[str, Callable[..., object]]`.

Fast tier 2692 passed, 1 skipped, 2 xfailed; four lint gates clean.

pzfreo:
## Re-review — one real defect, mine, now fixed

I reviewed the fixes rather than assuming them. The chaining fix I added in `9f68202` was itself broken on two supported Pythons.

### The defect

`previous(frame, event, arg)` assumes the previous profiler is callable. Under cProfile on Python ≤ 3.11, `sys.getprofile()` returns the `Profile` **object**, which is not callable. Verified on 3.10 (`uv run --no-project --python 3.10`):

```
py (3, 10) getprofile -> Profile
TypeError: 'Profile' object is not callable
```

Both paths blow up — the chain call inside the context, and `sys.setprofile(previous)` in the `finally`. The restore half predates my fix; the chain half I added. On 3.12+ neither arises: cProfile moved to `sys.monitoring`, so `getprofile()` is `None` and there is nothing to collide with. That is exactly why the local suite stayed green and why I only found it by running 3.10 on purpose. `requires-python = ">=3.10,<3.15"`, and 3.10 is in the CI matrix.

### The fix (`fcfe180`)

Refuse on entry, before the hook is installed, with a message that names the cause. Chaining a C-level profiler is not possible through the public API, so the three options were: crash from inside a context manager, return counts taken with the outer profiler silently switched off, or refuse. Refusing is the same fail-closed choice as the duplicate-code-object check — a helper whose purpose is removing silent blind spots must not open one.

The docstring claimed *any* previously-installed profiler is chained. It now says which kind, and what happens to the other kind.

New test `test_a_c_level_profiler_is_refused_rather_than_crashed_into` fakes `sys.getprofile`/`sys.setprofile` (the real `sys.setprofile` rejects non-callables on 3.12, so the object cannot be installed for real) and asserts the hook was never installed. Mutation-validated: disabling the guard turns it red.

### Also fixed

`test_an_outer_profiler_still_sees_calls_made_inside` restored `None` in its `finally` rather than the profiler that was there — the same clobbering defect the test exists to catch, committed by the test. Now restores `before`.

Full fast tier green (2693 passed, 1 skipped, 2 xfailed); ruff check, ruff format, mypy, codespell all clean.

Issue #1026, opened 2026-08-04
Migrate the three BUILD_MODEL_ONLY recogniser families and shrink the manifest


pzfreo:
`recognition/result.py`'s `DEFERRED` map holds three `BUILD_MODEL_ONLY` families —
`recognise_grooves`, `recognise_flats`, `recognise_slot_patterns` — for one reason: their
sole engine consumer is `build_part_model`, which the ADR 0011 declared path skips. The
aggregate runs unconditionally from `_analyse`, so hoisting one removes no scan from the
detected path and adds one to the declared path, where the result is discarded.

**Only these three.** An earlier draft of this issue said all six deferred families migrate
here; that was wrong, and #1021's review caught it. The other three
(`recognise_chamfers`, `recognise_fillets`, `recognise_plates`) are `CLASSIFICATION_GATED`,
which is a constraint on the **automatic** path — `build_part_model` skips them for turned
parts, so hoisting them unconditionally would scan every automatic turned build for a
discarded result. #1022 does nothing about that. They are **#1028**.

#1022 removes the cost side for these three: once a declared build performs no
recognition, hoisting is free and ADR 0017 completeness is reason enough on its own.

Also in scope: `recognise_slots` and `recognise_pocket_patterns` are already `MIGRATED` and
have the same property (nothing reads `Analysis.pocket_patterns` at all). They arrived with
#1020, before the principle was articulated; #1021 declined to un-migrate them for a few ms
that #1022 removes wholesale. This issue closes the inconsistency by making the list
uniform rather than annotated.

## Shrink the map

The reason **codes** landed with #1021 — `Deferred(Deferral.BUILD_MODEL_ONLY,
blocker=1026)` in place of the paragraphs of migration history that went stale inside one
review cycle. What is left for this issue is the map itself: three entries leave `DEFERRED`
for `MIGRATED`, and the comment above it stops needing to explain the `recognise_slots` /
`recognise_pocket_patterns` inconsistency because there no longer is one.

The fail-closed guard stays exactly as it is — a family in neither `MIGRATED` nor
`DEFERRED` fails CI. That is the part of the manifest with durable value.

## Acceptance

- `recognise_grooves`, `recognise_flats` and `recognise_slot_patterns` move to `MIGRATED`
  and are called once each by `build_recognition_result`.
- `DEFERRED` retains only `recognise_step_shoulders` (#1025), `recognise_face_levels` (no
[...]

Issue #1025, opened 2026-08-04
Split recognise_step_shoulders into raw riser evidence and pure per-consumer projections


pzfreo:
`recognise_step_shoulders(part, *, levels)` performs a full `part.faces()` scan and uses
`levels` only to *select* which risers qualify (plus an early-out when `levels` is empty).
The scan is the cost; the level set is a projection parameter.

#1021 read the differing level sets as a reason the family cannot be hoisted, and memoised
the coverage-side call on `CoverageState` instead — with an ADR 0017 §6 amendment to permit
it. Architectural review of #1021 rejects that shape: it makes lint a second recognition
owner, where ADR 0017's waist is one authoritative physical result with independent
*semantic* consumers.

## The split

```text
solid
  └─ RecognitionResult: raw riser/level evidence (one scan)
       ├─ model projection   (levels filtered by plate + edge-open-pocket ownership)
       └─ lint projection    (unfiltered analysis levels)
```

Both consumers project over fixed evidence; neither rescans. The ownership filtering stays
in `model/detect.py` where it belongs — a projection choice, not a recognition one, so no
drafting policy moves below the IR. `declare.step` is the same operation against a
different source solid.

This dissolves the deferral's stated blocker. #1021 argued a keyed memo "has no home below
both `model/` and `linting/`"; under the split the shared thing stops being a keyed cache
and becomes evidence, which already has a home.

## What this removes

- `recognise_step_shoulders`' `DEFERRED` entry — the last `CALLER_SPECIFIC_INPUT` one.
- The per-lint rescan it budgets (`_LINT_CALL_BUDGET` in `tests/test_recognition_manifest.py`
  goes to empty), which is phase 0's third guard: repeated lint reruns no recognition.

**Corrected 2026-08-04.** This section originally also listed removing
`CoverageState.step_shoulders` + its keyed memo, and withdrawing an ADR 0017 §6 amendment.
**Neither exists.** Both were written when #1021 still carried them; its reshape dropped the
memo, the `critique_step_shoulders` fallback and the ADR amendment before merge, so there is
nothing to remove. `linting/coverage.py` calls `recognise_step_shoulders` directly on every
pass — the rescan this issue ends — and ADR 0017 §6 still says one-result ownership, unamended.

The split below is unaffected: it was never justified by the memo's existence, and the
architectural argument against lint owning a second recognition result stands whether or not
that memo ever landed.

## Fold in: `step_zs=[]` silences the critique

[...]

PR #1029, opened 2026-08-04
feat(#1022): a declared build recognises nothing

### LINKED ISSUE #1022: Declared builds still recognise: the ADR 0011 gate needs sizing to source its own inputs
ADR 0011 says a caller-supplied `PartModel` **skips detection**. It does not. `_analyse`
calls `build_recognition_result` unconditionally (`analysis.py:516`), before it knows
whether a model was declared, so a declared build recognises the full aggregate and then
discards most of it — `build_part_model` is skipped on that path.

ADR 0017 §6 restates the same requirement for the aggregate. This is the phase-2 gate in
epic #1018.

## The gate is not one `if`

Naively wrapping the call in `if layout_model is None` breaks page selection, because
sizing consumes recognition even on the declared path:

- `step_zs` comes from `recognition.turned_steps` (via `TurnedProfile.from_steps`) or from
  `step_level_zs(part)` — and feeds `_legible_steps` / the step count that page and scale
  selection converge on (`analysis.py:521-527`, `624-645`, `741-743`).
- `Analysis.prof` feeds `lint_axial_coverage`, so gating recognition without a replacement
  silently disables axial critique for declared turned parts.

So the gate needs sizing to take those from the declaration first, plus the lazy
critique-recognition policy ADR 0017 describes for lint. Both halves are in scope here.

## The lazy aggregate cache belongs here, not to #1025

Gating the eager call is only half the boundary. A declared drawing whose caller explicitly
asks for physical critique still needs the aggregate — so it must be constructed lazily, at
most once, and owned by the typed `BuildState` rather than by a second private cache.

#1025 removes lint's shoulder-specific memo, which means lint has nothing *additional* to
cache; it does not give the declared path its one lazy aggregate. That is this issue's work.

## Cost today, measured

| part | whole aggregate |
|---|---|
| turned + flat | ~21 ms |
| prismatic | ~32 ms |

(Measured on the #1021 branch at the revision that also migrated `flats`, `grooves` and
`slot_patterns`, where the totals read 23.7 ms and 31.9 ms; that migration was reverted in
review, so the turned+flat figure loses its 3.1 ms component. Re-measure before quoting.)

The whole aggregate has been unconditional since #1020, and this gate removes all of it on
the declared path. That is why the manifest defers rather than adds per-family conditionals
— scattered gating is exactly what ADR 0017 exists to remove, and it would be deleted again
here.

## Acceptance

- A declared **build** invokes **zero** recognisers — asserted by instrumenting the
  recogniser families and building from a `Sheet`/`PartModel`, not by reading the source.
- **Amended 2026-08-04 (PR #1029).** This criterion originally said "build/export". Export
  is now excluded, deliberately: `Drawing.export` calls `_lint_and_log`, whose warnings are
  the coverage diagnostics users actually read. Reaching zero there means lint-and-log
  requesting `physical=False`, which would stop **every** export — detected builds included,
  where the critique is free because the inventory already exists — from reporting coverage
  problems. That trades a real user-facing diagnostic for a benchmark number.

  So exporting a declared drawing buys **one** aggregate, on the same lazy path as an
  explicit lint, and the guard asserts once-and-no-more rather than zero. A caller who wants
  a declared export to recognise nothing has the same lever as before: do not ask for the
  critique.

  Recorded here rather than silently diverging, per CLAUDE.md — an adversarial review of
  #1029 correctly called the original PR a *reinterpretation* of this line rather than an
  implementation of it. It also found a genuine bug behind the disagreement: `export`
  linted **before** validating its formats, so `export(..., formats=("bogus",))` recognised
  the whole solid and only then raised. That ordering is fixed regardless of this amendment.
- The first explicit physical lint on a declared drawing obtains **one** aggregate.
- Repeated lint obtains **none**, and returns results equivalent to the first.
- That aggregate is owned by the typed `BuildState`, not by another `Drawing` or lint
  private (ADR 0017 §6; `Drawing._cyl_cache` is the existing violation, #1023).
- Declared turned parts keep axial critique: `Analysis.prof` has a declaration-sourced
  replacement rather than a silently absent one.
- Page and scale selection are unchanged on declared builds, with `step_zs` sourced from
  the declaration.

Parent: #1018. Unblocks #1026.

pzfreo:
Closes #1022. Stacked on nothing — branches off `main`.

## What held it up

ADR 0011 says a caller-supplied `PartModel` skips detection; ADR 0017 §6 restates it for the aggregate. Neither held — `_analyse` ran `build_recognition_result` before it knew whether a model had been declared.

The issue predicted the gate would not be one `if`, and that was right, but not complete. Four things had to move:

**Sizing.** Page/scale selection read `prof` and `step_zs` off the recognition even for a declared model. `_declared_turned_profile` / `_declared_step_zs` source them from the declaration, mirroring the detected path's two branches (including the 0.6 mm end-exclusion) so selection is unchanged.

**The repair loop.** Gating `_analyse` alone left the declared build at *ten* families — all reached through `build_drawing` → `repair()` → `lint()`. Repair acts on `dim_inside_part` and nothing else (ADR 0002) but requested the full critique every iteration. `Drawing.lint(physical=False)` asks for the placement half. That also tightens repair's own net-worsened check: coverage issues cannot change from re-placing a dimension, so counting them only diluted the ratio.

**Critique's inventory.** On the declared path `a.holes` is empty because nothing looked, not because the part has none — feeding that to coverage reports every real hole as uncovered. `BuildState.ensure_recognition()` builds one lazily, at most once, in the typed build state. A lint-side or `Drawing`-side memo would make critique a second recognition owner, which is what ADR 0017 exists to remove and what #1021's review rejected.

**ADR 0015.** Lint takes `step_zs`/`pads`/`pockets` from that aggregate, never from `Analysis` on the declared path. My first cut got this wrong and four tests caught it: critique inventorying from the model made lint blind to exactly the geometry a sparse declaration omits — the case `unrecognised_defining_geometry` exists to report.

## Measured, not asserted from the source

| | before | after |
|---|---|---|
| declared build | 10 recognisers | **0** |
| first critique on a declared drawing | — | 1 aggregate |
| repeated critique | — | 0 (bar the `recognise_step_shoulders` deferral, #1025) |
| detected build | 16 | 16 |

## Two calls worth reviewing

**Export pays for one aggregate, deliberately.** `export` calls `_lint_and_log`, whose warnings users read. Making it skip coverage would stop *every* export on *both* paths from reporting coverage problems — trading a real diagnostic for a benchmark number. So "build/export without physical critique invokes zero" is met by build; export legitimately buys one aggregate, and the guard asserts once-and-no-more rather than restating the acceptance.

**`recognise_face_levels` migrates into the aggregate** as `step_levels`. Its `NO_INDEPENDENT_CONSUMER` deferral stopped being true the moment declared-path critique needed the geometry ladder; without the migration lint rescanned it once per pass. `DEFERRED` is down to seven. This is the manifest doing its job — a deferral whose stated reason expired.

## A latent bug on both paths

`_est_right_strip_depth` reserved one slot per ladder step, but `render_boss_heights` puts every Z-axis boss height in that same strip — sized for one occupant class, solved for two. Detected builds hid it inside the ladder's slack. A declared model with a boss and no ladder reserved one slot, needed two, and dropped the height **silently** (`on_drop` is a no-op there so lint reports the omission once rather than twice). `_n_right_strip_boss_heights` counts the real occupants, mirroring the renderer's own guards (turned models render none; X/Y bosses go to the *above* strips).

This changes detected-path reservations wherever a Z boss has a height. That is the fix working, not a side effect — ADR 0004 drops byte-identity — but it is the widest-blast-radius edit here and deserves the most scrutiny.

Known conservative gap: `is_rotational` is geometry-derived and not visible to the composer, so a rotational part declaring a Z boss over-reserves by one slot. Safe direction, and the same call the authored-Z-dim branch already makes.

## Verification

- Targeted: 1221 passed across the eleven affected suites; recognition suites re-run green after the last fix. **Full tier not run** (per instruction) — CI covers it.
- Four gates clean: ruff check, ruff format, mypy, codespell.
- **Mutation-validated**: reverting the gate → 7 red; reverting the repair split → 5 red; reverting the strip sizing → the boss-height guard red.
- `test_a_declared_build_does_not_grow_new_recognition` → `test_a_declared_build_recognises_nothing`: the #1021 ratchet reached its target.
- `test_the_migrated_families_are_the_ones_the_orchestration_actually_runs` moves off module-binding spies onto `counting_calls`. Migrating a family reached *indirectly* (via `step_level_zs`) made the old spy raise `AttributeError` rather than fail — the blind spot #1027 was written to close, still present in that one test.

Unblocks #1026.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -388,18 +388,31 @@ Current ADRs:
   deferral is a `Deferral` reason **code** plus the issue that removes it, not a paragraph:
   prose in a constant CI reads goes stale silently, and the first cut's did. The why-trail
   lives in the blocking issue.
-  Four things are still open, in this order. **#1022** — the ADR 0011 declared-path gate:
-  a declared build still recognises, and gating it is *not* just skipping the call, because
-  page/scale selection reads `step_zs` and the turned profile off that recognition even
-  when the model was declared, so sizing has to source them from the declaration first; the
-  same issue owns the one lazily-built `BuildState` aggregate a declared drawing needs when
-  physical critique *is* asked for. **#1025** — `recognise_step_shoulders` splits into raw
+  **#1022** landed the **ADR 0011 declared-path gate**: a declared build now recognises
+  **nothing**. It was not one `if` — sizing sources the turned profile and step ladder from
+  the declaration (`_declared_turned_profile` / `_declared_step_zs` in `analysis.py`), and
+  the lint→repair loop stopped asking for the feature-coverage half it never used
+  (`Drawing.lint(physical=False)`; repair acts on `dim_inside_part` alone, ADR 0002).
+  Critique on that path still needs an inventory, so `BuildState.ensure_recognition()` builds
+  one lazily, once — in the typed build state, not a lint- or `Drawing`-side memo, which
+  would make critique a second recognition owner. Exporting a declared drawing pays for that
+  one aggregate by design: `export` logs the coverage critique, and suppressing it to reach
+  "zero" would trade a user-facing diagnostic for a benchmark number. Two consequences worth
+  knowing: lint takes `step_zs`/`pads`/`pockets` from the **aggregate**, never from
+  `Analysis` on the declared path (ADR 0015 — critique must not inventory from the model);
+  and `recognise_face_levels` migrated into the aggregate as `step_levels`, its
+  `NO_INDEPENDENT_CONSUMER` deferral having stopped being true the moment declared-path
+  critique needed the geometry ladder. The gate also exposed a latent strip-sizing bug on
+  *both* paths: `_est_right_strip_depth` counted ladder steps only, while boss heights share
+  that strip — detected builds hid it in the ladder's slack (`_n_right_strip_boss_heights`,
+  `compose.py`).
+  Three things are still open, in this order. **#1025** — `recognise_step_shoulders` splits into raw
   riser evidence in the aggregate plus a pure projection per consumer, because its `levels=`
   input differs between model construction (filtered by plate and pocket ownership) and
   critique (unfiltered — ADR 0015 forbids lint taking its inventory from the model); until
   then lint rescans on every pass, budgeted rather than memoised, since a lint-owned memo
   would be a second recognition owner. **#1026** — migrate the three `BUILD_MODEL_ONLY`
-  families #1022 unblocks, and shrink the manifest. **#1028** — the classification gate is
+  families #1022 has now unblocked, and shrink the manifest. **#1028** — the classification gate is
   a *different* blocker on the automatic path, untouched by #1022: `chamfers`/`fillets`/
   `plates` migrate only once the orchestration itself carries the part classification, so
   hoisting them stops meaning "scan every turned build for a discarded result". Beyond the
--- a/src/draftwright/_core.py
+++ b/src/draftwright/_core.py
@@ -822,7 +822,12 @@ class Analysis:
     """
 
     part: Shape
-    recognition: RecognitionResult
+    #: The ADR 0017 aggregate, or ``None`` on a DECLARED build — which recognises nothing
+    #: (ADR 0011 / #1022).  ``None`` means "not detected", never "detected and empty": a
+    #: consumer needing an inventory on that path must go through the lazy
+    #: ``Drawing._recognition()``, which builds one on demand rather than reading an absence
+    #: as an answer.
+    recognition: RecognitionResult | None
     bb: BoundBox
     x_size: float
     y_size: float
--- a/src/draftwright/analysis.py
+++ b/src/draftwright/analysis.py
@@ -44,14 +44,15 @@ from draftwright.compose import (
     choose_scale,
 )
 from draftwright.model import build_part_model
-from draftwright.model.ir import Datum, PartModel, StepFeature
+from draftwright.model.ir import Datum, PartModel, StepFeature, StepLevelFeature
 from draftwright.model.planner import plan_dimensions
 from draftwright.recognition import (
+    RecognitionResult,
     TurnedProfile,
+    TurnedStep,
     analyse_cylinders,
     build_recognition_result,
     full_cylinders,
-    step_level_zs,
 )
 
 _log = logging.getLogger(__name__)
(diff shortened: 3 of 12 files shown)


pzfreo:

Three findings. Two were real defects and are fixed in `de45617`; the third was a fair reading of the spec that I have amended rather than diverged from.

### 1. `export` linted before validating its formats — fixed

Real, and the review found it as a sub-point of a larger disagreement. `export(..., formats=("bogus",))` on a declared drawing ran the critique — building the lazy aggregate, scanning the whole solid — and *only then* raised `ValueError`.

This was harmless before #1022, when the critique was free because a detected build already had its inventory. Making the critique the trigger for lazy recognition turned it into real work for a call that was never going to produce a drawing. Validation now precedes the lint, which is the same ordering the surrounding code already argues for with respect to the deprecation warning ("report the fault that matters").

Guarded by `test_a_rejected_export_does_not_recognise_first`, mutation-validated: restoring the old order turns it red.

### 2. The repeated-lint guard asserted half the criterion — fixed

Also real, and my mistake. #1022 requires repeated lint to obtain no new aggregate **and** return results equivalent to the first. My test counted recognisers and threw away all three return values.

The review's scenario is exactly right: an implementation whose later lints reuse the cached aggregate but return only placement issues — or nothing — would have stayed green. Now compares sorted `(code, severity, message)` tuples across three lints. Counting and equivalence are separate claims, and the second is the one a user would actually notice.

### 3. Export vs the zero-recognition criterion — spec amended

The review's charge is that treating export as physical critique **reinterprets** the acceptance criterion rather than implementing it. That is correct, and worth stating plainly: I did change the requirement, and disclosing it in the PR body is not the same as it being right.

The criterion said "a declared build/**export** without physical critique invokes zero recognisers". Reaching literal zero on export means `_lint_and_log` requesting `physical=False` — which would stop **every** export, detected builds included, from reporting coverage problems in its log. On the detected path that critique costs nothing, because the inventory already exists. So conforming would remove a real user-facing diagnostic from a path that was never the problem, to fix a number on the path that was.

Per CLAUDE.md, a change that conflicts with a recorded decision amends it in the same PR rather than diverging quietly. #1022's acceptance is amended accordingly — export buys **one** aggregate on the same lazy path as an explicit lint, and the guard asserts once-and-no-more.

Finding 2 of the review noted my export test would block a conformant fix. That is true, and it is now the intended contract rather than an accident.

### Not disputed

The review found the `_analyse` gate correctly placed and the declaration-sourced turned profile / step levels consistent with the ADR boundary.

### Caveat on the review itself

Codex could not execute the suite (`bwrap` cannot create a namespace in this environment), so its findings come from tracing the source. Both defects it found were reproduced and mutation-validated here before fixing.

### Verification after the fixes

- 41 passed across the recognition and export-touching suites; the gate suite is 13 green.
- Four lint gates clean.

Issue #1028, opened 2026-08-04
Carry the part classification in the recognition orchestration so the gated families can migrate


pzfreo:
`recognise_chamfers`, `recognise_fillets` and `recognise_plates` are `DEFERRED` with
`Deferral.CLASSIFICATION_GATED`: `build_part_model` runs each only for one part class
(chamfers and fillets on `rotational is None`; plates additionally on `prof is None`), so
hoisting them into `build_recognition_result` unconditionally would scan every **automatic
turned build** for a result that branch discards.

This is not what #1022 removes. #1022 stops the **declared** path recognising; the
classification constraint is on the **automatic** path and survives it untouched. The three
were briefly pointed at `blocker=1022` in #1021 — corrected to point here, because closing
#1022 would otherwise claim to unblock a migration it does not.

## What is needed

The aggregate has to carry the part classification and apply it, so a family can be
migrated *and* still not run for the class that discards it:

- the classification (`rotational`, `prof`) is computed once and lives on the
  `RecognitionResult` alongside the inventories, rather than being an argument
  `build_part_model` derives separately;
- families whose applicability is classification-dependent are recognised conditionally
  **inside** the one orchestration.

That second point is the part worth being careful about: per-family conditionals scattered
across call sites are exactly what ADR 0017 exists to remove, so the gate has to be a
property of the orchestration, not a set of `if`s at the boundary. The distinction is that
one place decides once, rather than each consumer deciding for itself.

## Acceptance

- The three families move from `DEFERRED` to `MIGRATED`.
- An automatic **turned** build still does not call them —
  `tests/test_recognition_manifest.py::test_a_classification_gated_reason_is_true_of_the_build`
  becomes a check on the aggregate rather than on `build_part_model`, and must stay red if
  the gate is deleted.
- An automatic **prismatic** build calls each exactly once, from the aggregate — no rescan
  in `build_part_model`.
- No net increase in recogniser calls on either build path, measured per family.

ADR 0017 phase 2. Parent: #1018. Related: #1026 (the `BUILD_MODEL_ONLY` families, which
#1022 *does* unblock).

PR #1030, opened 2026-08-04
feat(#1026): migrate the BUILD_MODEL_ONLY families and shrink the manifest

### LINKED ISSUE #1026: Migrate the three BUILD_MODEL_ONLY recogniser families and shrink the manifest

pzfreo:
Closes #1026. Branches off `main` (after #1029 merged).

## Why now

`recognise_grooves`, `recognise_flats` and `recognise_slot_patterns` were deferred for one reason, and it has expired.

Their sole engine consumer is `build_part_model`, which the ADR 0011 declared path skips. While the aggregate ran unconditionally from `_analyse`, hoisting one removed no scan from the detected path and *added* one to the declared path, where the result is discarded. #1021's review was right to defer them on that basis, and right to reject the earlier revision that migrated them anyway.

#1022 removed the cost side: a declared build now recognises nothing at all. So the deferral's stated reason no longer describes reality, and ADR 0017 completeness is reason enough on its own.

## What changed

All three move to `MIGRATED`, and `build_part_model` takes them by injection instead of detecting them itself — `slot_patterns` already had the parameter, `grooves` and `flats` needed one.

`slot_patterns` is a *derived* family, so the aggregate hands it the accepted slot records rather than letting it rediscover them. That is the rule the other two pattern families already follow, and the orchestration-injection test asserts identity, not equality, so a fresh-but-equal list fails.

## Shrinking the map

`DEFERRED` is down to **four**, and every remaining entry binds on the **automatic** path, which #1022 did not touch:

| family | reason | blocker |
|---|---|---|
| `recognise_chamfers` / `recognise_fillets` / `recognise_plates` | `CLASSIFICATION_GATED` | #1028 |
| `recognise_step_shoulders` | `CALLER_SPECIFIC_INPUT` | #1025 |

The comment block explaining why `recognise_slots` and `recognise_pocket_patterns` sat in `MIGRATED` despite having the `BUILD_MODEL_ONLY` property is deleted, because the inconsistency it apologised for is gone.

The `BUILD_MODEL_ONLY` enum member survives deliberately — a future family could earn that deferral. What does not survive is a *deferral* justified by a cost that no longer exists.

## Measured, per family

| path | before | after |
|---|---|---|
| detected | 16 families, one call each | 16 families, one call each |
| declared | 0 | 0 |

The migration adds no duplicate scan on the detected path, and #1022's gate still holds on the declared one. Both measured by instrumenting the recogniser families, not read off the source.

## Verification

- 1252 passed across `make_drawing`, `declare`, `sheet_emit`, `score` and the recognition suites. `score.py` is included deliberately: it calls two of these three directly and is meant to stay off the aggregate, so a migration that quietly rerouted it would show up there.
- Four lint gates clean.
- **Mutation-validated**: making `build_part_model` ignore the injected grooves and re-detect turns both rescan guards red (`test_an_automatic_build_runs_each_family_exactly_once_...` and `test_model_construction_does_not_rescan_a_migrated_family`).

Parent: #1018. Siblings still open: #1025, #1028.


pzfreo:

Two findings. One accepted and acted on; one declined, with reasoning.

### 2 (Low) — the value oracle restates the implementation — **accepted, fixed in `e46b674`**

Right, and it points at the exact claim this PR rests on. `_expected_inventory` derives slot patterns from `slots` and calls grooves/flats with the same cylinder substrate the orchestration does, so it checks that the aggregate *stored* what its recognisers returned — not that the *inputs* were the right ones. A recogniser needing a filtered or reconciled set rather than the raw accepted records would have production and oracle agreeing on the same wrong answer.

Now tested directly: `build_part_model(part)` against `build_part_model(part, <every inventory injected from the aggregate>)`, on a part carrying both a groove and a flat, and on a slot lattice — comparing the feature *kinds* and the feature *values*.

I validated it against the review's own scenario rather than an easier one. Feeding `recognise_grooves` the wrong cylinder substrate in **both** production and the oracle leaves the oracle's equality assertion green; it fails only through its separate non-emptiness guard, and only because that specific error happens to empty the result. A wrong input yielding different-but-non-empty records would pass it. The new test fails either way, because it compares against an independent path rather than a restatement of the code under test.

### 1 (Medium) — three required fields break the public `RecognitionResult` constructor — **declined**

The premise does not hold, on four counts:

1. **There is exactly one construction site in the repo** — `build_recognition_result`. Nothing else constructs a `RecognitionResult`, including the tests: the orchestration test monkeypatches the *recognisers*, not the constructor.
2. **The type is documented as in-flux.** Its own docstring says the aggregate "is intentionally incomplete in its first migration slice" and that the design is for families to "move here monotonically". `Drawing.recognition()` calls it "an experimental, read-only result". Growing fields is the stated plan, not a break in it.
3. **ADR 0017 is `Proposed`, not `Accepted`** — it lists the guards required before it can be. This is phase-1 work inside that epic.
4. **The proposed remedy would make things worse.** Empty-tuple defaults make "no grooves found" indistinguishable from "never computed" — the empty-vs-absent confusion #1022 spent real effort removing on the declared path. Worse, a defaulted field that a future orchestration forgets to set constructs silently as `()`; without defaults it is a `TypeError` on the first call. Fail-closed is the whole point of the manifest this PR shrinks.

If `RecognitionResult` becomes a supported construction surface, defaults and a migration path are the right answer. It is not one today, and adding them now would trade a real invariant for a hypothetical consumer.

### Not disputed

The review found the threading internally consistent: slot patterns derived from the accepted slot records, grooves/flats given the shared cylinder substrate, both `build_part_model` call sites injecting, and the standalone fallbacks intact.

### Caveat

Codex cannot execute anything in this environment (`bwrap` namespace error), so it traces source only. Both findings were reproduced or checked against the code before I acted.

### After the fix

- 36 passed across the recognition suites; four lint gates clean.
- CI was 19/19 green on the previous head and will re-run on `e46b674`.

PR #1031, opened 2026-08-04
feat(#1025): split step shoulders into shared evidence and per-consumer projection

### LINKED ISSUE #1025: Split recognise_step_shoulders into raw riser evidence and pure per-consumer projections

pzfreo:
Closes #1025. Branches off `main` (after #1030 merged).

## The shape of the problem

`recognise_step_shoulders(part, *, levels)` did a full `part.faces()` scan and used `levels` only to *select* which risers qualified. The scan was the cost; the level set was a projection parameter.

That mattered because the two consumers legitimately pass different level sets — model construction filters by plate and pocket ownership, critique must not (ADR 0015 forbids lint taking its inventory from the model). So the family could not be hoisted into the aggregate, and **lint rescanned the solid on every pass**.

#1021 read that as a property of the feature and budgeted the rescan. It is really a property of the function's *shape*.

## The split

- **`recognise_risers(part)`** — the scan and every level-independent gate: planarity, in-plane axis, the full-span test that separates a step from a pad or blind pocket, the interior-position test, the structural-ramp size floor, the area floor. Owned by the aggregate, run once per build.
- **`project_step_shoulders(risers, *, levels)`** — the filter. **No `part` argument by construction**, so it cannot become a second recognition site even by accident.

`RiserEvidence` carries `lo_at_envelope`/`hi_at_envelope` so the level-free half of the oblique tie-test is precomputed; the projection then needs the levels and nothing else about the solid.

## Measured

| | before | after |
|---|---|---|
| automatic build | 16 families, `recognise_step_shoulders` **twice** | 17 families, **once each** |
| lint of a built drawing | 1 scan **per pass** | **0** |
| declared build | 0 | 0 |

Linting now recognises nothing. That is **epic #1018's phase-0 third guard**, and the last of the three.

## Manifest

`DEFERRED` loses its `CALLER_SPECIFIC_INPUT` entry — down to the three `CLASSIFICATION_GATED` families (#1028). The enum member survives with a note that this reason is usually a *shape* problem rather than a fact about the feature: the scan never depended on the caller, only the filter did. Prefer the split before reaching for that member again.

## Folded in: a false-negative door

`lint_prismatic_coverage` took `step_zs=`, and that argument **fully determined** the shoulder answer: `step_zs=[]` yielded no shoulders, so `missing_transitions` was structurally zero and `unrecognised_defining_geometry` could never fire. The engine never passed that — but a completeness check is the one place a clean absence is indistinguishable from a clean part.

The parameter is gone. The guard asserts on the **signature**, not on behaviour: the door is closed by there being nothing to open it with.

## Two things worth reviewer attention

**`declare.step_level` keeps its own scan, deliberately.** It reads the object the author handed it — a different solid from the one the aggregate recognised — in the same sense `hole(cylinder)` reads a diameter off the cylinder you passed. Not an oversight.

**The record-type universe now derives from `project_*` as well as `recognise_*`.** `StepShoulder` is emitted by a projection now; deriving from `recognise_` alone would have silently dropped it from the completeness guard the moment it moved, and left a future `project_*` record type with no home and no test saying so. `RiserEvidence` gets a documented home in `_ORCHESTRATED_RECORDS`.

## Verification

- 68 passed across the recognition, contract, registry and guard suites; 1286 passed across the broad consumer suites on the previous revision.
- Four lint gates clean.
- **Five mutations, each producing a red guard**: reintroducing `step_zs=` → signature guard; making lint rescan → count guards; making the projection ignore its levels → discrimination guard; dropping `RiserEvidence`'s home → partition guard; narrowing the universe back to `recognise_` → partition guard.

## Issue corrections

#1025's body has been corrected in three places where it described work that #1021's reshape never landed: the `CoverageState.step_shoulders` memo, the `critique_step_shoulders` fallback and the ADR 0017 §6 amendment do not exist, and `test_the_critique_takes_no_shoulder_inventory` was cited as an existing guard but had never been written. The split's argument is unaffected — it was never justified by the memo's existence. This PR writes the missing guard for real.

Parent: #1018. Remaining sibling: #1028.

### DIFF
--- a/src/draftwright/analysis.py
+++ b/src/draftwright/analysis.py
@@ -647,6 +647,7 @@ def _analyse(
             # branch `recognition` is non-None by construction (it is the not-declared arm).
             slot_patterns=list(recognition.slot_patterns) if recognition else None,
             grooves=list(recognition.grooves) if recognition else None,
+            risers=list(recognition.risers) if recognition else None,
             flats=list(recognition.flats) if recognition else None,
             pockets=pockets,
             pocket_patterns=pocket_patterns,
--- a/src/draftwright/analysis.py
+++ b/src/draftwright/analysis.py
@@ -583,12 +583,11 @@ def _analyse(
     else:
         recognition = build_recognition_result(part, cylinders=(z_cyls, cross_cyls))
         _turned = TurnedProfile.from_steps(list(recognition.turned_steps))
-        if _turned is not None and _turned.axis == "z":
-            step_zs = [z for z in _turned.shoulders if bb.min.Z + 0.6 < z < bb.max.Z - 0.6]
-        else:
-            # The aggregate's own area-filtered extraction (#578 review; hoisted there by
-            # #1022 so critique on the declared path reads one value rather than rescanning).
-            step_zs = list(recognition.step_levels)
+        # The aggregate's own rule (#578 review; hoisted there by #1022, shared with critique
+        # by #1025). Both callers deriving this separately let lint project over a different
+        # ladder than the model was sized from — which is exactly the divergence one waist is
+        # supposed to prevent.
+        step_zs = recognition.step_ladder(bb)
     # The aggregate owns the shared substrate from here on.  Rebind the local projection so
     # model construction, Analysis and the finished BuildState all consume the same inventory
     # object rather than parallel list/tuple wrappers that merely happen to contain equal data.
--- a/src/draftwright/annotations/orchestrator.py
+++ b/src/draftwright/annotations/orchestrator.py
@@ -217,6 +217,7 @@ def build_model(a: Analysis):
         # #1024 collapses that duplication.
         slot_patterns=a.recognition.slot_patterns,
         grooves=a.recognition.grooves,
+        risers=a.recognition.risers,
         flats=a.recognition.flats,
         pockets=a.recognition.pockets,
         pocket_patterns=a.recognition.pocket_patterns,
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -2713,12 +2713,16 @@ class Drawing:
             bosses: list | None
             pads: list | None
             pockets: list | None
-            step_zs: list | None
+            recognition: object | None
             prof_kw: dict
             if a is not None and a.recognition is not None:
                 cyls = a.cyls
                 holes, patterns, bosses = a.holes, a.patterns, a.bosses
-                pads, pockets, step_zs = a.pads, a.pockets, a.step_zs
+                pads, pockets = a.pads, a.pockets
+                # The whole aggregate, not a level set: coverage projects the shared riser
+                # evidence over recognition's OWN levels, so no caller can narrow the
+                # shoulder inventory (#1025).
+                recognition = a.recognition
                 prof_kw = {"prof": a.prof}
             elif a is not None:
                 # A DECLARED build recognised nothing (#1022), so `a.holes` and friends are
@@ -2732,13 +2736,12 @@ class Drawing:
                 bosses = list(rec.bosses)
                 pads = list(rec.pads)
                 pockets = list(rec.pockets)
-                # NOT `a.step_zs`: that is declaration-sourced on this path, and critique
-                # taking its inventory from the model is what ADR 0015 forbids — it would
-                # make lint blind to exactly the geometry a sparse declaration omitted, which
-                # is the case `unrecognised_defining_geometry` exists to report. The
-                # aggregate's geometry-sourced ladder is the right answer, and reading it here
-                # rather than passing `None` is what stops coverage rescanning it every lint.
-                step_zs = list(rec.step_levels)
+                # The aggregate, NOT anything off `Analysis`: its levels and risers are
+                # geometry-sourced, where the declared `Analysis` carries what the author
+                # declared. Critique taking its inventory from the model is what ADR 0015
+                # forbids — it would make lint blind to exactly the geometry a sparse
+                # declaration omitted, the case `unrecognised_defining_geometry` reports.
+                recognition = rec
                 # `a.prof` IS declaration-sourced, and here that is right rather than a
                 # shortcut: axial critique judges the declared profile's dimensioning, so a
                 # declared turned part keeps it without the aggregate.
@@ -2748,7 +2751,7 @@ class Drawing:
                     self._cyl_cache = analyse_cylinders(self.part)
                 cyls = self._cyl_cache
                 holes = patterns = bosses = None
-                pads = pockets = step_zs = None
+                pads = pockets = recognition = None
                 prof_kw = {}
             issues += lint_feature_coverage(
                 self.part,
@@ -2791,7 +2794,7 @@ class Drawing:
                 pockets=pockets,
                 bbox=a.bb if a is not None else None,
                 features=getattr(model, "features", ()) if model is not None else (),
-                step_zs=step_zs,
+                recognition=recognition,
             )
             # Reverse direction (#487): a DECLARED feature with no matching geometry (a stale
             # phantom callout). Only for a caller-supplied model — detection can't over-declare.
--- a/src/draftwright/linting/coverage.py
+++ b/src/draftwright/linting/coverage.py
@@ -28,17 +28,17 @@ from build123d_drafting.helpers import CenterMark, Dimension, TitleBlock
 from draftwright._core import _DIAM_RE, _END_ON, HoleRef, _axis_letter, _fmt, _xyz
 from draftwright.linting.issues import LintIssue
 from draftwright.recognition import (
+    RecognitionResult,
     TurnedProfile,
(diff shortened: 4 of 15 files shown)


pzfreo:
## Adversarial review (Codex) — triaged, all three findings real

Fixed in `2263749`. Every one was a genuine defect, and the first two were in claims I had made in this PR's own description.

### 1 (High) — the false-negative door was renamed, not closed

Correct, and the sharper half of the finding is about the guard rather than the code. `recognition=` is duck-typed: `SimpleNamespace(risers=(), step_levels=())` forces `source_shoulders == []` exactly as `step_zs=[]` did. Verified by running it.

My guard asserted that a `recognition` parameter *exists*, which establishes nothing about narrowability — a test that could not fail for the reason it claimed, which is worse than no test. The PR description's "not narrowable to a subset the caller chose" was simply false at that boundary.

`lint_prismatic_coverage` now raises `TypeError` for anything that is not recognition's own frozen result, and the guard asserts that rejection behaviourally. The signature assertions stay as the necessary-but-insufficient half.

### 2 (Medium) — the split broke non-default tolerances

Also correct. The old one-stage call used a single `tol` for the geometric gates **and** the level ties. After the split, `recognise_risers(part, tol=0.1)` followed by a bare `project_step_shoulders(...)` mixed 0.1 with the projection's own 0.5 — a riser 0.3 mm off a level was rejected before and accepted after.

`RiserEvidence` now carries the tolerance it was scanned with and the projection inherits it, so the natural two-stage call is equivalent at *any* tolerance rather than only the default. Carried on the record because a caller holding the evidence should not have to remember how it was produced.

### 3 (Medium) — critique silently changed its level source for Z-turned parts

Correct, and I had not noticed. `_analyse` used the turned profile's shoulders for a Z-turned part; my coverage change always used the face levels. Those differ whenever a horizontal face is not an OD shoulder — a blind bore's flat floor being precisely the case the turned branch exists to exclude.

`RecognitionResult.step_ladder` is now the one rule both callers use. Divergent fixture, as requested: a Z-turned shaft with a blind bore, face levels `[15, 30]` versus ladder `[15]`.

### On the equivalence assessment

Noted and appreciated — the gate-reordering analysis (full-span and interior gates level-independent; ramp size and area floors level-independent; oblique endpoint predicate preserved as `level tie OR envelope tie`; rounding still after acceptance and before dedup) matches my own reading, and it is the part of this refactor I was least able to prove by testing alone.

### One process note worth recording

My **first** fix for finding 3 was itself unguarded. I asserted the rendered step-length count, which looked behavioural but was vacuous: on a turned part `step_zs` feeds only the strip reservation, so no rendered rung reads the difference and the mutation passed. I checked that rather than assuming, then replaced it with a call-count guard on the shared rule — structural, because the claim "one rule, two callers" is structural.

That is the third time in this epic that a test which *looked* behavioural could not fail. Running the mutation is the only thing that has caught it each time.

### Verification

- 970 passed across nine affected suites; 25 in the gate suite after the fixes.
- Four gates clean.
- **Four mutations, each red**: removing the isinstance check; making the projection ignore the evidence tolerance; making `step_ladder` skip its turned branch; making `_analyse` bypass it.
- The previous green CI run was on the pre-fix head and is superseded by this push.

pzfreo:
## Second adversarial review — triaged

Both findings real. One fixed, one classified and deferred with an issue rather than dismissed.

### 2 (Low, but the more serious of the two) — the ladder guard only proved *sizing* calls the rule — **fixed**

Correct, and this is the same trap for the third time in this epic. The test counted `RecognitionResult.step_ladder` across `build_drawing(part)` and asserted a nonzero total — but `_analyse` supplies that call by itself, so reverting critique to `levels=_rec.step_levels` left the guard **green** while sizing and critique used different ladders. Exactly the regression the test's name claims to prevent.

Now clears the counter after construction and asserts the critique call separately. Mutation-validated: that revert is red.

Worth stating plainly — this is the third guard in this epic that looked behavioural and could not fail. Running the mutation is the only thing that has caught any of them.

### 1 (Medium) — `isinstance` proves neither provenance nor correspondence — **classified, deferred to #1032**

Also correct, and verified: `lint_prismatic_coverage(stepped, ..., recognition=build_recognition_result(Box(20,20,20)))` is accepted, and the unrelated aggregate's empty risers suppress `unrecognised_defining_geometry`.

The sharp part of the finding is again about the guard, not the code: my test rejected only a `SimpleNamespace`, so it stayed green while the guarantee in its own name was violated using the accepted type. **Renamed** to `test_the_critique_rejects_an_assembled_shoulder_inventory` — what it establishes — with the limitation named inline rather than implied away.

Not fixed here, deliberately. Every injected inventory in the engine is trusted on type alone — `holes=`, `pockets=`, `pads=`, `slots=`, `grooves=`, `flats=`, `risers=` — which is the accepted cost of ADR 0008 Amendment 5. #1025 narrowed the door it opened (a *level set* is no longer routable; a duck-typed stand-in is rejected) but cannot close the general one. Adding provenance checking to this one parameter would be inconsistent with every sibling and *still* not fail-closed.

Filed as **#1032**, with the three real options weighed rather than a vague note: cheap correspondence check, private build-state path (your suggestion, and the strongest for completeness checks specifically), or an identity token. Low priority — no engine path passes a mismatched aggregate, and it is worth doing when ADR 0017 phase 4 touches that boundary anyway.

### Verification

- 50 passed across the affected suites; four gates clean.
- Mutation: reverting critique to the raw face levels → the ladder guard now fails.

Not dismissing either finding to reach a clean review: one is fixed, one is a named limitation with an issue and a rationale you can disagree with.

PR #1033, opened 2026-08-04
feat(#1028): gate the classification-dependent families inside the orchestration

### LINKED ISSUE #1028: Carry the part classification in the recognition orchestration so the gated families can migrate

pzfreo:
Closes #1028. The last child of epic #1018's phase 1.

## The deferral, and why it was not simply wrong

`recognise_chamfers`, `recognise_fillets` and `recognise_plates` were `DEFERRED` because `build_part_model` runs each for only one part class — chamfers and fillets on `rotational is None`, plates additionally on `prof is None`. Hoisting them unconditionally would have scanned **every automatic turned build** for a result that branch discards.

#1022 did not unblock this. It removes recognition from the *declared* path; this constraint binds on the *automatic* one. (#1021 briefly pointed these at `blocker=1022`; review corrected them here, precisely so closing #1022 would not claim to unblock a migration it does not.)

## The resolution

Not to keep them out of the aggregate, but to let the aggregate decide. **Owning a family and always running it are different things** — that is the distinction this issue turned on.

`build_recognition_result` now takes the caller's geometric classification, carries it on the result, and gates the three *inside* the one orchestration. One place decides once, rather than each consumer deciding again — which is exactly the scatter ADR 0017 exists to remove, and the reason a per-call-site `if` would have been the wrong shape even though it produces the same calls.

**`rotational` stays an argument rather than something recognition derives.** The rotational test reads bbox proportions and OD offset: that is sizing's question *about* the part, not recognition's inventory *of* it. Deriving it below the IR would move drafting classification into `recognition/`, an explicit ADR 0017 non-goal. It defaults to prismatic, so a caller with no classification — the lazy critique aggregate on a declared build — gets the **complete** inventory. Over-recognising costs time; under-recognising reports a real feature as absent, and this is a completeness check.

The result carries `rotational` so consumers can tell a gated-away inventory from an empty one — the same empty-vs-absent distinction #1022 drew for the declared path.

## `DEFERRED` is empty

Every public `recognise_*` family is now owned by the one orchestration. That is what ADR 0017 phase 1 was aiming at.

The mechanism stays fail-closed — a new family must still be classified, and I verified that by mutation with the map empty — and every `Deferral` member survives for a future family. What went was each *deferral*, as its stated constraint stopped being true:

| reason | retired by | because |
|---|---|---|
| `BUILD_MODEL_ONLY` | #1026 | #1022 removed the cost |
| `CALLER_SPECIFIC_INPUT` | #1025 | it was a *shape* problem — the scan never depended on the caller, only the filter did |
| `CLASSIFICATION_GATED` | this PR | the gate belongs in the orchestration, not outside it |

## Measured, per family

| path | families run | the gated three |
|---|---|---|
| prismatic | 17, once each | 1 each |
| turned | 14 | **0** |
| declared | 0 | 0 |

## Guards

`test_a_classification_gated_reason_is_true_of_the_build` becomes `test_a_classification_gated_family_does_not_run_for_the_excluded_class` — it checks the aggregate rather than `build_part_model`, per the issue's acceptance. It keeps both turned fixtures: the plain cylinder separates the plates conjunction (*not rotational* **and** *no profile*) that the stepped shaft satisfies twice over, so weakening the gate to either half alone still fails.

A one-directional gate test is satisfied by never running them at all, so `test_a_gated_family_still_runs_for_the_class_that_consumes_it` is the counterexample.

`test_the_gate_is_the_orchestrations_not_the_call_sites` runs **one solid under both classifications**, so any difference in the three inventories is the gate's doing and not the geometry's — and asserts the gate is narrow: holes, risers and step levels are untouched.

## Verification

- 964 passed across nine suites; four lint gates clean.
- **Mutations, each red**: deleting the gate → the exclusion guards; gating the consuming class away too → four guards including the oracle; unclassifying a family with `DEFERRED` empty → the manifest guard.

Parent: #1018. With this, phase 1 is complete and ADR 0017 can be considered for acceptance.


pzfreo:

Two findings, both real, both fixed in the push above. The review found **no source-level regression** and independently confirmed the gate equivalence, the `rotational=False` default's safety, the empty-vs-`None` injection semantics, and that the fail-closed mechanism survives an empty `DEFERRED`.

### 1 (Medium) — the plates conjunction guard did not guard the `prof is None` half — **fixed**

Correct, and I verified it by running the mutation: replacing the production condition with `if prismatic` alone leaves **17 passed**. Both turned fixtures in the manifest's exclusion test go through `build_drawing`, where they classify rotational — so the rotational half alone was sufficient to skip plates on each, and the profile half was never exercised.

**The claim in this PR's description — that weakening either half makes the guard fail — was unsupported.** That is the fourth guard in this epic that asserted more than it established, and again only the mutation caught it.

The untested half is the documented "caller has no classification" path: a stepped shaft with `rotational=False`, where only the profile keeps the scan away. `test_the_plates_gate_needs_both_halves_not_just_the_rotational_one` now covers it.

Your point about the oracle being insufficient is the sharper half and I took it: the new test **counts the call** rather than checking the inventory came back empty. `recognise_plates` naturally finds nothing on a shaft, so an empty result proves the skip only by coincidence — and the cost the gate exists to avoid is the scan itself, not the result.

### 2 (Low) — the injection-equivalence test could not see the three migrated families — **fixed**

Also correct. Its fixtures were a grooved/flatted shaft and a slot lattice; neither produces a chamfer, fillet or plate. So the exact defect you name — an injected non-empty inventory consumed as empty, where `None` rescans but `()` must suppress — would have passed a guard whose entire purpose is that equivalence.

Added a chamfered+filleted block and an L-bracket producing two plates on different axes. Mutation-validated: making `build_part_model` consume injected chamfers as `()` now fails on the new fixture.

### Verification

- 52 passed across the recognition suites; four gates clean.
- Both of your scenarios run as mutations, both now red.

Issue #1034, opened 2026-08-05
Second A/F callout on parallel stock has no corner anchor and is dropped


pzfreo:
#1013 gave `Flat` its stock identity, so a part with two parallel same-sized lobes now plans **two** A/F definitions instead of collapsing them into one. The renderer places one and reports the other as `flat_dropped` — "no clear room".

That is already a large improvement on the previous behaviour, where the second definition simply vanished and nothing said so. But the drawing is still incomplete, and it is now incomplete *loudly* rather than correctly.

## What happens

```python
lobe = Cylinder(15, 40) - Pos(12.5, 0, 0) * Box(10, 40, 50)
build_drawing(lobe + Pos(gap, 0, 0) * lobe)
# -> one m_flat_z* callout, one flat_dropped issue, for every gap tried (60/100/160)
```

The separation makes no difference, so this is not "the part is too crowded". Both jobs come from `_corner_candidates`, which offers view-corner anchors; the two groups compete for the same corners and `_leader_callout_pass` places whichever wins with `geom_clear=True`.

## Why it is not part of #1013

#1013's scope was the record and the grouping — "the fix is the record" (ADR 0013), plus `render_flats` grouping by it. That is done and guarded. Finding a second anchor is a placement-capacity question about corner-anchored leader callouts, which is ADR 0014 territory and would want the corridor solve rather than another bespoke fallback.

Doing it inside #1013 would also have meant changing placement while changing recognition, which makes any regression ambiguous between the two.

## Worth checking first

Whether the two groups genuinely have no second viable corner, or whether `_corner_candidates` is offering anchors that `geom_clear` rejects for a reason that does not apply here. A part 130 mm wide with two lobes 100 mm apart should have room for two leaders on opposite sides; if it does not, the candidate generation is the thing to look at, not the solver.

## Priority

Low-ish. Two-lobe parts are uncommon, and the failure is now reported rather than silent — a user gets a lint error naming the missing callout instead of a clean-looking wrong drawing. Worth doing when the flat passes are next touched.

Guarded by `tests/test_flat_stock_identity.py::test_two_parallel_lobes_are_two_definitions_and_a_lost_one_is_reported`, which asserts *placed + reported == 2* — so tightening this to two placed callouts will not need the test rewritten, only the assertion sharpened.

Related: #1013, #1011 (the coverage half). Parent epic: #1018 phase 2.

pzfreo:
## Reproduction and corrected root cause — current `main` (`130f236`)

[...]

PR #1035, opened 2026-08-05
fix(#1013): a flat carries its stock identity, so parallel lobes stop collapsing

### LINKED ISSUE #1013: render_flats collapses same-sized flats on parallel stock into one callout

pzfreo:
Closes #1013. First slice of epic #1018 phase 2 — geometry-derived stock identity.

## The defect

`Flat` recorded `axis` (the letter), `across` and `at`, and nothing that said **which piece of round stock** a flat belonged to. From that record these are indistinguishable:

- two faces at `(-12.5, 0, 0)` and `(12.5, 0, 0)` — one double-D, **one** A/F definition;
- two faces at `(0, 0, 0)` and `(100, 0, 0)` — two parallel lobes, **two** definitions.

`render_flats` grouped by `(axis, across)`, so a part with two same-sized lobes got **one** callout and the second lobe went undefined on the sheet — with nothing reporting it.

## The fix is the record

ADR 0013 is explicit about this case, and `FaceLevel` and `TurnedStep` have had the same treatment. `Flat` and `FlatFeature` now carry `axis_line` — the two coordinates perpendicular to the axis — and the renderer groups by it. The recogniser already had the owning cylinder in hand at the match point, so it costs nothing to carry.

## A bug the compiled-plan boundary caught, and my first fix hid

Worth surfacing, because it is exactly the failure mode ADR 0016 Amendment 1's fact allowlist exists to prevent.

My first cut read the fact as `getattr(g.facts, "axis_line", (0.0, 0.0))`. The compiled plan **refuses** facts renderers are not allowed to read — but the `getattr` default swallowed that refusal, both groups fell back to the same value, and the behaviour did not change at all. A defensive default defeated a deliberate guard.

`axis_line` is now declared as **structure** in `_FACTS["flat"]` (it says which stock a flat belongs to; the drawing never prints it), and the access is direct so an unlisted fact fails loudly.

## Behaviour

| case | before | after |
|---|---|---|
| single D-flat | 1 callout | unchanged, lint clean |
| double-D | 1 callout | unchanged, lint clean |
| two parallel lobes | **1 callout, silently incomplete** | 1 placed + 1 **reported** (`flat_dropped`) |

The last row is the improvement. Finding a second anchor is a corridor-capacity question — filed as **#1034**, deliberately not done here, because changing placement alongside recognition makes any regression ambiguous between the two. A completeness failure that is *visible* beats a clean-looking drawing that is wrong.

The guard asserts `placed + reported == 2`, so tightening #1034 to two placed callouts sharpens an assertion rather than rewriting the test.

## The declared path

`declare.flat` takes an optional `axis_line=`, defaulting to the origin line. It **cannot** be derived from `at`: a double-D's two faces have different face centres but one axis line, so deriving it would split every declared double-D in two. The default reproduces pre-#1013 grouping, so single-stock declarations are unaffected; a declarer with parallel lobes states them, because nothing in a flat's own geometry says it.

## Scope note

The coverage half of #1013 is **not** here. `lint_flat_coverage` lives on #1011, which is still an open PR, so it does not exist on `main`. It should group by `axis_line` when that lands.

## Verification

- 1357 passed across ten suites; four gates clean.
- **Mutations, each red**: reverting the renderer grouping to `(axis, across)`; dropping `axis_line` from the compiled-plan fact allowlist.

Parent: #1018 (phase 2). Related: #1011, #1034.

### DIFF
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -1710,9 +1710,16 @@ def render_flats(dwg, plan, a, *, ctx, only=None) -> int:
         )
         if pd is None:
             continue
-        collapse.setdefault((g.facts.axis, round(pd.value, 3)), []).append((g, pd))
+        # Grouped by the stock's AXIS LINE as well as the size (#1013). Two same-sized flats
+        # on one axis line are the two faces of one double-D — one A/F definition, one
+        # callout. On DIFFERENT axis lines they are separate lobes, each needing its own: the
+        # axis letter alone cannot tell those apart, so a part with two parallel 25 A/F lobes
+        # used to get one callout and leave the second undefined on the sheet.
+        collapse.setdefault((g.facts.axis, g.facts.axis_line, round(pd.value, 3)), []).append(
+            (g, pd)
+        )
     jobs = []
-    for gi, ((axis, _across), members) in enumerate(sorted(collapse.items())):
+    for gi, ((axis, _axis_line, _across), members) in enumerate(sorted(collapse.items())):
         if only is not None:
             # #426 Ph2b subset (finalize): filter members AFTER enumerating the collapse so gi
             # stays the full-drawing group index (Codex #811) — see render_fillets.
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -1710,16 +1710,21 @@ def render_flats(dwg, plan, a, *, ctx, only=None) -> int:
         )
         if pd is None:
             continue
-        # Grouped by the stock's AXIS LINE as well as the size (#1013). Two same-sized flats
-        # on one axis line are the two faces of one double-D — one A/F definition, one
-        # callout. On DIFFERENT axis lines they are separate lobes, each needing its own: the
-        # axis letter alone cannot tell those apart, so a part with two parallel 25 A/F lobes
-        # used to get one callout and leave the second undefined on the sheet.
-        collapse.setdefault((g.facts.axis, g.facts.axis_line, round(pd.value, 3)), []).append(
-            (g, pd)
-        )
+        # Grouped by the STOCK IDENTITY as well as the size (#1013). Two same-sized flats on
+        # one piece of stock are the two faces of one double-D — one A/F definition, one
+        # callout. On different stock they are independent, each needing its own; the axis
+        # letter alone cannot tell those apart, so a part with two parallel 25 A/F lobes used
+        # to get one callout and leave the second undefined on the sheet.
+        #
+        # Identity is the axis line AND the axial span, because neither alone is enough:
+        # parallel lobes share a span but not a line, and coaxial stacked stock shares a line
+        # but not a span. Grouping on the line alone merged the coaxial case silently — the
+        # same defect this fix is about, one arrangement over (Codex #1035 r1).
+        collapse.setdefault(
+            (g.facts.axis, g.facts.axis_line, g.facts.stock_span, round(pd.value, 3)), []
+        ).append((g, pd))
     jobs = []
-    for gi, ((axis, _axis_line, _across), members) in enumerate(sorted(collapse.items())):
+    for gi, ((axis, _line, _span, _across), members) in enumerate(sorted(collapse.items())):
         if only is not None:
             # #426 Ph2b subset (finalize): filter members AFTER enumerating the collapse so gi
             # stays the full-drawing group index (Codex #811) — see render_fillets.
--- a/src/draftwright/model/compiled.py
+++ b/src/draftwright/model/compiled.py
@@ -252,7 +252,10 @@ _FACTS: dict[str, tuple[str, ...]] = {
     # changes output, and a migration that claims byte-identity is the wrong place for it.
     "chamfer": ("frame", "axis", "leg2", "angle"),
     "fillet": ("frame", "axis"),
-    "flat": ("frame", "axis"),
+    # `axis_line` is STRUCTURE, not a measurement: it says which piece of stock a flat
+    # belongs to, so the renderer can tell a double-D's two faces from two parallel
+    # lobes (#1013). The drawing never prints it.
+    "flat": ("frame", "axis", "axis_line"),
     "groove": ("frame", "axis"),
     "plate": ("frame", "axis"),
     # Raw AP242 PMI is the documented non-generated exception: its source-authored label
--- a/src/draftwright/model/compiled.py
+++ b/src/draftwright/model/compiled.py
@@ -255,7 +255,7 @@ _FACTS: dict[str, tuple[str, ...]] = {
     # `axis_line` is STRUCTURE, not a measurement: it says which piece of stock a flat
     # belongs to, so the renderer can tell a double-D's two faces from two parallel
     # lobes (#1013). The drawing never prints it.
-    "flat": ("frame", "axis", "axis_line"),
+    "flat": ("frame", "axis", "axis_line", "stock_span"),
     "groove": ("frame", "axis"),
     "plate": ("frame", "axis"),
     # Raw AP242 PMI is the documented non-generated exception: its source-authored label
--- a/src/draftwright/model/declare.py
+++ b/src/draftwright/model/declare.py
@@ -497,7 +497,9 @@ def _read_flat_face(face) -> Point:
     return (round(c.X, 4), round(c.Y, 4), round(c.Z, 4))
 
 
-def flat(obj=None, *, axis=None, across=None, at=None, axis_line=None) -> FlatFeature:
+def flat(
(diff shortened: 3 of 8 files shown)


pzfreo:

Two High findings, both real and both fixed in `d620e07`. A third defect surfaced from the same area afterwards, found by the #964 emit-fidelity oracle rather than the review.

### 1 (High) — `axis_line` is not stock identity — **fixed**

The strongest finding of this review, and correct. Two D-shafts stacked **coaxially** with a gap share an axis line, so grouping on it alone merged two independent A/F definitions into one callout with **no report**. Verified before fixing: the coaxial fixture gave one callout and zero drops — the exact defect #1013 exists to fix, one arrangement over.

What makes it sting is that **the code already knew**. #1015 taught the *opposition* test that "the same infinite axis is not the same piece of stock", and the recogniser computes an axial span for precisely that reason. I read that comment while implementing and still grouped on the line alone.

`Flat`/`FlatFeature` now carry `stock_span`, and together the two are the identity — neither suffices: parallel lobes share a span but not a line; coaxial stock shares a line but not a span.

I did **not** propagate the recogniser's internal `stock` tuple wholesale: it includes a solid index, which its own comment marks as a same-run equality check. Putting that on a record would be a false identity, unstable across runs. The propagated key is purely geometric.

### 2 (High) — emitted scripts discarded the identity — **fixed**

Correct, and I had flagged this as the likely gap when commissioning the review. Because the declared defaults reproduce pre-#1013 grouping, an emitted script regenerated exactly the drawing detection had just fixed. Silence was not neutral; it was the old bug. `sheet.flat(...)` now emits both fields always, verified end-to-end through `generate_sheet_script`.

### 3 — found afterwards by the #964 oracle, not by this review or my guards

Worth reporting rather than folding in quietly. `declare.flat` **accepted `stock_span=` and silently ignored it** — the parameter landed but the return statement had been reformatted before my edit, so only `axis_line` was wired.

My new guard asserted the emitted script *text* contained `stock_span=`. It did. The round-tripped *feature* still lost it. The fidelity oracle compares detected-versus-redeclared field values, which is the stronger claim.

That is the same "a guard measures a proxy for its claim" pattern this epic keeps producing — this time in a test I had just written to fix that very class of bug. Both halves now share one `_stock_pair` helper so they cannot drift.

### 3 (Medium, review's own numbering) — the declared `(0,0)` default — **classified, not changed**

The concern is fair: omission looks valid and produces an incomplete drawing without diagnostics. But the proposed remedy — diagnosing ambiguous multi-flat declarations — would fire on the **commonest correct usage**. A declared double-D is exactly "two flats, same axis, same size, default identity", and that is right rather than ambiguous. A warning there trains users to ignore warnings.

The realistic path to the problem was the emitter, and finding 2 closes it: generated scripts now state the identity explicitly, so only a hand-written multi-stock declaration can hit the default, and the docstring says so. Deriving identity from owning stock geometry in the object-backed form is the better long-term answer and belongs with #1034's placement work, when the flat passes are next opened.

Happy to be overruled — it is a judgement about API ergonomics, not a fact.

### Verification

- 1334 passed across nine suites; four gates clean.
- **Mutations, each red**: grouping on the axis line alone → the coaxial guard; stripping the emitted fields → the emit guard; ignoring `stock_span` in declare → the fidelity oracle.
- The earlier green CI run was on the pre-fix head and is superseded.

pzfreo:
## Second adversarial review — triaged

One High finding. **Correct on the encoding**, and I have not fixed it — with reasons and evidence rather than a dismissal.

### The finding is right

`_axis_line` identifies the stock's axis line by dropping the *dominant* axis letter's coordinate, which identifies a line only for X/Y/Z-aligned axes. `analyse_cylinders` classifies a slanted cylinder by its dominant component — I verified a `(0.707, 0, 0.707)` axis reads as `"x"` — so for such an axis the encoding is:

- **not canonical**: the value depends on which point along the axis the cylinder record reports;
- **not sufficient**: two different slanted directions through one point encode identically.

Both true by inspection, and the suggested canonical form (`_same_axis_line`'s perpendicular projection plus direction, with the span) is the right one.

### Why it is not fixed here — measured, not asserted

**A slanted flat does not render at all.**

```python
slant = Rot(0, 45, 0) * (Cylinder(10, 40) - Pos(8, 0, 0) * Box(10, 40, 60))
build_drawing(slant)
# -> no m_flat_* callout; lint: ['view_annotation_inside_extents', 'flat_dropped']
```

It is recognised and then dropped at placement. So both failure modes are unreachable: the callout never exists to be split or collapsed. A canonical key would fix a component of an already-broken path, with no observable change to any drawing and **no test able to demonstrate the improvement** — which by this epic's own standard means it should not ship as an unverifiable change.

I also checked the false-split half against real geometry rather than accepting it: a slanted double-D's two faces match **one** cylinder record and share a key, so that half does not reproduce today. The false-collapse half is constructible but needs two slanted stocks agreeing on dominant letter, perpendicular coordinates, span *and* size.

### What I did fix — the claim

An overstated docstring is how the next reader repeats the mistake. `_axis_line` now states that it identifies axis-aligned stock only, why the slanted encoding fails on both counts, and what the canonical form would be.

### And a guard so the two cannot drift apart

`test_slanted_stock_is_an_acknowledged_limitation_not_a_silent_wrong_answer` pins the current behaviour and **fails if slanted flats start rendering** while the identity is still the aligned-only key. So a future placement fix cannot quietly ship slanted A/F callouts grouped by a key that cannot tell two shafts apart.

Both halves are filed together as **#1036**, because fixing either alone is untestable — with the identity fix verifiable by an actual slanted callout rather than by unit-testing an encoding.

### Your "otherwise" section

Agreed and confirmed locally: for axis-aligned, non-coincident stock, `(axis_line, stock_span)` handles both the parallel and coaxially-separated cases; the fields are allowlisted, converted, accepted by `declare.flat`, always emitted, and the structural fidelity oracle compares both — it is what caught `declare.flat` silently ignoring `stock_span`.

### Verification

- 7 passed in the identity suite; 1334 across nine suites on the previous revision; four gates clean.
- Happy to be overruled: if you would rather have the canonical key now, it is a contained change and I will do it — my argument is about verifiability, not effort.

Issue #1037, opened 2026-08-05
Part classification is recognition, not sizing — derive it in the aggregate


pzfreo:
`build_recognition_result(part, *, rotational=False)` takes the part classification as an **argument** rather than deriving it. That was #1028's choice, and its stated justification is wrong.

## The bad justification, in the code today

`recognition/result.py`:

> *rotational* is the caller's geometric classification (`analysis._classify_geometry`). It stays an argument rather than something recognised here: the rotational test reads bbox proportions and OD offset, which is sizing's question about the part, not recognition's inventory of it — deriving it below the IR would move drafting classification into `recognition/`, an explicit ADR 3 (was 0017) non-goal.

That does not survive contact with the code. Every input and every test in `_is_rotational` is geometric:

```python
def _is_rotational(x_size, y_size, od_diam, od_axis_offset):
    envelope = max(x_size, y_size)
    return (abs(x_size - y_size) <= _SQUARENESS_TOL * envelope
            and od_diam >= _OD_FILL_MIN * envelope
            and od_axis_offset <= _OD_AXIS_TOL * envelope)
```

Bbox proportions, largest external cylinder diameter, that cylinder's concentricity. "Is this turned stock?" is a fact recovered from the solid, the same kind of fact as "this face is a machined flat". Drafting *consumes* it (view selection, centrelines) exactly as it consumes hole diameters — which nobody calls drafting policy.

`recognition/` already uses `part.bounding_box()` freely (`recognise_risers`, `step_level_zs`, the flats chord tests), and `od_diam` comes from `analyse_cylinders`/`full_cylinders`, which are recognition substrate. So the stated obstacle does not exist.

The real reason was scoping: `_classify_geometry` lives in `analysis.py` and moving it was larger than #1028 warranted. That is a fine decision and a bad justification, and the bad one is now written where the next reader will find it.

## What to move

`analysis._classify_geometry` returns `_GeomClass(z_cyls, cross_cyls, z_diams, cross_diams, od_diam, od_axis, is_rotational)`. Every field is recognition substrate or a classification derived from it.

Verified before filing:

- **Single call site** — `_analyse` only. Nothing else imports it.
- **No layout coupling**, including the #222 horizontal-round-body fallback: it reads cylinders, bbox sizes and centres, and reuses the same predicate. Nothing about views, pages or scale.
[...]

Issue #1036, opened 2026-08-05
Slanted flats do not render, and their stock identity key is not canonical


pzfreo:
Two gaps in the same feature area, worth doing together because fixing either alone is untestable.

## 1. Slanted flats do not render

```python
slant = Rot(0, 45, 0) * (Cylinder(10, 40) - Pos(8, 0, 0) * Box(10, 40, 60))
build_drawing(slant)
# -> no m_flat_* callout; lint: ['view_annotation_inside_extents', 'flat_dropped']
```

The flat is *recognised* (`Flat(axis="x", across=13.0, ...)`) and then dropped at placement. So a D-shaft that is not axis-aligned gets no A/F callout, reported but not drawn.

## 2. `axis_line` is not a canonical line key for slanted axes

`recognition/flats.py::_axis_line` identifies the stock's axis line by dropping the coordinate of the *dominant* axis letter. That identifies a line only when the axis is along X, Y or Z. `analyse_cylinders` classifies a slanted cylinder by its dominant component — a `(0.707, 0, 0.707)` axis reads as `"x"` — and for such an axis the encoding is:

- **not canonical** — the value depends on which point along the axis the cylinder record happens to report, so two patches of one shaft could key differently (a false split);
- **not sufficient** — two different slanted directions through one point encode identically, so two shafts could key the same (a false collapse).

The canonical form is the perpendicular foot from the origin plus the direction, which `_same_axis_line` already computes for the opposition test; the identity should reuse that plus the axial span.

## Why they belong together

Raised by an adversarial review of #1035 (r2), which was right about the encoding. It was **not** fixed there because the two failure modes are unreachable while gap 1 stands: the callout is dropped before identity matters, so a canonical key would fix a component of an already-broken path with no observable change to any drawing, and no test able to demonstrate the improvement.

Doing them together means the identity fix can be verified by an actual slanted A/F callout rather than by unit-testing an encoding.

## Guard already in place

`tests/test_flat_stock_identity.py::test_slanted_stock_is_an_acknowledged_limitation_not_a_silent_wrong_answer` pins the current state and **fails if slanted flats start rendering** while the identity is still the aligned-only key. So the two cannot be fixed months apart with a wrong callout in between.

`_axis_line`'s docstring states the limitation and its reasoning rather than implying the general claim.

Related: #1013, #1034 (the parallel-lobe placement half). Parent epic: #1018 phase 2.

pzfreo:
[...]

Issue #1032, opened 2026-08-04
lint_prismatic_coverage takes an inventory it cannot vouch for, diverging from ADR 3 (was 0017 §6)


pzfreo:
`lint_prismatic_coverage(part, ..., recognition=...)` type-checks that it received a real `RecognitionResult`, but that establishes neither **provenance** (this aggregate came from this build) nor **correspondence** (this aggregate is of *this* solid). So this suppresses a completeness check without tripping anything:

```python
lint_prismatic_coverage(stepped_part, [], features=(), recognition=build_recognition_result(Box(20, 20, 20)))
```

Verified: the unrelated aggregate is accepted, and its empty `risers` make `source_shoulders` empty, so `unrecognised_defining_geometry` cannot fire.

## Corrected framing (2026-08-05)

This issue originally said the ADR had no answer for how a consumer knows a result corresponds to its part, and weighed three remedies (correspondence check, private path, identity token). **That was wrong — ADR 3 (was 0017 §6) already specifies the ownership path:**

> `Drawing.lint()` may lazily obtain and cache a recognition result when the documented lint policy calls for geometry reconciliation. **The cache belongs in the existing typed `BuildState`** (ADR 1 (was 0005) / #639), **whose builder fill site remains the single writer**; it must not become another ad-hoc private attribute on `Drawing`.

That *is* the correspondence answer. Under §6, lint obtains recognition through the drawing's own `BuildState`, single-writer. Provenance is structural: the state belongs to the drawing, which belongs to the part. There is no channel for a foreign result to arrive, so no check is needed.

So this is not a gap in the model. **It is an injection channel the model did not call for**, added by #1025 when `step_zs=` was replaced with `recognition=`. `Drawing.lint()` does thread from `BuildState` correctly; the parameter simply also permits anything else.

That makes the fix smaller and better specified than the three options previously weighed: **remove the channel**, rather than invent provenance machinery to police it.

## Scope

- `lint_prismatic_coverage` should not take an inventory it cannot vouch for. Options, in preference order: derive from `part` when no trusted state is available, or take the whole `BuildState` (which has a single writer) rather than a bare aggregate.
- Keep the current `isinstance` guard until then — it rejects the accidental duck-typed form, which is the realistic mistake.
- The behavioural guard is `tests/test_declared_recognition_gate.py::test_the_critique_rejects_an_assembled_shoulder_inventory`, which is named for what it establishes and documents this limitation inline.

## What remains genuinely unaddressed

[...]

PR #1039, opened 2026-08-05
docs: record epic #1018 / ADR 0017 status and next steps

pzfreo:
A working status view for the ADR 0017 epic, in `docs/plans/` alongside the other roadmaps. Docs only.

## Why

Phase 1 is complete and the reasoning behind it is currently spread across eight PR descriptions and a dozen issue comments. This puts the durable parts in one place.

## What it records

**State** — 17 families migrated, `DEFERRED` empty, all three phase-0 guards green, with the per-family call counts that demonstrate it (prismatic 17 once each, turned 14, declared 0, repeated lint 0).

**How each deferral ended** — each went when its stated constraint stopped being true, not when it came up on a list. The `CALLER_SPECIFIC_INPUT` case is the one worth re-reading: it was a *shape* problem, and the note says to prefer that split before reaching for the reason again.

**Four things implementation taught the ADR**, as candidates for a single amendment:

1. §2 may be over-specified — #1013 solved a listed reconciliation conflict by fixing the record instead, more simply.
2. §3 named the axis-**direction** component #1013 omitted, so #1036 is a knowing divergence rather than a discovery.
3. §3's "not a flat-only correction" is true of all five thin records, but only `Flat` had a live defect — because only `Flat` grouped without a count. Verified per record; table included.
4. §6 already answers the correspondence question #1032 was originally filed as a gap in.

**Next steps in order**, with #1037's contained-ness already verified, and the scoping decision phase 2's typed identities needs before starting.

**The recurring test lesson** — seven guards this epic asserted more than they established, every one caught by mutation, CI or an existing oracle rather than by a green suite, each measuring a proxy for its claim.

## Tracker, brought up to date alongside this

- **#1019 closed** — all seven of its guards ticked with the PRs that delivered them.
- **#1014 updated** — both objections that made it an issue are dissolved (flats are migrated so there is no double scan; #1022 gated the declared path), with measurements. Left open pending a decision to close or fold into #1011.
- **#1032 rewritten and retitled** — it diverges from §6 rather than exposing a gap, which makes the fix smaller: remove the injection channel.
- **#1036 updated** — recorded as a §3 divergence, with the verified per-record table.
- **#1037 filed** — derive the classification in the aggregate.

### DIFF
--- a/docs/plans/1018-adr0017-status.md
+++ b/docs/plans/1018-adr0017-status.md
@@ -1,0 +1,155 @@
+# Epic #1018 / ADR 0017 — status and next steps
+
+Working state as of **2026-08-05**. ADR 0017 is `Proposed`; this records what has landed
+against it, what implementation taught, and what to pick up next.
+
+Read `docs/adr/0017-recognition-inventory-correspondence-and-measurement-provenance.md`
+first — this file is the progress view, not the decision.
+
+## Where it stands
+
+**Phase 1 is complete.** Every public `recognise_*` family is owned by one orchestration.
+
+| | |
+|---|---|
+| `MIGRATED` families | 17 |
+| `DEFERRED` | **empty** |
+| phase-0 guards | 3 of 3 green |
+
+Measured per family, by counting calls at each recogniser's code object:
+
+| path | families run |
+|---|---|
+| automatic prismatic build | 17, once each |
+| automatic turned build | 14 (the three classification-gated excluded **by design**) |
+| declared build | **0** |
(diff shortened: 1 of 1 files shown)


pzfreo:
Closing unmerged — the status doc is a working handover for the maintainer rather than something the repo needs to carry. The durable parts already live where they belong: the tracker (#1019 closed with its guards ticked, #1014/#1032/#1036 updated, #1037 filed), and the four ADR observations will go into an ADR 0017 amendment rather than a plans file.

PR #1045, opened 2026-08-05
feat(#1043): soft-deprecate Sheet.auto_dimensions() and add_dimension()

### LINKED ISSUE #1043: Soft-deprecate Sheet.auto_dimensions() — authored dimensions are the going-forward surface
The `Sheet` API's going-forward path is **authored** dimensions: `authored_dimensions()` plus explicit `dimension(feature, role)` lines. `auto_dimensions()` should carry a deprecation marker so users — and LLMs editing generated scripts — are steered away from it.

## Why authored is the right default for this surface

- **It is already what the tool emits.** `--script` writes `authored_dimensions()` with an explicit line per dimension. So the hand-written auto path diverges from everything draftwright itself produces.
- **It is the phase-two surface.** "Generate a script, then edit it" only works if the dimensions are addressable text. An automatic set is a black box you cannot edit line by line.
- **Omission means suppression** (ADR 0016 #876). That contract only holds for an authored set. Under `auto_dimensions()` the author cannot express "not that one".
- **For an LLM, an authored list is a lookup; an automatic set is a guess.** "Drop the depth dimension" is a line to delete rather than an intent to infer.

## Scope — what this does NOT deprecate

**`build_drawing(part)`'s automatic path is untouched.** Point the CLI at a STEP or a build123d object and get a fully dimensioned drawing: that is the product's front door and stays exactly as it is. This is about the `Sheet` declaration surface only, where the author has already chosen to be explicit about features and should be equally explicit about dimensions.

## Three dependents that need deciding, not assuming

1. **`add_dimension(feature, role)`** exists solely to augment the planner's set — `build()` raises if it is used without `auto_dimensions()`. Deprecating auto orphans it. **Recommend deprecating both together**, since "augment the set I am discouraging you from using" is not a coherent thing to keep.

2. **`Sheet.from_part(part)` sets the auto source *implicitly***, so deprecating the explicit verb does not cover it. **Recommend keeping `from_part` as-is.** It is the on-ramp — detect the features, get a drawing, then author over it — and its docstring already says adding `dimension(...)` lines *overrides* the implicit choice rather than conflicting. That is precisely the intended migration, and forcing callers to redeclare every feature by hand to reach authored would be a bad trade (as #921 review already concluded).

3. **The "no dimension source" error message** currently leads with `auto_dimensions()` as the first suggestion. It should lead with `authored_dimensions()` and mention auto second, if at all.

## Soft deprecation, not deprecation

**Decided: this is discouraged, not scheduled for removal.** `auto_dimensions()` keeps working
indefinitely.

That rules out a `DeprecationWarning`, because `docs/deprecations.md` carries ADR 0005 §4's rule
— *"a compat surface carries a tracking issue **and** a removal target… a facade with no exit
date is a failure mode"* — and `tests/test_deprecation_dates.py::test_every_deprecation_names_its_removal`
enforces it, requiring removal *language plus a version* (a bare version is rejected by negative
fixtures). Writing "Removed in 0.5.0" without meaning it would be the exact failure that rule
names, wearing a date.

So: **soft deprecation**, the term Python itself uses for "discouraged, still supported, no
removal planned".

- A distinct `SoftDeprecationWarning(UserWarning)` — *not* `DeprecationWarning`, so it neither
  trips the removal-date test nor claims something untrue.
- Docstrings lead with *"Soft deprecated — supported and not scheduled for removal; prefer
  `authored_dimensions()` for new code"*, and say why.
- A **separate table** in `docs/deprecations.md` — *Discouraged: supported, no removal planned* —
  so the exit-date rule stays intact for real deprecations.

Two reasons this is the better instrument for the goal, not merely the more honest one:

1. **`DeprecationWarning` is filtered by default** in notebooks, several test runners, and
   library-internal call paths. A `UserWarning` subclass shows unconditionally. For dissuasion,
   the softer label is the louder warning.
2. **The docstring does most of the work for LLMs**, which read source far more reliably than
   they observe runtime warnings. "Soft deprecated, prefer X" at the top of the docstring is the
   highest-leverage part of this change.

## Work

- `SoftDeprecationWarning(UserWarning)` on the public surface, exported so users can filter it.
- Warn from `auto_dimensions()` and `add_dimension()`, each naming `authored_dimensions()` and
  stating plainly that it is **not** going away.
- Docstrings updated to lead with the recommendation.
- New "Discouraged" table in `docs/deprecations.md`, with a note that these rows deliberately
  have no removal target and why that does not violate ADR 0005 §4.
- Reword the no-dimension-source error to lead with `authored_dimensions()`.
- Update README / docs examples to the authored form.
- **ADR 0016 amendment**: the ADR presents the two sources as co-equal alternatives. Record that
  authored is preferred on the `Sheet` surface, that auto is soft-deprecated there, and that
  `build_drawing`'s automatic path is unaffected.

## Acceptance

- `auto_dimensions()` and `add_dimension()` each emit `SoftDeprecationWarning` naming the
  replacement, and the message does **not** promise removal.
- `test_deprecation_dates` is unaffected — the new warnings are not `DeprecationWarning` and are
  not scanned by it. A test asserts that, so a future change to `DeprecationWarning` fails loudly
  rather than silently acquiring a removal obligation.
- `build_drawing`'s automatic path emits **no** warning. Guarded, since that is the boundary most
  likely to be crossed by accident.
- `Sheet.from_part(...)` followed by `dimension(...)` lines still works without warning — the
  recommended migration must not nag.
- The suite does not warn its way to noise: tests that legitimately exercise auto either move to
  authored or assert the warning deliberately.

Related: ADR 0016, #874 (made the source explicit), #876 (omission means suppression), #921 (the `from_part` override).

pzfreo:
Closes #1043. Authored dimensions become the going-forward surface on `Sheet`.

**Discouraged is not deprecated.** `auto_dimensions()` keeps working, with no removal planned. That distinction is the design decision here, not an implementation detail.

## Why authored, on this surface

- **It is what draftwright emits.** `--script` writes `authored_dimensions()` and a line per dimension, so the automatic path is the only form the tool itself never produces.
- **"Omission means suppression" (#876) only holds for an authored set.** Under the automatic one there are no omissions to read, so an author cannot say *not that one* — the expressiveness ADR 0016 added is unavailable.
- **An authored list is editable text**, which is what makes "generate a script, then refine it" work for a person or a model.

A caller who chose the declaration surface chose to be explicit about *features*. Being implicit about *dimensions* in the same script is the inconsistency.

## Why soft, and why that is not a dodge

`docs/deprecations.md` carries ADR 0005 §4 — a compat surface names a removal target, because *"a facade with no exit date is a failure mode, not a success"* — and `test_deprecation_dates` enforces removal **language plus a version**.

Since we do not intend to remove this, a `DeprecationWarning` would mean writing a date we do not mean: the exact failure §4 names, wearing a date. So it raises **`SoftDeprecationWarning`** (a `UserWarning` subclass) and lives in a separate *Discouraged* table with no removal target.

Two things make this the better instrument rather than merely the more honest one:

1. `DeprecationWarning` is **filtered by default** in notebooks, several test runners, and library-internal call paths. A `UserWarning` shows unconditionally — for dissuasion, the softer label is the louder warning.
2. For steering LLMs, the **docstring** does most of the work; they read source far more reliably than they observe runtime warnings. Both docstrings now lead with the recommendation.

## Explicitly unaffected, both guarded

- **`build_drawing(part)`** — the detected front door. Point the CLI at a STEP or a build123d object and get a dimensioned drawing. Automatic by design; no warning.
- **`Sheet.from_part(part)`** — the on-ramp, where adding `dimension(...)` lines *overrides* the implicit choice (#921). Warning there would scold the recommended migration.

## Also here

- The no-dimension-source error leads with `authored_dimensions()`.
- The README example is authored — run under `-W error::UserWarning` to prove the documented path is silent.
- **ADR 0016 Amendment 4.** The ADR presents the two sources as co-equal, so without this the deprecation contradicts the decision record.

## The one judgement call

The suite calls `auto_dimensions()` at **~170 sites across 21 files**, deliberately — it is a supported path whose behaviour must keep being tested. Migrating them would be churn that tests *less*, so the category is silenced in `pyproject.toml` with the reasoning inline. `pytest.warns` fires regardless, so the guards are unaffected: **zero warning lines, 347 passed** on the noisiest suites.

## Verification

- **1402 passed** across eight affected suites; four gates clean.
- **Mutations, both red**: making it a `DeprecationWarning` fails the category guard; letting the message read like a scheduled removal fails two others.
- Guards cover the automatic path staying silent and the migration path not nagging — the two boundaries most likely to be crossed by accident.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -126,6 +126,10 @@ entry. Keep `_LAYERS` and this section in step.
   detected as declared (#969); the DAG's bottom leaf (guarded
   by `test_geometry_is_a_leaf`) so the IR waist uses them without importing
   `_core`.
+- **`_warnings.py`** — the public warning categories (`SoftDeprecationWarning`), a
+  dependency-free leaf. Separate from `_core` on purpose: `_core` imports build123d, so a
+  category defined there costs the CAD kernel (~6 s) to reach, and the pytest
+  `filterwarnings` entry naming it pays that on every invocation (#1043).
 - **`fits.py`** — the ISO 286 fit tables (`fit_deviation`, `FitClass`; ADR 0011
   P2a.2): a rank-0 leaf consumed by `_core`, `model/ir` and `sheet`.
 - **`intents.py`** — the deferred-placement "low IR" behind `Drawing.finalize()`
--- a/README.md
+++ b/README.md
@@ -96,13 +96,16 @@ plate = Box(120, 80, 20)
 bore = Pos(0, 0, 0) * Cylinder(4, 20)
 
 sheet = Sheet(plate - bore, title="Plate", number="DWG-002")
-sheet.auto_dimensions()                                   # the planner's set (ADR 0016:
-                                                          # a build must say where its
-                                                          # dimensions come from)
-sheet.envelope()
+env = sheet.envelope()
 sheet.datum("A", plate.faces().sort_by()[-1])             # datum A on the top face
 hole = sheet.hole(bore).finish("1.6").note("M3x0.5 TAP")  # ⌀8 bore, Ra 1.6, tapped
 sheet.control(hole).position(0.1, to="A", diameter=True)  # ⌀0.1 position wrt A
+
+sheet.authored_dimensions()                               # THIS is the complete set
+sheet.dimension(env, "width.length")                      # (ADR 0016: a build says where
+sheet.dimension(env, "height.length")                     #  its dimensions come from, and
+sheet.dimension(hole, "bore.diameter")                    #  omitting one means suppress it)
+
 sheet.export("plate")                                     # writes plate.pdf
 ```
 
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -1251,3 +1251,61 @@ facade and not on the IR type. Worth revisiting if the raw route grows users.
   caught the grid-pitch collision, because two ungrouped pitches are two addressable units);
   a test that an ambiguous selector raises rather than picking; and a stability test that
   re-detecting a part yields the same `ParameterId`s.
+
+## Amendment 4 — authored is the preferred source on `Sheet`; automatic is soft deprecated there (2026-08-05)
+
+This ADR presents the two dimension sources as co-equal alternatives: *either*
+`auto_dimensions()` *or* an authored set, mutually exclusive, and a build must say which.
+The mutual exclusion and the requirement to say stand. The even-handedness does not.
+
+**On the `Sheet` surface, authored dimensions are the recommended source, and
+`auto_dimensions()` / `add_dimension()` are soft deprecated** (#1043) — discouraged, still
+supported, and **not scheduled for removal**.
+
+### Why the surfaces differ
+
+The choice reads as symmetric in the abstract and is not, once you look at what each surface
+is for.
+
+- **It is what draftwright emits.** `--script` writes `authored_dimensions()` and one
+  `dimension(...)` line per measurement. The automatic path is the only form the tool itself
+  never produces, so a hand-written script using it diverges from every generated one.
+- **Amendment/§ "omission means suppression" only holds for an authored set.** Under the
+  automatic source there are no omissions to read, so an author cannot say "not that one" —
+  the very expressiveness this ADR added is unavailable.
+- **An authored list is editable text.** That is what makes "generate a script, then refine
+  it" work, for a person or a model. An automatic set is decided at runtime, cannot be seen
+  in the file, and cannot be edited line by line.
+
+A caller who has already chosen the declaration surface has chosen to be explicit about
+*features*. Being implicit about *dimensions* in the same script is the inconsistency.
+
+### What is explicitly unaffected
+
+`build_drawing(part)`'s automatic dimensioning — the detected front door. Point the CLI at a
+STEP file or a build123d object and get a fully dimensioned drawing with no ceremony. That
+path is automatic by design and carries no warning; a guard asserts it.
+
+`Sheet.from_part(part)` also keeps its implicit automatic source. It is the on-ramp — detect
+the features, get a drawing, then author over it — and adding `dimension(...)` lines
+*overrides* rather than conflicts (#921). Warning there would scold the recommended
+migration.
+
+### Why soft, and why not a removal date
+
+`docs/deprecations.md` carries ADR 0005 §4: a compat surface names a tracking issue *and* a
+removal target, because "a facade with no exit date is a failure mode, not a success".
+
+That rule governs surfaces kept alive only so old code keeps working. This is a different
+thing: `auto_dimensions()` works, is supported, and there is simply a better way to say the
+same thing. Attaching a removal date we did not intend would be the exact failure §4 names,
+wearing a date.
+
+So it raises `SoftDeprecationWarning` (a `UserWarning` subclass), not `DeprecationWarning`,
+and lives in a separate "Discouraged" table with no removal target. A guard asserts the class
+relationship, so a future change to `DeprecationWarning` fails at the definition rather than
+silently acquiring an obligation.
+
+Incidentally the softer label is the louder signal: `DeprecationWarning` is filtered by
+default in notebooks, several test runners and library-internal call paths, whereas a
+`UserWarning` shows unconditionally.
--- a/docs/deprecations.md
+++ b/docs/deprecations.md
@@ -12,7 +12,7 @@ something.
 ## Discouraged — supported, no removal planned
 
 These are **soft deprecated**: they warn, they steer you elsewhere, and they are **not going
-away**. They raise `SoftDeprecationWarning` (a `UserWarning` subclass), *not*
+away**. They raise `draftwright.SoftDeprecationWarning` (a `UserWarning` subclass), *not*
 `DeprecationWarning`, and they deliberately carry **no removal target**.
 
 That is not a violation of ADR 0005 §4's exit-date rule — it is the reason the two categories
@@ -23,6 +23,15 @@ failure §4 names, wearing a date.
 
 `tests/test_deprecation_dates.py` scans `DeprecationWarning`s only, so these rows are outside
 it by construction.
+
+To silence the category in your own code:
+
+```python
+import warnings
+from draftwright import SoftDeprecationWarning
+
+warnings.filterwarnings("ignore", category=SoftDeprecationWarning)
+```
 
 | Surface | Prefer | Since | Why |
 |---|---|---|---|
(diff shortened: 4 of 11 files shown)


pzfreo:
## Adversarial review (Codex) — triaged, both findings real

Fixed in `9d7e05f`.

### 1 (Medium) — the `add_dimension` guard could not fail — **fixed**

Correct, and verified before fixing: **deleting `add_dimension`'s warning entirely left all seven tests green.**

`add_dimension` is only legal against an automatic set, so reaching it necessarily warns twice. I wrapped both calls in one `pytest.warns` and asserted on `rec[-1]` — and `auto_dimensions`' warning satisfies every assertion on its own, including both required phrases. The test was checking that *something* warned, not that `add_dimension` did.

Worse, I had written a comment saying *"reaching it necessarily warns twice"*. I saw the fact and read it as an ordering detail rather than as the thing that made the assertion vacuous.

Each verb is now triggered in isolation: any prerequisite warning is absorbed outside the capture, exactly one warning is asserted, and the message must **name the verb it came from**. The no-removal-version guard is parametrised over both verbs — it previously exercised only `auto_dimensions`, so `add_dimension` could have started promising a removal without failing anything, exactly as you noted.

Both mutations now red: deleting `add_dimension`'s warning fails two guards; letting its message read like a scheduled removal fails a third.

### 2 (Low) — the warning category was only importable privately — **fixed**

Also right, and it was self-inflicted in a visible place: `pyproject.toml` filtered `draftwright._core.SoftDeprecationWarning` — a private path in public configuration — while `docs/deprecations.md` described the class as user-facing policy and its own docstring justified the dedicated subclass so callers *could* filter it.

Now exported from the package root and in `__all__`. The suite filter uses the public path, and the docs carry the snippet:

```python
import warnings
from draftwright import SoftDeprecationWarning

warnings.filterwarnings("ignore", category=SoftDeprecationWarning)
```

### On your "areas that look sound"

Useful and I checked them rather than assuming: no production caller of either verb, `from_part`'s implicit source genuinely bypasses both, and the suite-wide filter does not touch ordinary `UserWarning` or existing `DeprecationWarning` categories. Your read that the filter mechanism was not the weakness — the false-positive guard was — is the right diagnosis.

### Verification

- 669 passed across six suites **after rebasing onto current `main`** (the earlier green CI ran against a stale base and is superseded).
- Four gates clean.
- Three mutations red, including the exact one that previously passed silently.

pzfreo:
## Second adversarial review — triaged, all three real

Fixed in `eab3145`.

### 1 (Medium) — the root export dragged the CAD kernel — **fixed**

Correct. I had found this independently while the review was running, and reached the same diagnosis and the same remedy (`_warnings.py`). Measured before:

```
from draftwright import SoftDeprecationWarning   ->  5.91 s
pytest --collect-only (one tiny file)            ->  6.33 s
```

The `filterwarnings` entry naming the category paid that on **every** pytest invocation, including `--collect-only` and the ~30 s smoke tier. A warning class is the last thing that should have a geometry kernel behind it — and it was my own previous commit that put it there, while fixing your round-1 finding.

After the move to a dependency-free leaf: **0.10 s** to import the category, and `pytest --version` (which parses the filter but collects nothing) is **0.26 s**. The remaining ~6 s of `--collect-only` is the test module importing the engine itself — pre-existing and unrelated, confirmed by re-measuring with `--override-ini=filterwarnings=`.

**The more valuable half of your finding was the part I had not acted on**: the suite could not have caught this, because the test module imports build123d before importing the category, so an in-process `sys.modules` check passes wherever the class lives. There is now a subprocess guard, and moving the class back into `_core` fails it.

### 2 (Low) — the module docstring example raises — **fixed**

Right, and worth noting it has been wrong since #874 made the dimension source explicit — `Sheet` + features + `export()` with no source hits `_check_dimension_source`. Not introduced here, but the surrounding prose ("the auto-pass dimensions exactly the declared features") now contradicted the policy this PR sets, so leaving it would have been actively misleading. Rewritten to the authored form, with the soft-deprecation noted.

### 3 (Low) — `stacklevel=2` was implemented but unguarded — **fixed**

Correct, and it is the same family as round 1's finding: the behaviour was right, nothing would have noticed it stopping. A warning attributed to `sheet.py` tells a user nothing about which of *their* lines caused it, which for a "stop using this" message is the entire point. Now asserted per verb against `__file__`; mutating either site to `stacklevel=1` fails both.

### Verification

- 11 passed in the soft-deprecation suite; four gates clean.
- **Three mutations, all red**: `stacklevel=1` (two guards), class back in `_core` (import boundary).
- CI re-running on `eab3145`.

Thank you for the `sys.modules` suggestion specifically — I had fixed the code but would have shipped it unguarded, which is exactly how it would come back.

pzfreo:
## Third adversarial review — triaged, all three real

Fixed in `89d4d02`.

### 2 (Medium) — two `SoftDeprecationWarning` classes — **fixed**, and the important one

`_core` kept the original definition alongside the new leaf, so `Public is not Core`. `Sheet` emits the public one, meaning anyone filtering or catching the other matched nothing — silently — while the kernel-dragging definition the move existed to remove was still there.

**How it happened is worth recording.** The deletion was correct. Cleaning up a mutation, I ran `git checkout src/draftwright/_core.py` to undo it — which restored the file from HEAD and clobbered my *uncommitted deletion in that same file*. I then committed the result. Using `git checkout` as an undo, in a file I had deliberate uncommitted changes to.

**No behavioural test could have caught this.** `pytest.warns(SoftDeprecationWarning)` passes with either class present, and so does every other guard in the suite. The new one scans the source and asserts exactly one definition — the only form that can see the two-class state at all.

### 1 (High) — the branch failed its own import-boundary gate — **fixed**

Correct. I found and fixed this independently while your review was running, but only *after* telling the maintainer the head was ready — so the sequencing was mine to get wrong. `_warnings` is now rank 0 with its reason, and `CLAUDE.md`'s architecture map is updated, which the failure message explicitly asks for.

Worth noting for its own sake: this is the second time this session a structural guard was missed by a targeted selection because its name does not resemble the code being touched (`test_detect_once` was the first). CI would have caught both; my selection should have.

### 3 (Low) — missing from the `TYPE_CHECKING` block — **fixed**

Right. The name was in `_LAZY` and `__all__` but not in the static-analysis branch that exists precisely to expose lazy names, so a checker resolved `from draftwright import SoftDeprecationWarning` to `Any` rather than the class. Added.

### Verification

- **1280 passed** across eight suites; four gates clean.
- **Mutations, all red**: restoring the duplicate → single-definition guard; unranking the leaf → two boundary guards; making the leaf import build123d → import-boundary guard.
- I verified the three fixes are actually present in the tree before committing, rather than assuming — given that a `git checkout` silently reverted one of them an hour earlier.

### Standing back

Seven findings across three rounds on this PR, and **four were regressions introduced while fixing earlier ones**. That ratio is the signal, not the individual bugs: I was iterating faster than I was verifying. The `git checkout` mistake in particular is something a slower pass would have caught, and it produced a defect that no behavioural test in the suite could see.

PR #1047, opened 2026-08-05
Render slanted flats with canonical stock identity

### LINKED ISSUE #1036: Slanted flats do not render, and their stock identity key is not canonical

pzfreo:
## What

Render recognised slanted machined flats without making their stock grouping ambiguous.

A flat stock region is now identified by:

- the real axis direction, canonicalised with its dominant component positive;
- the axis line's perpendicular foot from the origin, retaining the existing two in-plane coordinates;
- the axial span expressed along that canonical direction.

That identity travels recognition → IR → compiled facts → completeness correspondence → emitted declaration. Flat placement keeps the established centre-out candidates first, then uses the boundary-aware fallbacks from #1034 for aligned and slanted stock alike.

The temporary `axis_aligned` field from #1046 is no longer stored state. It remains a derived convenience property; `axis_direction` is the source of truth.

## Observable result

The normal-scale #1036 fixture now has:

- one recognised slanted flat requirement;
- one placed `m_flat_x0=13 A/F` callout;
- compiler measurement provenance on that callout;
- no `flat_dropped`, completeness gap, overlap, or silhouette-crossing issue.

Two parallel slanted stocks produce two definitions. A slanted double-D still produces one grouped definition with both measurement identities.

## Adversarial review

1. **Precision at real-case scale.** Computing the perpendicular foot from a six-decimal direction can amplify angular rounding at 40-metre coordinates. Geometric projection now uses the full-precision unit vector; only the stable identity is rounded.
2. **Emitter fidelity.** The generic point formatter kept three decimals, which is drawing-coordinate precision rather than angular precision. Axis directions now emit at six decimals, and canonicalisation is idempotent across emit → declare.
3. **Opposition still ignored direction.** The recogniser's old same-line check projected positions along one direction but never verified the other cylinder shared it. Different slanted lines through one foot could still pair as opposing faces. It now compares canonical directions first.
4. **Direction sign and span.** Reversing a supplied direction now reverses and reorders its axial span, so equivalent unoriented axes retain one identity.
5. **Load-bearing mutations.** Replacing the completeness direction with the principal axis collapses two requirements to one; removing the recogniser direction comparison treats different lines as one. Both named guards fail under mutation and pass restored.

Only naming/compatibility nits remain.

## Validation

- Expanded recognition/IR/emit/declare/completeness suite: `722 passed, 2 xfailed`
- Targeted restored mutation guards: `2 passed`
- `ruff format --check src tests`
- `ruff check src tests`
- `mypy src/draftwright`
- `codespell src tests docs`

Closes #1036.

### DIFF
--- a/src/draftwright/_geometry.py
+++ b/src/draftwright/_geometry.py
@@ -12,6 +12,7 @@ drawing grab-bag (ADR 0008; #584 WP2). This module imports nothing from
 from __future__ import annotations
 
 import logging
+import math
 from dataclasses import dataclass
 
 from build123d import Compound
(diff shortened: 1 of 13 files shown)


pzfreo:
Final gate on `c086b40`: lint and all 15 Ubuntu/macOS/Windows Python 3.10-3.14 jobs passed; Codecov patch 95.18% and project 91.62% both passed. No human review comments, requested changes, or unresolved threads.

Issue #909, opened 2026-07-28
Recognise and fully dimension sloped multi-tier prismatic parts


pzfreo:
## Case study

`Basic_Part_Design_Tutorial_017-Body.step` is a 53 × 37.7 × 33 mm FreeCAD/Open CASCADE part containing multiple upper tiers, a notch/recess, and a sloped side face. On current `main` (`7b61a35`), the CLI exports a visually clean A4 drawing but it is not manufacturing-complete.

## Current result

The recognised IR contains only:

```text
envelope     53 × 37.7 × 33
step_level   base=-13, levels=(8, 13), one Y shoulder
```

The drawing consequently emits only:

```text
53 overall width
37.7 overall depth
33 overall height
21 front step height
21 side shoulder location
```

Lint reports only:

```text
warning step_dim_dropped: 1 step height(s) too closely spaced to dimension at this scale
```

with a misleading 0.95 score, because the unrecognised geometry never reaches completeness lint.

## Missing manufacturing definition

- upper block/bridge widths and offsets
- the intermediate notch/recess dimensions
- the sloped face angle, or equivalent endpoint dimensions
- remaining tier/shoulder heights, including the closely spaced step
- enough dimensions to reconstruct the profile without interrogating the STEP model

The two visible `21` dimensions are on different axes and are not themselves duplicates.

## Expected architecture

This should enter the normal compiler pipeline rather than receiving hand-placed annotations:

1. Recognition projects the prismatic profile into structured planar/tier, notch, and slope features.
2. The planner selects a minimal non-redundant dimension basis for those features.
3. Crowded tier dimensions use the existing detail-view escalation rather than dropping silently.
4. Completeness lint compares recognised/profile-defining geometry with emitted coverage, so a sparse IR cannot produce an overly optimistic score.
5. Placement remains collect → solve → emit per ADR 0014; no raw annotation coordinates.

## Suggested synthetic fixture

Add a build123d solid with:

- rectangular base envelope
- two narrower stacked upper tiers
- one rectangular side notch
- one oblique planar side face
- one pair of closely spaced Z levels requiring a detail view

This avoids depending solely on the external FreeCAD case-study STEP and permits fast PR-gated mutation tests.

## Acceptance criteria

- [ ] Synthetic fixture recognises structured tier/notch/slope features, not only envelope + `step_level`.
- [ ] Drawing supplies a sufficient, non-redundant manufacturing dimension basis for every profile transition.
[...]

PR #1054, opened 2026-08-06
Recognise wedge-mounted raised pads

pzfreo:
Part of #909.

## What changed

- pin the byte-identical `Basic_Part_Design_Tutorial_017-Body.step` artifact
- classify a rectangular upper pad when lower ledges touch it only in plan and sit below its recovered local base
- retain the existing exclusion for a true tiered staircase whose ledges meet the candidate at that base
- guard the real record through the recognition aggregate, IR, four owned dimensions, completeness lint, detail placement, and generated declaration

## Evidence

The real body has 13 planar faces and no pocket. Existing face-level/riser evidence already describes the wedge: `detail_view=True` places the 21 mm and 26 mm level rungs and the 21 mm slope station with clean lint. The remaining first-slice false negative is its 31 x 16.7 upper pad.

The pad recognizer proves the top and all four perimeter walls, then discarded the record because two lower side ledges touch its plan boundary at Z=8. The recovered local pad base is Z=13; those ledges are not a nested tier. A true staircase has touching ledges at the candidate base and remains excluded.

## Result

The real drawing gains one `PadFeature` plus 31 x 16.7 footprint dimensions and X/Y location dimensions, all owned by the feature with compiler measurement provenance. Removing a size produces `pad_footprint_not_defined`. With the existing detail option the case has clean lint.

## Adversarial checks so far

- restoring the old any-other-Z rejection loses the real pad;
- choosing the lowest rather than highest perimeter-wall base loses the real pad;
- requiring positive-area overlap admits the top tier of the existing forbidden staircase fixture.

## Validation so far

- 31 focused issue/pad/recognition-contract tests
- locked Ruff format/lint, mypy, and codespell

The inclined-face-pocket auxiliary-view request recorded later on #909 has no geometric witness in this artifact and is deliberately outside this slice.


pzfreo:
Adversarial review is complete on `e405612`.

Production behavior was challenged with three mutations:

- restoring the old “any other Z” rejection loses the real #909 pad;
- replacing the recovered local base with the minimum candidate base also loses it;
- weakening staircase detection to positive-area overlap admits the existing nested-staircase counterexample.

The tests were then reviewed for overclaim. They now pin the pad record, aggregate ownership, required size/location annotations, deletion lint, generated-script round trip, and the detail ladder without freezing the rest of the future #909 inventory or a redundant Y-location policy.

Local validation:
- full locked gate: formatter, Ruff, mypy, codespell, then `2843 passed, 1 skipped, 19 deselected, 2 xfailed`;
- final test-wording SHA: all static checks plus `32 passed` focused tests.

No substantive review findings remain. The last commit only clarifies the test local-base rationale; CI has restarted on that SHA.

Issue #915, opened 2026-07-28
Dense prismatic drawings need coordinated callout packing and step-detail escalation


pzfreo:
## Case study

`19-case_study_2` on current `main` (`a3f3c5f`) recognises a useful model:

```text
fillet: 4, hole: 3, pattern: 2, pocket: 2, envelope: 1, step_level: 1
```

and generates plan/front/side/isometric plus section A–A. However:

- several long hole/pattern callouts are packed into the same plan-view region and are visually difficult to separate;
- lint reports an `annotation_overlap`;
- five defining step heights are dropped at 1:2 on A2;
- the engine recommends a detail view but does not automatically use that remedy.

```text
annotation_overlap: labels '' and '' overlap by 36.8×1.7 mm
step_dim_dropped: 5 step height(s) too closely spaced to dimension at this scale
```

A large page with substantial empty area still ends with crowded local callouts and omitted definition, indicating that view-local placement failure is not being fed back into page/view/detail composition.

## Synthetic reproduction

Use a smaller compound plate that creates the same competing annotation classes:

```python
from build123d import Box, Cylinder, Pos
from draftwright import build_drawing

part = Box(100, 70, 24)
part -= Pos(0, 0, 12) * Cylinder(18, 18)       # central blind pocket/bore
for x, y in ((-30, -20), (30, -20), (-30, 20), (30, 20)):
    part -= Pos(x, y, 0) * Cylinder(3, 24)     # four-hole pattern
for x in (-18, 18):
    part -= Pos(x, 0, 14) * Cylinder(5, 10)    # competing blind features
# Closely spaced top levels that need a chain or detail.
part += Pos(0, 0, 24) * Box(70, 50, 3)
part += Pos(0, 0, 27) * Box(45, 35, 2)

dwg = build_drawing(part, page="A4")
```

Adjust only fixture construction details needed for stable recognition; preserve the essential combination: a pattern, multiple compound callouts, and closely spaced step levels.

## Expected architecture

This needs coordinated planning, not per-renderer nudges:

1. collect all competing callout and step candidates;
2. solve shared corridor occupancy with full annotation geometry;
3. if defining steps cannot fit, escalate to an automatically composed detail or a larger/rebalanced view allocation;
4. rerun placement after composition changes;
5. report unresolved feature coverage by provenance.

This should use the existing escalation and detail-view machinery and ADR 0014 placement solve. No raw coordinates or case-specific offsets.

## Acceptance criteria

[...]

PR #1055, opened 2026-08-06
Preserve step-detail measurement provenance

pzfreo:
## Summary

- register each enlarged prismatic detail rung against the approved compiled measurement it draws
- attribute those detail annotations to the source step-level feature, so `annotations_of()` and `drop(feature)` remain complete
- remove the audit ratchet exemption that incorrectly classified detail dimensions as non-measurement geometry
- pin the contract against the real #909 STEP artifact without requiring the duplicate main-view rung to survive

## Evidence

On current `main`, `dim_detail_a_step0` and `dim_detail_a_step1` render the approved 21 mm and 26 mm rungs but both return `[]` from `measurement_keys()` and neither belongs to the step feature.

Mutations removing `measurement=` and `feature=` fail independently. The first is caught by both the real-artifact assertion and the direct-placement inventory; the second is caught by semantic ownership and `drop(feature)`.

## Validation

- formatter, Ruff, mypy, codespell
- 64 focused tests covering #909, the audit differential, detail script parity, #915 dense details, and slanted-profile details

Part of #909 and #996. This PR does not decide whether detail views should be automatic or whether the main-view 21 mm rung should remain duplicated.

### DIFF
--- a/src/draftwright/annotations/sections.py
+++ b/src/draftwright/annotations/sections.py
@@ -759,7 +759,7 @@ def _request_prismatic_detail(dwg, a: Analysis, *, ctx, plan) -> None:
     # source view would have.
     levels = [r.span[1][2] for r in rungs]
     datum_z = rungs[0].span[0][2]
-    label_of = {r.span[1][2]: r.final_label for r in rungs}
+    rung_by_level = {r.span[1][2]: r for r in rungs}
     has_shoulders = plan.ladder("step_position") is not None
     z0, z1 = min(levels), max(levels)
     pad = 0.08 * (z1 - z0) + 1.0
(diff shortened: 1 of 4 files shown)


pzfreo:
Adversarial review is complete on `c544431`.

Review findings and mutations:

1. The first artifact test incorrectly derived detail identity from `dim_step_0`, which would have frozen the known duplicate main-view rung. The oracle now compares each detail label with the compiled ladder directly.
2. The first behavioral guard covered only automatic builds. The generated `Sheet` path now proves identical non-empty measurement identity and feature ownership.
3. Replacing `measurement=rung.id` with `measurement=None` fails both the real #909 guard and the direct-placement inventory.
4. Retaining measurement but replacing source feature ownership with `feature=None` fails `annotations_of(step)` and `drop(step)`.
5. Current correlated-ladder identity remains `step_height.length`; this PR restores that existing granularity and does not claim per-rung IDs.

No substantive findings remain. Final validation:
- formatter, Ruff, mypy, codespell: green;
- full local suite: `2844 passed, 1 skipped, 19 deselected, 2 xfailed`;
- all 15 OS/Python CI lanes: green;
- Codecov project and patch: green.

PR #1064, opened 2026-08-06
Speed up quality-gated iteration

pzfreo:
## Symptom and evidence

- User-visible problem: semantic review and coverage feedback arrived only after a roughly 16-minute remote matrix, so failed iterations paid the full wait before exposing known classes of mistake.
- Fixture/reproducer: Epic #1018's adversarial-review history and both #1063 coverage artifacts; the first patch needed another full remote run after Codecov failed.
- Before/after result: contributors now have a one-command quick gate and a fail-closed full local gate; stale PR heads are cancelled, compatibility results arrive earlier, and canonical coverage moves from the measured 16:21 Python 3.12 path to the equivalent 7:15 Python 3.13 path.
- Mutations that make the guard fail: an unresolved `BASE_REF` and an untracked `src/` module both exit before tests; rejected #1063 head `0b939d6` fails the 93% changed-line floor while corrected head `c1e1569` passes at 96%.

## Contract and scope

- Smallest contract this change tests: local preflight mirrors the repository-owned static/full-suite floors, fails when Git cannot see new source, and provides a conservative changed-line warning before the branch is pushed.
- Relevant ADRs: no architectural decision changes; the contribution workflow generalises ADR 0017's evidence-gated, mutation-first lesson.
- Explicitly deferred: do not split test modules until repeated cross-platform duration reports identify a stable scheduling tail; do not add another recognition harness until a third slice demonstrates harmful shared duplication.

## Validation

- [x] `scripts/pr-check --quick` (73 passed on the final local amendment)
- [x] `scripts/pr-check` before final review (2,916 passed, 1 skipped, 2 xfailed; 93.64% combined coverage)
- [x] The named mutations were observed failing before the implementation passed them
- [x] Automatic, declared, and generated-script paths were checked where affected: N/A, infrastructure and contributor-process change only
- [x] Recognition and repeated-lint call counts were checked where affected: N/A, no recognition behavior changed
- [x] Workflow YAML parsing, shell syntax, clean-diff checks, missing-base failure, untracked-source failure, and rejected/corrected #1063 coverage replay

## Stop check

- [x] New prerequisites were split or the work was explicitly reclassified as discovery: no project dependency was added; `diff-cover` runs as an exact isolated `uvx` tool

## CI impact

- Retains all 15 Ubuntu/macOS/Windows Python 3.10-3.14 fast-test jobs and the post-merge slow suite.
- Keeps line and branch coverage, the project floor, Codecov, and retained reports on canonical Ubuntu/Python 3.13; Ubuntu/Python 3.12 remains a full compatibility job. Branch protection requires both Codecov project and patch checks.
- The same source baseline previously measured 93.65% coverage in 7:15 on 3.13 versus 93.64% in 16:21 on 3.12.
- Cancels obsolete pull-request heads only; every main push has a unique concurrency group so its slow safety net cannot be cancelled by a later merge.
- Reports the slowest 25 tests on every platform before any test-module split is proposed.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,27 @@
+## Symptom and evidence
+
+- User-visible problem:
+- Fixture/reproducer:
+- Before/after result:
+- Mutation that makes the guard fail:
+
+## Contract and scope
+
+- Smallest contract this change tests:
+- Relevant ADRs:
+- Explicitly deferred:
+
+## Validation
+
+- [ ] `scripts/pr-check --quick`
+- [ ] `scripts/pr-check` before final review
+- [ ] The named mutation was observed failing before the implementation passed it
+- [ ] Automatic, declared, and generated-script paths were checked where affected
+- [ ] Recognition and repeated-lint call counts were checked where affected
+
+Use `N/A` with a reason for items that do not apply, such as fixture evidence on a docs-only PR.
+
+## Stop check
+
+- [ ] No two consecutive review rounds invalidated the same core association strategy
+- [ ] New prerequisites were split or the work was explicitly reclassified as discovery
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,11 +39,11 @@ jobs:
         # 0.11 (cadquery-ocp-novtk — VTK dropped, and the only kernel with 3.13/3.14 wheels).
         # A few exact-position snapshots are skipped on 0.11 (see tests/_kernel.py).
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
-        # Ubuntu/Python 3.12 is the canonical coverage job below. Excluding it
+        # Ubuntu/Python 3.13 is the canonical coverage job below. Excluding it
         # here keeps the OS/Python matrix complete without running it twice.
         exclude:
           - os: ubuntu-latest
-            python-version: "3.12"
+            python-version: "3.13"
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -67,7 +67,7 @@ jobs:
         # No --timeout here: it would override the 300 s set in
         # [tool.pytest.ini_options]. The heaviest test (full specimen-sheet SVG
         # export) can exceed 120 s on the slower macOS runners.
-        # Coverage is measured once by the canonical Ubuntu/Python 3.12 job below.
+        # Coverage is measured once by the canonical Ubuntu/Python 3.13 job below.
         # Instrumenting all fourteen compatibility jobs adds ~13% runtime without
         # producing another retained or uploaded report.
         run: >-
@@ -77,7 +77,7 @@ jobs:
   coverage:
     # The fifteenth OS/Python matrix entry, split out so only the one job that
     # uploads to Codecov receives an OIDC token and renders retained reports.
-    name: test (ubuntu-latest, 3.12)
+    name: test (ubuntu-latest, 3.13)
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -88,7 +88,7 @@ jobs:
       - uses: astral-sh/setup-uv@v8.2.0
         with:
           enable-cache: true
-          python-version: "3.12"
+          python-version: "3.13"
 
       - name: Install dependencies
         run: uv sync --group dev
@@ -115,7 +115,7 @@ jobs:
         if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v7
         with:
-          name: coverage-linux-py3.12
+          name: coverage-linux-py3.13
           path: |
             coverage.xml
             htmlcov/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,10 +5,10 @@ on:
     branches: [main]
   pull_request:
 
-# A PR needs results only for its latest head. Main runs are not cancelled because they own
-# the post-merge slow suite, which must not be starved during a burst of merges.
+# A PR needs results only for its latest head. Each main push gets a unique group because it
+# owns a post-merge slow-suite run that must survive later merges.
 concurrency:
-  group: ci-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ci-${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
@@ -106,9 +106,9 @@ jobs:
         uses: codecov/codecov-action@v6
         with:
           files: coverage.xml
-          # The repository's fail_under gate is authoritative; a third-party
-          # service outage must not block an otherwise valid change.
-          fail_ci_if_error: false
+          # Branch protection requires Codecov's project and patch checks. Fail this
+          # job too when the upload cannot produce those checks, rather than hanging.
+          fail_ci_if_error: true
           use_oidc: true
 
       - name: Upload coverage report
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,12 @@ on:
     branches: [main]
   pull_request:
 
+# A PR needs results only for its latest head. Main runs are not cancelled because they own
+# the post-merge slow suite, which must not be starved during a burst of merges.
+concurrency:
+  group: ci-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   lint:
     runs-on: ubuntu-latest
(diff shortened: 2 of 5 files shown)


pzfreo:
## Adversarial review complete

Three substantive findings were fixed before publication:

1. Adding `diff-cover` as a dev dependency caused 347 lines of unrelated lock normalization, so the gate now uses exact isolated `diff-cover==9.7.2` via `uvx` and leaves the project dependency graph untouched.
2. An invalid `BASE_REF` was initially checked after the full suite; it now fails in zero seconds before static or test work.
3. Cancelling main-branch runs could starve the post-merge slow suite during rapid merges; cancellation is now limited to obsolete pull-request heads.

Final validation is green on `5c2283e`: lint, all 15 Ubuntu/macOS/Windows Python 3.10-3.14 jobs, Codecov patch, and Codecov project. Local full preflight passed 2,916 tests with 93.64% combined coverage.

The first duration report does not justify a test-module split in this PR. Canonical coverage remains the 16:21 critical path, led by separate 61.21s `test_issue_915_dense_case` and 60.45s `test_dense_sheet_canary` calls. Compatibility feedback is materially earlier: Ubuntu 3.11 fell from about 15:33 to 8:36 and macOS 3.11 from about 15:33 to 9:22. Only follow-up profiling/refactoring nits remain.

pzfreo:
## Fresh adversarial review — changes required

### 1. Untracked source files bypass the advertised changed-line gate

`scripts/pr-check:38-40` delegates the source set entirely to Git/diff-cover. A probe added as a new, uncovered `src/draftwright/_preflight_untracked_probe.py` remained `??` in `git status`; diff-cover then reported `No lines with coverage information in this diff` and exited 0. `git diff` includes staged and unstaged tracked files, but not untracked files. A final preflight run before the first `git add` can therefore certify a brand-new uncovered module. Fail closed on untracked `src/` files (with an instruction to stage them), or explicitly include them in the coverage comparison.

### 2. The 92% local gate does not catch the motivating #1063 failure

I replayed the retained canonical coverage artifact from first head `0b939d6` against base `2bbbc26`, normalising only the XML source root in the same way this PR's `relative_files = true` does. Diff-cover measured 470 changed lines, 36 missing: 92.34%, rendered as 92%, and exited 0 at `--fail-under=92`. Codecov rejected that exact SHA at 85.10% patch coverage and also rejected project coverage (91.53%, -0.23%). The line-only lower bound is legitimate, but the present threshold would not have prevented the second remote cycle used to justify it. At minimum calibrate above this counterexample (93% rejects it and the corrected head remains above it), and narrow any claim of Codecov equivalence; a line-only gate cannot reproduce branch partials or the project ratchet.

### 3. The successful PR critical path is not faster yet

The new Actions run took 16:50 (`14:48:43` to `15:05:33`); the final #1063 run took 16:37. Compatibility feedback and runner consumption improved substantially, which is useful, but the unchanged canonical coverage job remains the critical path. The PR should distinguish runner/early-failure savings from time-to-green. If the primary acceptance is faster complete iterations, that remains a measured follow-up rather than an outcome of this change.

### Nits / residual risks

- Exact `diff-cover==9.7.2` still resolves unpinned transitive dependencies through `uvx` and needs network/cache outside `uv.lock`; it is isolated, not fully reproducible.
- `origin/main` is a reasonable owner-checkout default, but fork contributors may need `BASE_REF=upstream/main`; call that out explicitly.

The worktree is clean after both probes. Existing local/full CI results remain valid, but findings 1 and 2 should be fixed before merge.

pzfreo:
## Adversarial findings fixed in `a2d14c7`

1. **Untracked source:** the full preflight now enumerates untracked `src/` paths before static/tests, fails in zero seconds, names each path, and requires staging. The original uncovered-module mutation now exits 2 instead of silently passing.
2. **#1063 calibration:** the floor is 93%. Replaying retained artifact `0b939d6` now exits 1 at 92.34% line coverage; corrected head coverage passes at 96%. Documentation explicitly says this is not branch/project Codecov parity.
3. **Critical path:** canonical coverage moves from Ubuntu/Python 3.12 (93.64%, 16:21) to Ubuntu/Python 3.13 (93.65%, 7:15 on the same source baseline). Python 3.12 remains in the full compatibility matrix, so no test/platform contract is dropped. Based on the first matrix, the expected successful critical path becomes the roughly 12:43 slowest compatibility job instead of 16:21 coverage.

Local quick/static, workflow parsing, both fail-closed probes, and rejected/corrected coverage replays are green. The PR stays draft until the role-swapped matrix and both Codecov statuses pass on this SHA.

pzfreo:
## CI rerun blocked by GitHub Actions incident

The validation run for `a2d14c7` was hit by GitHub's active Actions partial outage: multiple Ubuntu/Windows jobs failed in `Set up job` while resolving action download metadata (`Service Unavailable`), before checkout, and GitHub later cancelled queued/running jobs. The Actions cancel API is also returning `502`.

This is not a test or workflow-configuration failure. The branch and SHA are unchanged, and the PR remains draft. I will rerun the complete workflow from the same SHA after GitHub reports recovery: https://www.githubstatus.com/incidents/qcvjkzcs7j74

pzfreo:
## Fresh re-review — changes required

### 1. Codecov is described as a gate but is not required

`CONTRIBUTING.md:61-62` says branch partials and the project-coverage ratchet remain remote gates. Branch protection currently requires only `lint` and the 15 Actions test contexts; neither `codecov/patch` nor `codecov/project` is required. The exact rejected #1063 head `0b939d6` proves the gap: every required Actions context succeeded while both Codecov checks failed. It could have been merged despite the two cases the local line-only gate explicitly cannot detect.

Either make both Codecov contexts required (accepting that Codecov availability can then block merges), or describe them as advisory and enforce equivalent branch/project rules inside a required Actions job. Merely changing the wording would not satisfy this PR's “without sacrificing quality” goal.

### 2. Main runs can still be cancelled while pending

`.github/workflows/ci.yml:8-12` and the PR body say main runs remain uncancelled. They share one concurrency group by `github.ref`. GitHub's current semantics retain at most one pending run by default; a newer main run cancels and replaces an existing pending main run even with `cancel-in-progress: false`. Only an already-running main run is protected.

If every post-merge slow run must survive, give push runs a unique group (for example `github.run_id`) or use explicit queueing. If coalescing pending main commits is acceptable because the newest SHA is cumulative, narrow the comment and PR claim to that actual contract.

### 3. The documented canonical reproduction does not select Python 3.13

`CONTRIBUTING.md:38-48` calls the command a reproduction of canonical Linux/Python 3.13, but it runs whichever interpreter owns the local uv environment. This checkout has no `.python-version`; the documented command currently runs Python 3.12 with build123d 0.10, while canonical CI runs Python 3.13 with build123d 0.11 and different skips. Use `uv run --python 3.13 pytest ...` (or document an equivalent 3.13 environment setup).

### Validation / residual

`bash -n`, workflow YAML parsing, `git diff --check`, Ruff, formatting, mypy, and codespell are green. The worktree is clean and there are no review threads. Final remote validation is still unavailable because the current Actions run was damaged by GitHub's declared incident; that is infrastructure state, not a fourth code finding.

pzfreo:
## Re-review findings fixed in `a76d1f7`

1. **Codecov enforcement:** branch protection now requires `codecov/patch` and `codecov/project`, pinned to Codecov app ID 254, while retaining all 16 existing Actions requirements. The upload step also uses `fail_ci_if_error: true`, so an upload failure cannot leave the required reports silently absent.
2. **Main-run concurrency:** pull-request heads still share their PR-number group and cancel obsolete runs; every main push now uses its unique `github.run_id`, so pending post-merge slow runs cannot be replaced by later pushes.
3. **Canonical reproduction:** the contributor command now uses `uv run --python 3.13`; an isolated verification selected Python 3.13.11 and build123d 0.11.1.

Workflow YAML, shell syntax, diff checks, Ruff, formatting, mypy, and codespell are green. Replacement Actions run `31123437606` exists for this SHA but remains entirely queued behind GitHub's active Actions incident; the PR remains draft and blocked until the required matrix and both Codecov checks report.

pzfreo:
## Final adversarial re-review

No new substantive findings on `a76d1f7`.

- Branch protection contains all 16 Actions requirements plus `codecov/patch` and `codecov/project`, both pinned to Codecov app 254.
- PR runs still coalesce by PR number; main pushes use unique run IDs, so no pending main slow run can be replaced.
- The canonical local command selects Python 3.13/build123d 0.11, and public fork PR uploads remain supported without exposing repository secrets.
- Shell syntax, workflow YAML, clean diff, Ruff, formatting, mypy, and codespell are green; the worktree is clean and there are no review threads.

Only nits/residual risk remain: `uvx --from diff-cover==9.7.2` still resolves transitive tooling outside `uv.lock`, and run `31123437606` has not executed because GitHub's active Actions incident left its check suite queued with zero jobs. That prevents final timing and remote Codecov verification, but is not a code finding.

Issue #1067, opened 2026-08-07
Release v0.4.1 and restore a weekly release decision


pzfreo:
## Outcome

Ship the accumulated post-v0.4.0 correctness and recognition work as a patch release, then make release readiness an explicit weekly decision instead of allowing unreleased changes to accumulate invisibly.

## Acceptance

- [ ] Confirm the complete local and GitHub Actions test tiers are green on the release commit.
- [ ] Review the changelog against every commit since v0.4.0.
- [ ] Publish v0.4.1 through the existing trusted-publishing workflow.
- [ ] Verify the package from the registry in a clean environment.
- [ ] Record the next weekly release decision in the Project status update, even when the decision is not to release.

## Evidence

Link the release, package verification, and workflow run before closing.

pzfreo:
Release readiness is now in progress.

Baseline:
- v0.4.0..main contains 39 commits.
- `pyproject.toml` is already at `0.4.1.dev0`, which the release workflow normalises to `0.4.1`.
- The current Unreleased changelog covers only four themes, so the first slice is a commit-by-commit reconciliation before tagging.
- The last complete main CI run was green; the newly merged planning-only commit is running now and does not block changelog preparation.

pzfreo:
Release preparation PR is open: #1069. The 40-commit audit and validation gates are recorded in the PR body; Project status is now In review.

pzfreo:
## Release evidence

- Release: https://github.com/pzfreo/draftwright/releases/tag/v0.4.1
- Release commit: `801a624242a7bd26fc0d17610986dd5559e5d869`
- Main CI, including slow tier: https://github.com/pzfreo/draftwright/actions/runs/31159315177 — 17/17 jobs green
- Trusted publish: https://github.com/pzfreo/draftwright/actions/runs/31160704632 — build and `publish-pypi` green
- Registry verification: fresh Python 3.12 environment, fresh cache, `draftwright==0.4.1`, no `direct_url.json`, real drawing build produced one recognised feature and clean lint
- Project status update: Week 1 ON_TRACK; next weekly release decision recorded for 2026-08-14

The publish workflow’s post-release direct push failed under branch protection exactly as tracked by #779. PR #1072 restores `0.4.2.dev0`; this issue will close when that protected-branch recovery PR is merged.

pzfreo:
[...]

PR #1072, opened 2026-08-07
chore: bump version to 0.4.2.dev0 after release

### LINKED ISSUE #779: Publish workflow: bump-version fails on protected main (GH006) — open a PR instead of pushing
## Bug

The `bump-version` job in `.github/workflows/publish.yml` (runs after a successful PyPI publish on a GitHub Release) does `git push` **directly to `main`**, which is a protected branch:

```
remote: error: GH006: Protected branch update failed for refs/heads/main.
```

So the post-release version bump fails on **every** release. The **PyPI publish itself is unaffected** (v0.3.4 published fine); only the automated `0.3.x.dev0 → 0.3.(x+1).dev0` bump doesn't land, and must be done by hand (as it was for 0.3.4 → #778).

## Fix options
1. **Open a PR** instead of pushing to main — e.g. `peter-evans/create-pull-request`, or `gh pr create` after committing to a branch. Cleanest; respects branch protection. (Would still need auto-merge or a manual merge.)
2. A token/app with **bypass** permission on the protected branch (a repo/app setting, not a workflow change).

Option 1 is preferred (no elevated-permission token in CI). Until fixed, the maintainer bumps by hand after each release.

Refs the v0.3.4 release; #778 is the manual bump.

pzfreo:
## Outcome

Restore the post-release development version after v0.4.1. The release workflow produced this exact two-line change, then branch protection correctly rejected its direct push; #779 tracks replacing that push with a PR.

Without this bump, later main pushes are rewritten as `0.4.1.devN` for TestPyPI, which sorts below the released v0.4.1 while containing newer code.

## Adversarial review

One pass compared the workflow's logged intended version, the previous release convention, project metadata, lock metadata, built wheel/sdist, and CLI output. No substantive findings remain.

- only `pyproject.toml` and the editable package row in `uv.lock` change;
- both move `0.4.1 -> 0.4.2.dev0`;
- no dependency markers or resolutions changed;
- this does not close #779 because the automation still attempts a protected direct push.

## Validation

- [x] `uv lock --check`
- [x] wheel and sdist build as `0.4.2.dev0`
- [x] wheel METADATA and `draftwright --version` report `0.4.2.dev0`
- [x] `scripts/pr-check --quick` (73 passed)
- [ ] hosted matrix and Codecov

Refs #779 and #1067.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "draftwright"
-version = "0.4.1"
+version = "0.4.2.dev0"
 description = "Automated technical-drawing generation for build123d"
 readme = "README.md"
 license = { file = "LICENSE" }
--- a/uv.lock
+++ b/uv.lock
@@ -675,7 +675,7 @@ wheels = [
 
 [[package]]
 name = "draftwright"
-version = "0.4.1"
+version = "0.4.2.dev0"
 source = { editable = "." }
 dependencies = [
     { name = "build123d", version = "0.10.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.13'" },


pzfreo:
## Final adversarial pass

No substantive findings.

- The two-line diff is identical in scope to the release workflow’s rejected generated commit.
- PEP 440 ordering is verified: `0.4.1 < 0.4.2.dev0 < 0.4.2`.
- The main-push TestPyPI step will strip `.dev0` to base `0.4.2` and assign its unique run-number suffix.
- The lock, wheel, sdist, CLI and quick gate all agree on `0.4.2.dev0`.
- #779 remains open because this PR recovers the state but does not repair the workflow.

Only process nits remain; hosted CI is the remaining gate.

Issue #1073, opened 2026-08-07
Slot and pocket pattern recognition groups features across separate assembly bodies


pzfreo:
## Outcome

A pattern may contain features from one physical body only. Identical slots or pockets on separate components of a multi-solid assembly must remain independent requirements unless an explicit assembly-level semantic declares otherwise.

## Reproduce

Three independently slotted bodies at Y = 0, 60, 120 each correctly produce one slot. The flat aggregate then loses body ownership:

```python
from build123d import Box, Compound, Pos
from draftwright.recognition import recognise_slot_patterns, recognise_slots

slotted = Box(60, 30, 10) - Box(30, 8, 20)
assembly = Compound(children=[Pos(0, y, 0) * slotted for y in (0, 60, 120)])

slots = recognise_slots(assembly)          # three legitimate per-body slots
patterns = recognise_slot_patterns(slots) # one false SlotArray, pitch 60
```

The pocket mirror behaves the same way: three bodies with one 30 x 8 x 10 pocket each become one `PocketArray(pitch=60)`.

## Why this is separate from #958

#958 fabricates a *single feature* by pairing faces from different solids; per-solid face scans fix it locally. This defect happens later: each source feature is valid, but aggregation discards body ownership before the derived pattern family runs.

A traversal-order `solid_idx` on public records would be the easy patch and the wrong identity: ADR 0017 rejects topology/traversal-derived values as stable semantic identity. The design must either retain a geometry-derived body correspondence or keep pattern derivation inside a per-body orchestration while preserving the one-public-family ownership contract.

## Acceptance

- [ ] Three independently slotted bodies produce three slots and no cross-body slot pattern.
- [ ] Three independently pocketed bodies produce three pockets and no cross-body pocket pattern.
- [ ] A real three-member slot/pocket pattern within one body remains grouped.
- [ ] Reordering an assembly's solid traversal does not change feature or pattern identity.
- [ ] Automatic and declared paths preserve their existing recognition ownership: one aggregate orchestration, zero recognition on declared builds.
- [ ] No numeric-position heuristic silently treats assembly spacing as manufacturing pattern intent.
- [ ] Mutations that discard body ownership make both physical counterexamples fail.

Related: #958, ADR 0013, ADR 0017.

pzfreo:
Sequenced as the next implementation under #1113: Project 3 is Ready / Next / W2. Delivery remains bounded to the cross-body slot/pocket pattern trust defect and its rejection mutations.

[...]

Issue #1071, opened 2026-08-07
Annotation obstacle performance guard flakes under the documented full preflight


pzfreo:
## Outcome

Make the annotation-sized empty-rectangle regression a stable complexity/performance guard under the repository's documented `scripts/pr-check --full` command.

## Evidence

On v0.4.1 release commit `3edc949`, all hosted lanes passed, but the local full preflight failed:

```
tests/test_make_drawing.py::TestIsoEmptyRect::
  test_annotation_sized_obstacle_set_stays_bounded
Timeout (>10.0s)
```

The test passed when rerun alone **with coverage**, taking 8.37 s. Under the documented four-worker, coverage-instrumented full preflight it crossed the 10 s wall-clock mark while other CAD tests competed for CPU.

This is the regression guard added by #1065/#1066 to prevent the old roughly 34-billion-rectangle path. Its subject is important; its current wall-clock oracle is too close to normal local runtime to distinguish an algorithmic regression from runner contention.

## Acceptance

- [ ] Keep a load-bearing guard that fails if the old unbounded candidate search returns.
- [ ] Measure an invariant tied to work/complexity, or isolate the benchmark from xdist contention; do not merely delete the timeout.
- [ ] Prove the guard passes repeatedly under `scripts/pr-check --full` on supported local and hosted environments.
- [ ] Mutate back toward the pre-#1066 search and show the replacement guard fails.
- [ ] Document any explicit performance budget with baseline and runner assumptions.

## Release impact

Not a v0.4.1 product blocker: the exact test passes in isolation, all hosted coverage/matrix checks pass, and the release diff is metadata only. It is a reliability defect in the local gate itself.

pzfreo:
Reproduced again during #958 full preflight on 2026-08-07:

- `scripts/pr-check` (four xdist workers with coverage): only failure was `TestIsoEmptyRect.test_annotation_sized_obstacle_set_stays_bounded`, timing out at 10 s; the other result was `2927 passed, 1 skipped, 2 xfailed`, 93.66% coverage.
- The exact node alone **with coverage** also timed out at 10 s, so this instance is not xdist contention.
- The exact node alone with `--no-cov` passed in 7.25 s pytest time (9.06 s wall).

[...]

Issue #917, opened 2026-07-28
U-channel: two profile transitions unrecognised, and the wall geometry is dimensioned by repetition


pzfreo:
## Status: partially reproduces; the fixture as filed was wrong (verified 2026-08-03 on `main` @ `20b9205`)

The synthetic reproduction in the original report does not build the described part.
build123d's `Box`/`Cylinder` default to `Align.CENTER`, and the snippet was written as if
they were min-corner aligned, so two of the three cutting operations remove nothing:

- the counterbore `Pos(0, 0, 8) * Cylinder(6, 4)` spans z 6 → 10, which lies in the empty
  gap *between* the walls — it cuts no material;
- the cross-hole at `Pos(0, 0, 23)` sits above the part entirely (the walls, spanning
  z 3 → 21, are themselves misplaced: they float above the base, which spans z −6 → +6).

Recognition on that solid returned `{'hole': 1, 'envelope': 1, 'plate': 3}` — a single
plain bore. The "cross-hole's vertical location is absent" observation was therefore an
artefact: there was no cross-hole.

### Corrected fixture

```python
from build123d import Align, Box, Cylinder, Pos, Rot
from draftwright import build_drawing

# 50 x 50 x 30 U-channel: 12 mm base, two 12.5 mm walls standing on it.
part = (
    Box(50, 50, 12)
    + Pos(0, -18.75, 15) * Box(50, 12.5, 18)
    + Pos(0, 18.75, 15) * Box(50, 12.5, 18)
)
part -= Cylinder(2, 12)                # through bore in the base
part -= Pos(0, 0, 4) * Cylinder(6, 4)  # counterbore, opening at the base's top face
part -= Pos(0, 0, 15) * Rot(90, 0, 0) * Cylinder(
    4, 60, align=(Align.CENTER, Align.CENTER, Align.CENTER),
)                                       # transverse hole through both walls

dwg = build_drawing(part)
```

### What `main` actually produces

```text
kinds: {'hole': 2, 'envelope': 1, 'plate': 3, 'step_level': 1}

m_env_width        = 50      dim_height          = 30
m_env_depth        = 50      dim_plate_z0        = 12
dim_plate_y0       = 12.5    dim_plate_y1        = 12.5
m_locx0            = 25      m_locy0             = 25
dim_loc_front_x2500 = 25     dim_loc_front_z2100 = 21
hc_front0, hc_plan0, m_cm0, m_cm1, + section A-A furniture
```

[...]

PR #1075, opened 2026-08-07
Recognise milled features within physical solids

### LINKED ISSUE #958: Multi-solid recognition combines faces into phantom milled features

### LINKED ISSUE #1074: Dimension mirror corpus relies on mis-modelled pad and slot fixtures
## Problem

The dimension-mirror corpus currently claims coverage from two geometries that do not represent their names:

- `pad = Box(80, 60, 10) + Pos(0, 0, 10) * Box(30, 20, 4)` contains two solids separated by a 3 mm air gap because `Box` is centre-aligned. It only yields `pad` and `slot` records through the cross-solid false recognition tracked by #958.
- `slot = Box(80, 60, 12) - Pos(10, 0, 6) * Box(30, 8, 10)` is a blind recess and recognises as a pocket, not a through-slot.

The phantom slot from the first fixture made `test_corpus_kinds_own_an_approved_dimension` pass even though the fixture named `slot` did not exercise the slot route. Fixing #958 correctly makes the corpus gate fail and exposes the missing real slot coverage.

## Acceptance

- The pad fixture is a one-solid attached pad and claims only `pad`.
- The slot fixture is a through-cut and claims `slot`.
- The corpus recognition guard, approved-dimension ownership guard, and emitted-script round trip pass for both fixtures.
- Reverting either geometry to its old form makes the relevant guard fail.

Discovered during the adversarial review of #958.

pzfreo:
## Summary

- scan slots, pockets, channels, and rectangular pads independently within each physical solid
- preserve the public once-per-family recognition boundary and deterministic aggregate ordering
- correct dimension-mirror fixtures that relied on a detached-pad phantom to claim slot coverage
- preserve genuine features on every body while keeping cross-body derived pattern ownership in #1073

## Adversarial review

- aggregate-scan mutations for all four families are killed by independent physical counterexamples
- attached one-solid pad remains one pad, no slot, dimensioned once, and lint-clean
- genuine slots, pads, pockets, and channels on two separate bodies remain two records
- ADR 0013/0015/0017 review found no boundary change: records remain geometry-only and each public family is still called once

## Validation

- `scripts/pr-check --quick`
- 60 focused recognition/manifest/contract tests
- 460 broad slot, pocket, model, Sheet emit tests after correcting the latent corpus fixtures
- #958 regression file: 10 passed
- full fast tier: 2927 passed, 1 skipped, 2 xfailed; only failure was the known coverage-sensitive 10 s layout performance guard tracked in #1071
- changed production lines: 100% diff coverage

Closes #958
Closes #1074
Refs #1073
Refs #1071

### DIFF
--- a/src/draftwright/recognition/pads.py
+++ b/src/draftwright/recognition/pads.py
@@ -24,13 +24,8 @@ class RaisedPad(Record):
     z1: float
 
 
-def recognise_rectangular_pads(part, *, tol: float = 0.2) -> list[RaisedPad]:
-    """Return rectangular horizontal faces that are bounded in plan and raised.
-
-    A candidate is a planar +Z face whose area fills its XY bounding rectangle
-    and is bounded on both in-plane axes. Full-span steps are excluded;
-    non-rectangular pocket floors and perforated plate faces fail the area test.
-    """
+def _recognise_rectangular_pads_one(part, *, tol: float) -> list[RaisedPad]:
+    """Recognise pads using one solid's faces and bounds."""
     bb = part.bounding_box()
     raw_tops: list[tuple[float, float, float, float, float]] = []
     for face in part.faces():
@@ -136,3 +131,18 @@ def recognise_rectangular_pads(part, *, tol: float = 0.2) -> list[RaisedPad]:
             abs(other.z1 - pad.z0) <= tol and touches_plan(pad, other) for other in raw_regions
         )
     )
+
+
+def recognise_rectangular_pads(part, *, tol: float = 0.2) -> list[RaisedPad]:
+    """Return bounded rectangular raised faces independently per solid.
+
+    A candidate is a planar +Z face whose area fills its XY bounding rectangle
+    and is bounded on both in-plane axes. Full-span steps are excluded;
+    non-rectangular pocket floors and perforated plate faces fail the area test.
+    Body-local walls and bounds prevent a detached component from being treated
+    as a pad raised from another component (#958).
+    """
+    solids = list(part.solids())
+    sources = solids if len(solids) > 1 else [part]
+    pads = [pad for solid in sources for pad in _recognise_rectangular_pads_one(solid, tol=tol)]
+    return sorted(pads)
--- a/src/draftwright/recognition/slots.py
+++ b/src/draftwright/recognition/slots.py
@@ -774,15 +774,8 @@ def _recognise_obround_from_ends(part, faces, *, blind: bool = False):
     return out
 
 
-def recognise_slots(part) -> list[Slot]:
-    """Recognise enclosed through-slots with rectangular walls in *part*.
-
-    Returns a list of :class:`Slot`, one per physical feature, in a
-    deterministic order (co-located candidate pairs are merged, keeping the
-    narrower width).  See the module docstring for the recognition predicate and
-    its (deliberately narrow) scope. Obround slots too stubby for their flat walls
-    to pair are recovered from their end caps (#816).
-    """
+def _recognise_slots_one(part) -> list[Slot]:
+    """Recognise slots using one solid's faces and bounds."""
     faces = _planar_faces(part)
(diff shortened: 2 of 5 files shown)


pzfreo:
Adversarial review round 2 found and fixed one substantive issue before merge: channel ordering omitted depth, so two identical U-channels stacked on separate bodies changed record order when Compound children were reversed. `9115277` adds a geometry-complete channel tie-breaker and a reversal guard. Removing the depth terms makes that guard fail at index 0; restored code passes #958 + #917 (21 tests) and all static gates. Hosted CI has restarted on this head.

pzfreo:
Adversarial review round 3: clean pass; only nits remain (minor wrapper/fixture duplication, intentionally kept local to this slice). Clean-head validation: 69 targeted regression, established-family, ADR 0017 manifest/call-contract, and Sheet corpus tests passed; worktree clean and diff whitespace clean.

pzfreo:
Adversarial review round 4 / Codecov gate: `codecov/project` exposed that the old D-cutout guard stopped reaching `_has_side_walls` after per-solid scanning because its fixture was actually three solids. `922fae9` adds a thin exterior bridge, proves the fixture is one solid, and restores the predicate-level guard. Removing inward-normal checks now makes it fail with a phantom `Slot(width=8, length=28)`. This recovers the exact 2 line hits + 1 branch behind Codecov’s -0.01% delta without changing production behavior.

PR #1076, opened 2026-08-07
Reuse detected whole-part envelopes

### LINKED ISSUE #1000: Sheet.envelope() on a from_part sheet creates a second envelope feature, invisible on the drawing
Found while building the #996 suppression ledger — and found *by* it, which is the point worth noting: the drawing hides this entirely.

## What happens

`Sheet.envelope()` appends unconditionally (`sheet.py:1158-1161`):

```python
def envelope(self, obj=None) -> _Params:
    """Declare the overall bounding dimensions. Defaults to the whole part."""
    self._features.append(_envelope(obj if obj is not None else self._part))
```

`Sheet.from_part()` already seeds a **detected** envelope. So the documented-looking sequence

```python
s = Sheet.from_part(part, title="T", number="N").auto_dimensions()
s.envelope()          # a second envelope feature
```

produces a model with **two** `EnvelopeFeature`s.

## Why it is invisible on the sheet

`render_envelope` takes `envs[0]`, so the drawing is identical either way:

```
from_part alone       : ['m_env_depth', 'm_env_width']
from_part + envelope(): ['m_env_depth', 'm_env_width']
envelopes in model    : 2
lint: clean
```

No duplicate dimensions, no lint. Nothing surfaces it.

## Where it does show

The audit ledger (#996). The second envelope's parameters are all reported as omitted, so `Drawing.suppressions()` carries rows for a feature the author never knowingly created:

```
height.length | envelope@(0.000,0.000,0.000)/z[...] | not in the authored dimension set
depth.length  | envelope@(0.000,0.000,0.000)/z[...] | not in the authored dimension set
height.length | envelope@(0.000,0.000,0.000)/z[...] | not in the authored dimension set   <- the phantom
depth.length  | envelope@(0.000,0.000,0.000)/z[...] | not in the authored dimension set   <- the phantom
```

Worse under an authored set: `s.dimension(s.envelope(), "width.length")` names **one** envelope, so the other's `width.length` is reported as *"not in the authored dimension set"* — the author did name that measurement, and the ledger says they did not.

## Suggested direction

`envelope()` reads as *"dimension the overall size"*, and a part has one overall size. Returning a handle to the existing envelope when the model already carries one — rather than appending a second — matches what the verb means and keeps the model honest. If a caller genuinely wants a *sub*-envelope of some object, that is the `envelope(obj)` form and is a different feature.

Whatever the fix, the two-envelope model should not be reachable by a documented sequence.

## Note

The `feature_key` in #996 renders both identically (`kind@origin/axis[scalars]`), because they have identical geometry — its documented limit. That is not the bug here; two features that are genuinely the same measurement *should* key alike. The bug is that there are two.

Refs #996, #997.

pzfreo:
Closes #1000

## What changed
- make no-argument `Sheet.envelope()` reuse an equal whole-part envelope already seeded by `Sheet.from_part()`
- preserve token identity so the returned handle decorates and dimensions the detected feature
- keep explicit `envelope(obj)` declarations distinct, including equal geometry
- add suppression-ledger and adversarial regressions for unconditional append, kind-only matching, and over-eager explicit deduplication

## Validation
- `uv run pytest -q tests/test_issue_1000_envelope_reuse.py` (6 passed)
- `uv run pytest -q tests/test_declare.py tests/test_sheet_identity_invariant.py tests/test_suppression_ledger.py` (334 passed)
- `scripts/pr-check --quick` (73 passed; static gates clean)

## Review evidence
- unconditional-append mutation fails the cardinality and suppression tests
- kind-only matching mutation binds the handle to a reordered sub-envelope and fails
- removing the no-argument gate deduplicates explicit equal geometry and fails

### DIFF
--- a/src/draftwright/sheet.py
+++ b/src/draftwright/sheet.py
@@ -1176,8 +1176,19 @@ class Sheet:
         return _Params(self, len(self._features) - 1)
 
     def envelope(self, obj=None) -> _Params:
-        """Declare the overall bounding dimensions. Defaults to the whole part."""
-        self._features.append(_envelope(obj if obj is not None else self._part))
+        """Declare the overall bounding dimensions. Defaults to the whole part.
+
+        The no-argument form reuses an equal whole-part envelope already seeded by
+        :meth:`from_part`; its returned handle therefore decorates and dimensions that
+        detected feature instead of appending a duplicate. An explicit *obj* remains a
+        distinct declaration, including when another envelope is already present.
+        """
+        feature = _envelope(obj if obj is not None else self._part)
+        if obj is None:
+            for index, existing in enumerate(self._features):
+                if existing == feature:
+                    return _Params(self, index)
+        self._features.append(feature)
         return _Params(self, len(self._features) - 1)
 
     # -- GD&T / finish aspects (ADR 0011 P2c, #479) ---------------------------
--- a/tests/test_declare.py
+++ b/tests/test_declare.py
@@ -2012,6 +2012,9 @@ def test_the_declared_envelope_equals_the_detected_one():
             f for f in build_part_model(_solids_body(obj)).features if f.kind == "envelope"
         )
         declared = declare_envelope(obj)
+        assert detected == declared, (
+            f"{name}: exact equality is required for Sheet.envelope() to reuse detection"
+        )
         # EVERY field, not just the two that were wrong. The name says equal, so it compares
         # equal — `frame.axis`, `bbox_min` and `bbox_max` went unchecked in the first cut, and a
         # property test that inspects a subset is the shape both earlier bugs hid in.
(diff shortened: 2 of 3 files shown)


pzfreo:
Adversarial review round 2: clean; only nits remain.

Checked against ADR 0011 hybrid declaration, ADR 0016 token/authored-omission semantics, and ADR 0017 declared-path recognition ownership. Focused coverage executes every changed line and branch. Mutations proved the guards reject unconditional append, kind-only reuse, and deduplication of explicit equal-object declarations. No further code change required.

pzfreo:
Adversarial review round 3 found and fixed one guard gap: the existing producer-parity test was approximate, but this implementation depends on exact `EnvelopeFeature` equality. It now asserts exact equality on the AP203 STEP fixture, a plain solid, and an off-origin solid; all pass. Field-level approximate assertions remain for useful diagnostics.

Issue #1077, opened 2026-08-07
audit: mixed featureless and feature-owned suppressions crash diff_builds


pzfreo:
Raised from the adversarial review of #1004.

`draftwright.audit.diff_builds()` creates suppression tuples as `(feature, parameter_id, reason)` and calls `sorted()` on the set difference. A real authored drawing can contain both:

- a featureless overall-height omission, whose public `feature` is `None`
- feature-owned step omissions, whose public `feature` is a string

When the differing rows include both, tuple sorting compares `None` with `str` and raises:

```
TypeError: '<' not supported between instances of 'str' and 'NoneType'
```

Minimal real path:

```python
part = Rot(0, 90, 0) * Cylinder(4, 20) + Pos(15, 0, 0) * Rot(0, 90, 0) * Cylinder(6, 10)
before = Sheet.from_part(part).build()
sheet = Sheet.from_part(part)
steps = [f for f in sheet.features if f.kind == "step"]
sheet.dimension(steps[0], "step.length")
after = sheet.build()
diff_builds(before, after)  # crashes
```

Expected: a deterministic diff whose ordering handles nullable feature keys and attributes the lost `m_steplen1` to the authored omission.

The fix should use an explicit total-order key rather than coercing the public ledger data or dropping featureless omissions. Add the real drawing regression above.

pzfreo:
## Iteration plan

Take this immediately after #1004 because it blocks the real automatic-versus-authored audit check that exposed it.

1. Add the real two-step drawing regression from the issue: automatic build versus `Sheet.from_part` authored to one `step.length`. Assert `diff_builds` does not raise, keeps the featureless height suppression row, reports the lost second step dimension, and attributes it to the authored omission.
2. Give suppression rows an explicit private total-order key used only for deterministic presentation. Preserve the public row values exactly, including `feature=None`; do not stringify the ledger itself or drop featureless omissions.
3. Apply the same key to both gained and lost suppression sorting so the inverse comparison cannot retain the crash.
4. Mutation gates: raw tuple sorting must fail on the mixed real ledger; filtering `None` rows must fail the retained-row assertion; coercing the public row must fail exact-value checks; reversing before/after must exercise `suppressions_lost`.
5. Run the audit suite, authored-dimension suite, quick gate, and strict placement goldens. This is an audit leaf fix and should not touch compiler, renderer, registry, or placement behavior.

Size remains S. It is Now/Ready in W3 and assigned; Project audit is green.

Issue #1006, opened 2026-08-03
audit: same-kind measurement replacement is invisible to the cross-build key


pzfreo:
## First AI-edit milestone — #1277

Pair this with #883 and the usable controls in #1466. A comparison must reject a same-count, same-kind substitution and preserve the named measurement's value, tolerance and datum/member/axis meaning outside the explicit edit. Begin with an identity contract for the supported declaration/edit workflow; report unresolved cross-build correspondence as unknown rather than certifying preservation.

The historical candidate of an ordinal within kind is not sufficient proof of ownership or durable cross-run correspondence. Do not reconstruct provider occurrence identity from order, coordinates or equal records. Persistent identity across arbitrary recognition reruns remains separate pzfreo/quiddity#485 feasibility work; the first edit milestone must state its narrower correspondence bound honestly.

Raised from #1002 review r5 (Codex).

`audit._correspondence` joins two builds on `(feature KIND, parameter_id)`, dropping the feature's origin and scalars. That is deliberate — the full ledger key embeds coordinates, so it cannot join two builds that differ, which is the only thing the module does (#1002 r1).

The cost: two features of the SAME kind are not separated. If slot A's width dimension is replaced by slot B's under the same annotation name, `measurements_substituted` sees an unchanged key and reports nothing. Multiplicity is now compared as a multiset (r5), so a *count* change is caught; a same-count, same-kind swap is not.

Documented in `src/draftwright/audit.py` under "What this cannot do, stated first", and in `test_a_replacement_within_one_kind_is_not_separated`.

Closing it needs a cross-build feature correspondence that survives perturbation — e.g. ordinal within kind, or a stable feature identity minted at detection time rather than derived from geometry. Related: #883 (per-member location identity).

PR #1078, opened 2026-08-07
Record turned step measurement identity

### LINKED ISSUE #1004: Turned step-length chain records no measurement identity
Split out of #1002 (Codex r3).

`_draw_step_chain` in `annotations/from_model.py` places step-length dimensions from bare `(p0, p1, value, tolerance)` segments. The `ApprovedDimension.id` is never plumbed into those tuples, and the segments are transformed through eight call sites (`fsegs`/`dsegs`/`hsegs`/collapsed runs) before reaching the placer.

Consequence: a lost turned step-length dim reads "nothing claims it" in `draftwright.audit`, and a same-name substitution is invisible.

Tracked as an explicit exemption in `tests/test_audit_differential.py::test_every_direct_placement_records_identity_or_says_why_not` — remove the entry when fixed (a stale entry fails the test).

pzfreo:
## Summary

- replace the turned step chain's positional tuples with a private structured segment that carries compiled measurement identities through projection and detail redraws
- record exact identities on ordinary dimensions, union all member identities on repeated-pitch marks, and attach affected identities to chain-drop diagnostics
- leave synthetic aggregate head/block spans deliberately unattributed, because they are not any one approved step measurement
- remove the direct-placement exemption and the duplicate prose inventory it made stale

## Why

`render_step_lengths` received `ApprovedDimension.id` but discarded it while converting the plan into bare geometry tuples. As a result the registry and `draftwright.audit` could not relate a placed or dropped step-length annotation to the compiled measurement it represented.

This remains renderer provenance only. It does not change recognition, compiler selection, labels, placement, or cross-build feature correspondence. Same-kind replacement remains #1006. The real omission review exposed the separate sorting crash in #1077.

## Adversarial review

- source-id removal is killed by X- and Z-turned exact-id tests
- retaining only the first id is killed for both the general uniform collapse and the Y-axis repeat collapse
- synthetic main-view blocks are asserted empty while their detail dimensions retain one exact id each
- existing strict placement goldens remain unchanged
- ADR 0010/0014/0015/0016 review found no new decision or alternate path

## Validation

- `scripts/pr-check --quick` (73 passed)
- 325 adjacent turned/render/audit/tolerance/declaration tests (324 passed, 1 skipped)
- `tests/test_refactor_golden.py` (14 passed)
- `mypy src/draftwright/`
- focused mutation checks for source, general collapse, and Y collapse identity loss

Closes #1004.
Refs #955, #1006, #1077.

### DIFF
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -19,6 +19,7 @@ once the holes epic landed (#251).
 from __future__ import annotations
 
 import math
+from dataclasses import dataclass, replace
 from typing import Any
 
 from build123d_drafting import DatumFeature, FeatureControlFrame, SurfaceFinish, TextBlock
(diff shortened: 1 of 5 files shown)


pzfreo:
## Adversarial review log

**Round 1:** found missing guards for the Y-specific repeat collapse and for truthful treatment of synthetic aggregate blocks. Added public regressions for both X/Y detail paths and the Y `4×` path.

**Round 2:** found that the proposed acceptance overclaimed cross-build substitution. #1006 remains the owner of same-kind correspondence; #1004 now proves exact compiled ids within a build. The real omission exercise also exposed the independent `diff_builds` sorting crash, recorded as #1077 and added to the Project.

**Round 3:** clean. Source-id removal, first-member-only general collapse, and first-member-only Y collapse mutations all fail. Strict placement goldens are unchanged. ADR 0010/0014/0015/0016 boundaries remain intact: this consumes approved ids, changes no content or placement, and creates no alternate inference path.

Only nits remain: `pa`/`pb` are compact private field names, and the structured-record conversion makes this provenance-only diff larger than the behavioral change. Both are preferable to retaining the positional tuple convention that caused the loss.

pzfreo:
## CI review

The first matrix run failed uniformly on the shrink-only private-test-import guard: `tests/test_render_seam.py` imported `_StepChainSegment` directly. No product test failed.

Fixed in 67199a1 by using a duck-typed `SimpleNamespace` fixture through the test's already-sanctioned `_draw_step_chain` seam. I did not widen the allowlist.

Local recheck: the private-import guard and the drop diagnostic test both pass.

PR #1079, opened 2026-08-07
Order nullable audit suppression rows

### LINKED ISSUE #1077: audit: mixed featureless and feature-owned suppressions crash diff_builds

pzfreo:
## Summary

- order suppression ledger deltas with an explicit nullable-aware key
- preserve the public tuple values, including featureless `feature=None` omissions
- cover both gained and lost suppression directions with a real authored two-step drawing

## Why

`diff_builds()` sorted raw `(feature, parameter_id, reason)` tuples. An authored drawing can omit both the featureless overall height and feature-owned step measurements, so Python attempted to compare `None` with `str` and crashed before reporting the dimension loss.

The key is presentation-only. Set comparison, public row values, correspondence, and attribution behavior are unchanged.

## Adversarial review

- restoring raw tuple sorting fails with the original `TypeError`
- filtering featureless rows fails the retained-row assertion
- applying the key only to gained suppressions fails the inverse comparison
- no compiler, renderer, registry, or placement code changes

## Validation

- `scripts/pr-check --quick` (73 passed)
- audit/authored-dimension/strict-golden suites (169 passed, 1 skipped)
- focused audit coverage 97.7%
- full mypy and Ruff
- three mutation checks above

Closes #1077.
Refs #1004, #955.

### DIFF
--- a/src/draftwright/audit.py
+++ b/src/draftwright/audit.py
@@ -101,6 +101,16 @@ def _rows(dwg) -> set[tuple]:
     return {(r["feature"], r["parameter_id"], r["reason"]) for r in dwg.suppressions()}
 
 
+def _suppression_sort_key(row: tuple) -> tuple:
+    """Total presentation order for public suppression rows, whose feature is nullable."""
+    feature, parameter, reason = row
+    return (
+        (feature is not None, feature or ""),
+        parameter,
+        (reason is not None, reason or ""),
+    )
+
+
 def _correspondence(feature, parameter) -> tuple:
     """The cross-build key: ``(feature KIND, parameter_id)``.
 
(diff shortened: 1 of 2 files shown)


pzfreo:
## Adversarial review log

**Round 1:** no scope finding. The fix changes only deterministic presentation order; it does not normalize, filter, or mutate ledger rows.

**Round 2 mutations:**
- raw tuple sorting reproduces the `NoneType`/`str` crash;
- filtering featureless rows fails the public-value assertion;
- keying only gained suppressions leaves the inverse comparison crashing.

**Round 3 published patch:** clean. The real fixture is intentionally retained despite being slower than a fake: it proves `Sheet.from_part` produces the mixed ledger and that #1004's step identity lets the lost dimension receive the authored-omission explanation.

Only nit remaining: the nested nullable sort key is more explicit than a string-coercion lambda, but that verbosity is what prevents public values from being silently changed.

PR #1080, opened 2026-08-07
Recover Z-turned height when step chain drops

### LINKED ISSUE #955: Z-turned overall height is suppressed unconditionally, so a dropped step chain leaves the height dimensioned by nothing

pzfreo:
## Summary

- compile a distinct overall-height contingency when an approved Z-turned step chain is the primary representation
- release that already-approved ladder only when the complete chain places zero marks, while preserving ordinary placement-drop behavior if the fallback also cannot fit
- keep authored omission authoritative, approve height directly when no step chain exists, and mirror the contingency into generated `Sheet` scripts
- remove the inactive suppression when the fallback is released and avoid duplicating the generic axial warning when the specific chain-drop diagnostic remains
- record the contingent-plan refinement in ADR 0016 and update ADR 0015's compiler shape

## Root cause

The compiler suppressed Z-turned overall height on the assumption that the step chain tiled the axial extent. The chain's transactional legibility decision happens later at placement, so a crowded chain could place zero marks after the only available overall height had already been withheld. Letting the renderer derive `60` from the bounding box would violate ADR 0016's compiled-plan boundary.

The compiler now owns both representations. Placement reports whether the primary survived, and the orchestrator selects the compiler-approved alternative without inventing dimensional content.

## Validation

- `scripts/pr-check --quick`
- focused compiler, suppression-ledger, renderer-seam, emitter parity, axial-coverage, and #955 regression suites
- detected and generated-declared paths both render `60`, clear `axial_length_missing`, retain `step_dim_dropped`, and expose no stale height suppression
- controls cover a surviving chain, authored height omission with both step lengths still approved, no approved chain, authored overall-only incompleteness, and an unplaceable fallback

## Adversarial review

The following mutations each failed independently before being reverted:

- release the fallback when any chain marks survive
- create the contingency despite authored height omission
- leave the inactive suppression after release
- retain the legacy envelope suppression when no approved chain exists

Closes #955

### DIFF
--- a/docs/adr/0015-part-drawing-compiler-as-built.md
+++ b/docs/adr/0015-part-drawing-compiler-as-built.md
@@ -41,6 +41,10 @@ The part-drawing engine is a **compiler**:
   dimensioning planner (model/planner.py)   │  plan_dimensions → DimensionGroup
   plan_locations / plan_sections            │  per feature; convention + view +
                                             ▼  model-level suppression + datum
+  compile_dimensions() → RenderableDimensionPlan
+        approved groups / ladders / locations / contingencies + omission diagnostics
+                                            │
+                                            ▼
   render-intents → the IR render layer (annotations/from_model.py, holes.py)
                                             │
   shared layout / projection / export  (ADR 0014 placement, ADR 0004 pack,
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -1309,3 +1309,30 @@ silently acquiring an obligation.
 Incidentally the softer label is the louder signal: `DeprecationWarning` is filtered by
 default in notebooks, several test runners and library-internal call paths, whereas a
 `UserWarning` shows unconditionally.
+
+## Amendment 5 — placement may release a compiler-approved contingency (2026-08-07)
+
+**A plan may carry an approved alternative that stays inactive unless the primary
+representation places no marks.** The alternative is dimensional content, not permission for
+a renderer to derive content after placement.
+
+#955 supplied the concrete case. A Z-turned part normally states its axial extent through the
+step-length chain, so the overall height is omitted. Whether that chain survives is known only
+at placement: a crowded chain can be dropped transactionally. The compiler cannot predict the
+drop, while the renderer must not recover the height from the bounding box. Treating either
+layer as the owner breaks Amendment 1's WHAT/WHERE split.
+
+`RenderableDimensionPlan.contingencies` therefore carries the compiler-built
+`ApprovedLadder` for the overall height, paired with the primary representation and its
+inactive diagnostic. The orchestrator observes only the primary renderer's placement result:
+when the complete chain places zero marks, it releases the already-approved ladder and removes
+the now-stale omission. `render_height_ladder` receives an ordinary approved entry and remains
+unable to read the model or bounding box. If the released height also cannot fit, that is an
+ordinary placement drop, not suppression.
+
+Three constraints keep this from becoming a conditional-suppression flag under another name:
+
+- an authored omission prevents the contingency from being compiled at all;
+- a surviving primary leaves the fallback inactive and the omission diagnostic intact;
+- contingencies are addressable, so a generated authored `Sheet` script carries the fallback
+  intent and makes the same runtime selection as the detected build.
--- a/src/draftwright/annotations/orchestrator.py
+++ b/src/draftwright/annotations/orchestrator.py
@@ -37,6 +37,7 @@ from draftwright.analysis import _sizing_bores
 from draftwright.annotations._common import PlacementContext
 from draftwright.annotations.balloons import render_balloons
 from draftwright.annotations.from_model import (
+    ladder_plan_for,
     render_boss_diameters,
     render_boss_heights,
     render_centermarks,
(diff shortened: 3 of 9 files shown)


pzfreo:
## Adversarial review record

Four rounds completed against the pushed commit.

1. **Compiled-plan boundary / round trip:** found the generated-script envelope's planner suppression was initially mistaken for authored intent. Fixed by giving authored omission explicit precedence while allowing the Z-turn convention to feed the contingency. Direct/generated annotation parity now passes.
2. **Test validity:** found the authored-omission control had also removed the primary chain, so it did not prove the intended gate. Rebuilt it with both step lengths approved and only height omitted; the mutation now fails.
3. **No-chain and diagnostics:** found a Z model with no approved chain could retain the old envelope suppression, and a generated script could retain a second differently-worded height omission after release. Fixed both and added controls.
4. **ADR/placement/provenance pass:** no blockers. `render_height_ladder` remains plan-only, the fallback enters the shared corridor solve, authored omission stays authoritative, and an off-page fallback reports an ordinary placement drop. Remaining observations are naming/style nits only.

Mutation gates independently reject release with surviving marks, release despite authored omission, stale suppression after release, and legacy suppression with no primary chain.

pzfreo:
CI review iteration 2

The Python 3.14 failures were deterministic oracle gaps, not runtime divergence:

- the fidelity corpus compared generated declarations directly with detector output even though the established mirror contract deliberately synthesises one envelope when a model-level height must become nameable; it now compares with `mirror_model()` and independently constrains the extra feature to exactly that envelope
- the grooved-shaft refactor golden still encoded #955's known missing-height defect; it now records only the deliberate `dim_height` / item-count change

Local gate after the correction: Ruff clean; `tests/test_sheet_emit.py`, `tests/test_refactor_golden.py`, and `tests/test_issue_955_height_contingency.py`: 374 passed, 2 expected xfails. The three CI failures also pass directly.

pzfreo:
Adversarial review iteration 3

Codecov's project gate exposed one permissive branch worth fixing rather than papering over: an unregistered contingency ladder kind was silently skipped by generated-script addressability. It now fails closed through the compiler's role registry, with a mutation-style regression test.

Also covered:
- a non-Z planner suppression still withholds overall height
- a model-level height the compiler withholds does not trigger a synthetic envelope
- removed an unreachable empty-shoulder branch (the helper only receives a constructed `TurnedProfile`)

Local gate: Ruff clean; issue + compiler boundary suite 65 passed.

Issue #676, opened 2026-07-17
Recogniser gap: hexagonal/polygonal-prism boss not recognised — across-flats size undimensioned (CTC-01)


pzfreo:
## Observed

CTC-01 has a central **hexagonal boss** (a raised hex prism with chamfered top edges). In the generated drawing it renders correctly in every view, but its **defining size — the across-flats (or across-corners) of the hex — is never dimensioned**. A machinist cannot make the boss from this sheet.

The feature dump (`dwg.model().features`) shows why — the boss is not recognised as a prism/boss at all:

```
hole 25.0 / hole 35.0 / hole 20.0
slot ×2
envelope
step_level ×1        <- the boss's raised TOP FACE (→ dim_step_0 "100")
chamfer ×3           <- the two C5 top-edge chamfers (+ the C50 corner)
fillet ×7            <- 7× R50
```

So draftwright captures only the boss's **top-face level** (as a `step_level`, giving the `100`/`150` heights) and its **chamfers** (`C5`). The **hexagonal prism itself** — its across-flats width and its 6-sided nature — is not a recognised feature, so nothing dimensions it. Its protrusion height is only *inferable* (`150` overall − `100` plate), not directly called out.

## Repro
```python
from draftwright import build_drawing
dwg = build_drawing("tests/fixtures/nist_ctc_01_asme1_ap242.stp", pmi="annotate")
print([f.kind for f in dwg.model().features])   # no 'boss'
# → the central hex boss appears only as step_level + chamfers; no across-flats dim on the sheet
```

## Desired
Recognise a **polygonal-prism boss** (hex being the common case) and dimension its **across-flats** (with the appropriate hex/AF designation), plus a direct **boss height**. Round out the machined-feature family with the prism-boss case.

## Relation to existing issues
Distinct from:
- **#631** — `Sheet.step(boss)` (the *declarative* path) drops the boss *height* dim; this is the *detect/recognition* path and the *across-flats* size.
- **#605** (closed) — across-flats on **round stock** (flats on a shaft); this is a standalone **prism boss**, not round-stock flats.
- **#84** — completeness *lint* blind to non-cylindrical *revolved* features; a hex prism is prismatic, and this is the recogniser, not the lint.

Fits the machined-feature recognition family (epic #148: flats/grooves/slots/pockets) as the un-covered prism-boss case. Found reviewing the CTC-01 render (2026-07-17).

pzfreo:
## Delivery plan (next after #955)

[...]

PR #1081, opened 2026-08-07
Recognise and dimension hexagonal bosses

pzfreo:
## Summary
- recognise attached regular hexagonal bosses from closed, outward, opposed side-plane evidence
- lower them through the shared recognition result into a dedicated IR feature with A/F and direct-height measurements
- render planner-approved `HEX … A/F` leaders and boss heights through the shared placement stages
- preserve authored suppression and generated-script feature/annotation fidelity

## Evidence
- CTC-01 AP203: one boss, `HEX 100 A/F`, direct height `50`, no circular-boss/pad duplicate, no new drop/overlap lint
- adversarial fixtures reject recesses, detached/standalone solids, irregular or inconsistent six-sided rings, shallow projections, and unproved polygon counts
- rotated geometry preserves geometric A/F; a local face notch cannot move the support-plane-derived axis
- removing the placed height produces live `boss_height_missing` coverage lint
- malformed anchors and unsupported declared axes fail closed without guessing or crashing

## Scope
This first #676 slice deliberately limits automatic inference to attached positive-Z hex bosses. Fluent `Sheet` vocabulary, other polygon counts, and orientation breadth remain in the recorded second slice.

## Validation
- `scripts/pr-check --full` (2,976 passed, 4 skipped, 2 xfailed; 96% line diff coverage)
- `uv run pytest tests/test_issue_676_polygonal_boss.py -m slow -q` (2 passed, CTC-01 real-part evidence)

Part of #676


pzfreo:
## Adversarial review record

### Round 1 — scope and pipeline
- **Fixed:** automatic recognition generalized to every even polygon without corpus evidence. It is now hex-only; octagons are explicitly rejected until slice 2 proves them.
- **Passed:** one `RecognitionResult` owner, one record-to-IR converter, planner-fed labels, shared placement, authored suppression, live height coverage, and real generated-script feature/annotation parity.

### Round 2 — geometry mutations
- **Fixed:** averaging side-face centroids let a local notch shift the boss axis. The axis now comes from a least-squares intersection of opposed support midplanes, with the notch mutation pinned.
- **Passed:** rotated A/F, irregular ring, recess, standalone prism, detached component, circular-boss/pad non-duplication, and CTC-01 transition-band evidence.

### Round 3 — repository gates
- **Fixed:** the first full preflight was behavior-green but 1 line below the 93% changed-line coverage threshold. The missing IR datum-reference contract is now asserted.
- **Passed locally:** static checks, 2,969 fast tests, 93% diff coverage, and 2 real CTC-01 slow tests.

No remaining non-nit findings. Explicit slice-2 scope: fluent declaration, other regular polygon counts, and orientation breadth.

pzfreo:
## Adversarial review round 4 — external coverage gate

Codecov counts changed branch destinations in addition to the repository's line-only `diff-cover` check. Its first report was 86.58% patch / 91.85% project even though all platform tests passed.

**Fixed:**
- added geometric mutations for sub-tolerance projections, near-regular but non-opposed normals, unequal opposed widths, and nonconcurrent midplanes;
- pinned degenerate-normal and malformed-anchor fail-closed behavior;
- removed a determinant guard made unreachable by the preceding six-normal angular-ring proof;
- an unsupported-axis mutation exposed a real `render_boss_heights` `KeyError`; the shared renderer now declines unsupported axes consistently instead of guessing or crashing.

**Revalidated locally:** static checks, both CTC slow tests, 2,976 fast tests, 93.73% total coverage, and 96% changed-line coverage. No remaining non-nit finding.

PR #1085, opened 2026-08-07
Prove repeating radial profile evidence

### LINKED ISSUE #1062: Define evidence-safe recognition for gears and repeating radial profiles
## Context

The `wheel_rh.step` case-study fixture has 13 equal, evenly spaced arcs sharing one circle on its outer boundary. Issue #1061 deliberately does not call the whole profile periodic or label it as a gear: the STEP geometry proves that radial arc pattern, but the current evidence does not establish correspondence between every intervening boundary segment, module, pressure angle, tooth standard, helix, backlash, or manufacturing intent.

## Problem

Draftwright currently reports the boundary as `unrecognised_defining_geometry`. We need a separate design/implementation slice for useful radial-profile recognition without manufacturing claims the source cannot support.

## Questions to resolve

- Which facts are directly recoverable from B-rep geometry: common-circle arc count, axis/centre, candidate tip/root diameters, profile depth, repeat phase, helix evidence?
- What correspondence proves that the complete sector geometry repeats, rather than only one subset of its arcs?
- What further evidence is sufficient to call a profile a gear rather than a generic spline/serration/impeller/toothed wheel?
- Which standard parameters may be derived, and with what tolerance/confidence and ambiguity diagnostics?
- What IR and callout vocabulary represents geometric facts separately from declared manufacturing semantics?
- How should a caller declare known gear metadata when geometry alone is insufficient?
- What completeness lint remains when cyclic geometry is recognized but tooth semantics are unresolved?

## Acceptance direction

1. Start with the geometry-only common-circle arc inventory already evidenced by the real fixture, plus synthetic rejection mutations.
2. Require full intervening-segment correspondence before introducing a `repeating_radial_profile` inventory.
3. Add gear semantics only where an explicit standard fit is unique and mutation-tested, or where the caller declares them.
4. Never infer a standard merely from tooth count and outside diameter.
5. Preserve the current #1061 warning until an end-to-end drawing carries enough truthful information to replace it.

Related: #1058, #1061.

pzfreo:
## Summary

- prove complete 13-fold outer-wire correspondence on the pinned wheel STEP fixture
- mutation-gate arcs-only, altered-spline, open-cycle, unequal-pitch, start-edge, and traversal cases
- record the evidence boundary: repeating radial geometry is proven; gear semantics and parameters are not

## Product impact

This is a bounded discovery decision. It leaves runtime recognition and `unrecognised_defining_geometry` unchanged, and defines separate follow-ups for reusable geometric correspondence and explicit declared gear requirements.

## Validation

- `scripts/pr-check --full`
- 3,028 passed, 4 skipped, 2 xfailed
- 93.75% combined line and branch coverage
- changed-source coverage: no production source lines changed

Closes #1062

### DIFF
--- a/docs/research/1062-repeating-radial-profile-evidence.md
+++ b/docs/research/1062-repeating-radial-profile-evidence.md
@@ -1,0 +1,103 @@
+# Repeating Radial Profile Evidence (#1062)
+
+> **Status:** discovery decision, 2026-08-07. The executable evidence is in
+> `tests/test_issue_1058_wheel_profile.py`. No automatic gear semantics are approved by
+> this note.
+
+## Decision
+
+The wheel fixture proves a **closed 13-fold repeating radial outer profile** about its Z
+axis. It does not prove that the profile is a gear, nor does it establish module, pressure
+angle, helix angle, backlash, tooth thickness, quality grade, or manufacturing process.
+
+Keep `unrecognised_defining_geometry` for this part. Do not add an automatic `GearFeature`,
+emit a gear callout, or treat outside/root diameters and repeat count as a complete
+manufacturing definition. The useful next boundary is:
+
+1. reusable geometry-only correspondence for a repeating radial profile; and
+2. a separately designed, explicit declaration of gear requirements and their drawing
+   presentation.
+
+The second may consume the first to check that declared tooth count and axis correspond to
+the solid. Recognition must not manufacture the declaration.
+
+## Pinned Evidence
+
+| Item | Observed fact |
+|---|---|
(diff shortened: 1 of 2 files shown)


pzfreo:
Adversarial review pass 2: clean. The real fixture pins both Z end faces, all 52 boundary edges, four 13-edge rotational orbits, and traversal/start invariance. Mutations reject arcs-only evidence, one altered intervening spline, an open cycle, and a closed unequal-pitch cycle. The note explicitly limits the claim to sampled geometry-only correspondence and preserves the completeness warning; no automatic gear semantics or unused production record is introduced. Residual nit: the proof helper is intentionally test-only until a production consumer exists.

Issue #623, opened 2026-07-14
PMI completeness lint: reconcile AP242 source records through extraction, lowering, and rendering


pzfreo:
## Summary

There is no lint that checks whether source AP242 PMI survives into the generated drawing. A file can contain authored dimensions, datum symbols, and GD&T; draftwright can extract only part of them, lower unsupported records to raw fallbacks, render only a subset, and still report Drawing.lint() as clean.

This makes --pmi annotate overstate its result and prevents users from distinguishing a complete manufacturing drawing from a partial transcription.

Related to #62, but independently valuable: #62 adds GD&T/datum lowering and rendering, while this issue makes every remaining gap explicit and measurable before and after that work lands.

## CTC01 evidence

Fixture:

tests/fixtures/nist_ctc_01_asme1_ap242.stp

The current extractor reports:

    PMI extracted ... 12/21 dims, 6/6 gtols

But the generated drawing omits official PMI including datum feature symbols and feature-control frames such as the profile tolerance 1.25 | A | B | C.

The six GD&T records currently preserve little more than characteristic type: values are zero, references/anchors are absent, they remain raw PmiFeature fallbacks, and render_pmi deliberately ignores them.

Despite that, tests/test_pmi.py::test_pmi_annotate_drawing_lint_clean asserts there are no lint errors. Existing feature, location, axial, and structural coverage lints do not compare source PMI with rendered annotations.

## Required reconciliation stages

Track stable source-record identity and outcome through:

1. source AP242 inventory discovered
2. successfully extracted
3. lowered to typed drafting IR
4. accepted by the placement/render path
5. actually rendered, or explicitly dropped

A count alone is insufficient; diagnostics must identify the record and failure stage.

## Proposed lint codes

- pmi_not_extracted — source record was discovered but extraction could not preserve it
- pmi_not_lowered — extracted record remains an unsupported raw fallback
- pmi_not_rendered — typed PMI reached the model but produced no annotation
- pmi_dropped — a renderable annotation was rejected by placement/capacity

Names may be adjusted to fit existing lint conventions, but the stages must remain distinguishable.

## Mode/severity semantics

- pmi=off: no completeness warning; omission is intentional
- pmi=report: report extraction/lowering coverage diagnostically, but absence from the drawing is intentional
- pmi=annotate: every source PMI record must either render or produce a visible lint issue
- Unsupported PMI must never disappear behind debug-only logging

[...]

PR #1088, opened 2026-08-07
Expose AP242 PMI extraction gaps

pzfreo:
## Summary

- make a frozen `PmiExtractionReport` the canonical result of the AP242 XCAF pass while preserving `extract_pmi()` as a list projection
- retain stable TDF source identities and explicit outcomes for every dimension, geometric tolerance, and datum label
- surface incomplete extraction as structured lint issues in `report` and `annotate` modes
- mutation-gate failed and partial dimension/GTOL conversion so records cannot disappear from the source denominator

## Why

The existing extractor returned only successful records. Unsupported datums, partial geometric tolerances, and caught conversion failures therefore became indistinguishable from PMI that never existed, allowing an incomplete AP242 drawing to appear lint-clean. This is the first source-to-extraction slice of #623; lowering and rendering reconciliation remain separate follow-ups.

## Compatibility

`extract_pmi()` still returns `list[PmiRecord]`, and `Analysis.pmi` remains a list projection. `PmiRecord` is now frozen and uses tuple reference points so the canonical report is transitively immutable; no repository caller mutates either surface.

## Validation

- `scripts/pr-check --full`
- 3,037 passed, 4 skipped, 2 xfailed
- 93.94% combined line/branch coverage
- 100% changed-source coverage (183/183 lines)

Refs #623

### DIFF
--- a/src/draftwright/__init__.py
+++ b/src/draftwright/__init__.py
@@ -29,8 +29,11 @@ _LAZY = {
     "FeatureInfo": "draftwright.drawing",
     "Sheet": "draftwright.sheet",
     "lint_feature_coverage": "draftwright.linting",
+    "PmiExtractionReport": "draftwright.pmi",
     "PmiRecord": "draftwright.pmi",
+    "PmiSourceEntity": "draftwright.pmi",
     "extract_pmi": "draftwright.pmi",
+    "extract_pmi_report": "draftwright.pmi",
     "choose_scale": "draftwright.compose",
     # A warning category users are told to filter must be importable without reaching into a
     # private module (#1043 review) — and without paying for the CAD kernel. It lives in the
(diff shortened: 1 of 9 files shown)


pzfreo:
## Adversarial review

Two passes complete.

- Pass 1 found one blocking fail-closed defect: an exception from XCAF `ReadFile()` escaped while `Transfer()` failures were structured. Fixed and mutation-gated before the full run.
- Pass 2 checked ownership, source-denominator completeness, stable identity, compatibility projections, lint severity/mode behavior, dependency boundaries, and ADR 0008/0015/0016/0017 constraints. No further substantive findings; no remaining review item above nit severity.
- Full local preflight: 3,037 passed, 4 skipped, 2 xfailed; 93.94% combined coverage; 100% changed-source coverage.

Merge remains gated on the complete hosted matrix.

PR #1090, opened 2026-08-07
Reconcile AP242 PMI render outcomes

pzfreo:
## Summary

- reconcile source-bearing authored PMI against ADR 0010's annotation registry
- report typed PMI that produces no annotation as `pmi_not_rendered`
- carry source IDs on `pmi_dropped` and distinguish solver rejection from no viable render candidate
- keep report/off modes free of per-record rendering requirements

## Evidence

CTC01's eight typed authored dimensions now reconcile to six registry-owned annotations and two structured placement drops.

The mutation pair proves the stage boundary:

- deleting one render dispatch produces `pmi_not_rendered` for that source ID
- forcing that same source through placement rejection produces `pmi_dropped`, not `pmi_not_rendered`

A source-less malformed authored dimension remains visibly diagnosed rather than disappearing after the false capacity label was removed.

## Validation

- `scripts/pr-check --full`: 3,042 passed, 4 skipped, 2 xfailed
- combined line/branch coverage: 93.98%
- changed source coverage: 100% (21/21)
- impacted regression set: 452 passed, 2 xfailed
- static analysis, formatting, mypy, and codespell: green

Part of #623. This PR deliberately leaves mode-level ignored-PMI diagnostics and summary stage counts to the fourth slice.

### DIFF
--- a/src/draftwright/annotations/_common.py
+++ b/src/draftwright/annotations/_common.py
@@ -1130,13 +1130,7 @@ class PlacementContext:
             ids = tuple(item for item in measurement if item is not None)
         else:
             ids = (measurement,)
-        source_ids: tuple[str, ...]
-        if source is None:
-            source_ids = ()
-        elif isinstance(source, (list, tuple)):
-            source_ids = tuple(str(item) for item in source if item)
-        else:
-            source_ids = (str(source),) if source else ()
+        source_ids = (str(source),) if source else ()
         self.registry.record_issue(
             LintIssue(
                 severity=severity,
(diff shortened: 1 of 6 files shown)


pzfreo:
## Adversarial review - clean

Reviewed against ADR 0010 (registry-owned provenance), ADR 0014 (shared placement solve), ADR 0015 (pipeline/lint boundaries), and ADR 0017's evidence-gated scope.

Substantive findings fixed during the loop:

1. No viable render candidate was incorrectly reported as a placement-capacity drop.
2. Removing that false label initially allowed malformed source-less authored IR to disappear; it now remains visibly diagnosed.
3. The missing-bore-bounds early return bypassed the generic no-candidate outcome; it now reports the source-specific failure.
4. A test widened private `Drawing._analysis` access beyond the architecture guard; the test was refactored and the allowance was not weakened.
5. `record_issue` accepted an unnecessary source collection shape; it is narrowed to the one-source AP242 contract.

The mutation pair independently proves dispatch loss versus solver rejection. Full local validation is green with 100% changed-source coverage. No substantive review findings remain; only nits remain.

PR #1092, opened 2026-08-07
Preserve AP242 range tolerances

### LINKED ISSUE #675: AP242 PMI: range-encoded diameter tolerances render as bare nominal callouts
## Observed

Building CTC-01 with `pmi="annotate"` renders the `4× ⌀35` corner holes as **four separate PMI leaders**, with the tolerance fragmented across them:

| annotation | view | label |
|---|---|---|
| `pmi_d_1` | plan | `ø35 -0.2` |
| `pmi_d_2` | plan | `ø35 +0.2` |
| `pmi_d_5` | plan | `ø35` |
| `pmi_d_6` | plan | `ø35` |

On the sheet this reads as an **ambiguous, non-releasable** callout: the `+0.2` sits on one hole's leader, the `−0.2` on another's, and two holes carry a bare `ø35` — a reader cannot reconstruct that the spec is a single `ø35 +0.2/−0.2` applied to all four holes. (The same fixture also renders `ø20 +0.1/-0.1` correctly as one bilateral callout — `pmi_d_3`/`pmi_d_4` — so the split is specific to how these `⌀35` records lowered.)

## Repro
```python
dwg = build_drawing("tests/fixtures/nist_ctc_01_asme1_ap242.stp",
                    title="CTC-01", number="NIST-01", pmi="annotate")
for n in dwg.annotations():
    o = dwg.get_annotation(n)
    if "35" in (getattr(o, "label", "") or ""):
        print(dwg.view_of(n), n, repr(o.label))
```

## Root cause — two candidates to disambiguate
1. **Extraction split a single bilateral record** (`ø35 +0.2/−0.2`) into two half-deviation records — an `extract_pmi` / AP242-lowering pairing bug; or
2. The **source authors the tolerance per-hole** (genuinely 4 records), and the lowering renders each independently — i.e. `render_pmi` / the diameter lowering lacks (a) **upper/lower deviation pairing** for coincident-nominal records on the same feature, and (b) **identical-diameter grouping** into one `n× ⌀` callout.

Either way the *rendered* result is malformed. First determine which by dumping the raw `AuthoredDimension`/`PmiFeature` records for the ⌀35 features (nominal, upper, lower, ref feature).

## Desired behaviour
- Pair a `+x` and `−y` record sharing nominal + ref feature into one `ø35 +x/−y` (or `ø35 ±x` when symmetric).
- Group identical-diameter coincident-spec records into a single `n× ⌀` callout (matching how the auto hole-recogniser groups `4× ⌀35 THRU`).

## Relation to existing issues
Distinct from #623 (PMI *completeness* lint — does source PMI survive at all) and #62 (GD&T/datum lowering). Touches the same lowering path as #449 (`tolerance=` on Leader/HoleCallout) and #422 (lower remaining PMI to drafting concepts) but is specifically about **pairing + grouping** the diameter/tolerance records, not the suffix plumbing.

Found while reviewing the CTC-01 render (professor-persona drawing critique, 2026-07-17): the fragmented tolerance was the headline release-blocker on an otherwise clean sheet.

pzfreo:
Closes #675.

## What changed

- extract XCAF range bounds from `IsDimWithRange()` / `GetLowerBound()` / `GetUpperBound()`
- carry the bounds through `PmiRecord`, `AuthoredDimension`, the Sheet declaration surface, and generated-script round trips
- render a normalized limit label (`ø34.8 - ø35.2`) without pairing unrelated source requirements
- retain source identities and the existing PMI stage accounting

## Why

NIST CTC-01 contains two distinct diameter requirements encoded as ranges. Draftwright read only their nominal `GetValue()` and rendered both as bare `ø35`, silently losing the 34.8/35.2 limits. The source contains no semantic evidence that these should be grouped with the two separate one-sided diameter requirements.

## Validation so far

- mutation-first tests fail on the previous extraction, lowering, and script-emission paths
- focused extraction/lowering/declaration/emitter tests pass
- real CTC-01 annotate build renders both source IDs as `ø34.8 - ø35.2`
- CTC-01 accounting remains 38 source / 18 extracted / 8 lowered / 6 rendered / 2 dropped
- `ruff check`, `ruff format --check`, `mypy`, and `codespell` pass

The full local regression and adversarial review loop are still running; this PR remains draft until they are clean.

### DIFF
--- a/src/draftwright/model/declare.py
+++ b/src/draftwright/model/declare.py
@@ -1910,6 +1910,8 @@ def measured_dimension(
     axis: str | None = None,
     upper_tol: float | None = None,
     lower_tol: float | None = None,
+    lower_bound: float | None = None,
+    upper_bound: float | None = None,
     source: str = "sheet",
     source_kind: str | None = None,
     source_id: str = "",
@@ -1919,8 +1921,9 @@ def measured_dimension(
     callers can author one without the façade). Validates the kind against
     :data:`~draftwright.model.ir.AUTHORED_DIMENSION_KINDS`, needs ≥2 ``ref_pts``, and derives
     ``at`` (the ``ref_bbox`` centre, else the ``ref_pts`` centroid) when not given.
-    ``source_id`` preserves an external record identity; ordinary Sheet declarations leave it
-    blank."""
+    ``lower_bound`` and ``upper_bound`` carry a limit range and must be supplied together; they
+    are mutually exclusive with deviation tolerances. ``source_id`` preserves an external
+    record identity; ordinary Sheet declarations leave it blank."""
     _require_positive(value=value)
     dim_kind = str(kind).lower()
     if dim_kind not in AUTHORED_DIMENSION_KINDS:
@@ -1936,6 +1939,19 @@ def measured_dimension(
     if dom not in ("X", "Y", "Z"):
         if not (dom == "?" and dim_kind in ("diameter", "radius") and bbox is not None):
             raise ValueError("measured_dimension() dominant_axis must be X, Y, or Z")
+    if (lower_bound is None) != (upper_bound is None):
+        raise ValueError("measured_dimension() needs both lower_bound and upper_bound")
+    lower = None if lower_bound is None else float(lower_bound)
+    upper = None if upper_bound is None else float(upper_bound)
+    if lower is not None and upper is not None:
+        if lower > upper:
+            raise ValueError("measured_dimension() lower_bound must not exceed upper_bound")
+        if not lower <= float(value) <= upper:
+            raise ValueError("measured_dimension() value must lie within its range bounds")
+        if upper_tol is not None or lower_tol is not None:
+            raise ValueError(
+                "measured_dimension() cannot combine range bounds with deviation tolerances"
+            )
     if at is None:
         if bbox is not None:
             x0, y0, z0, x1, y1, z1 = bbox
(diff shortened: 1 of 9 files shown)


pzfreo:
## Adversarial review gate

Reviewed the published diff against the NIST CTC-01 semantic oracle and ADRs 0011, 0014, 0015, 0016, and 0017.

### Finding fixed before publication

- **Positional IR compatibility:** the first implementation inserted `lower_bound` / `upper_bound` before existing optional dataclass fields. That could silently reinterpret positional `PmiRecord` or `AuthoredDimension` construction. The fields are now appended, preserving the public IR record shape.

### Clean-pass checks

- no cross-source pairing or fabricated `4×` relationship; each source ID remains independent
- range semantics cross the existing source → record → typed IR → renderer path; no second PMI engine
- rendering stays in the existing collect-then-solve candidate path; no raw annotation placement
- generated Sheet scripts carry and re-declare both bounds, preserving ADR 0011/0016 fidelity
- source reconciliation remains 38/18/8/6/2 on CTC-01
- mutation gates fail if extraction, IR propagation, emission, or rendered-label use is removed
- positional compatibility and invalid public range declarations are covered

No actionable findings remain; only wording/style nits would be possible.

### Validation

- hosted: all 15 OS/Python jobs, lint, Codecov project, and Codecov patch are green
- local: static preflight green; 100% changed-source coverage (33/33 executable lines)
- full local result: 3,051 passed, 1 skipped, 2 xfailed, with only the known unrelated #1071 coverage-instrumented timeout; the exact guard passes in 2.01 s without coverage and 10.00 s with coverage

Issue #62, opened 2026-06-16
AP242 GD&T/datum PMI: extract values/references and lower to drafting IR


pzfreo:
## Problem

The PMI extractor (`pmi.py`) already reads geometric-tolerance and datum records out of a STEP file's PMI — they arrive as `PmiRecord`s with `kind` in `{"gtol", "datum", ...}`. But the annotation path drops them:

```
# make_drawing.py ~2616
if n_gtol:
    _log.debug("PMI annotate: %d gtol/datum record(s) not yet annotatable (Phase 4)", n_gtol)
```

So a part authored with real GD&T in CAD loses all of it when draftwright draws it — only linear/diameter/angular dims survive. This is "Phase 4" of the PMI work, explicitly deferred.

## Suggested direction

Once the manual GD&T placement API exists (#61), wire the extracted records into it:

- Map each `gtol` `PmiRecord` → `FeatureControlFrame` via its `kind` → `GDTCharacteristic`, tolerance value, and datum refs.
- Map each `datum` record → `DatumFeature` symbol at the referenced face/edge.
- Anchor using the record's reference points projected into the relevant view (the same projection the dimension PMI path already uses for `FX/FZ/...`).
- Gate behind the existing `pmi="annotate"` mode; `pmi="report"` should already list them.

## Acceptance

- A STEP file carrying a position/perpendicularity tolerance with datums A/B/C renders the feature control frame in the correct view, not just a debug log line.
- Datum feature symbols render for referenced datums.
- `pmi="off"` and `pmi="report"` behaviour unchanged.

## Notes

- Depends on #61 (placement API) landing first so this reuses one code path rather than hand-rolling frame geometry in the PMI branch.
- Anchoring/leader routing for GD&T pulled from PMI ref-points is the hard part; start with frames that attach to a single planar face or cylindrical axis.

pzfreo:
## Scope confirmed against the code (2026-07-06) — bigger than wiring

With the P2b/P2c GD&T stack now on main (#478/#480/#482), the **placement half is done**:
`render_gdt` places `ControlFrame`/`DatumRef` IR items, and `declare.gdt_target(ref, part) -> (view, side, site, axis)` derives the target from a feature/face. A PMI→`ControlFrame` converter would reuse both directly. The characteristic vocabulary already matches: `pmi._GTOL_TYPE` emits exactly the 14 ISO 1101 lowercase names `FeatureControlFrame` expects.

[...]

PR #1094, opened 2026-08-07
Preserve AP242 geometric tolerance relationships

pzfreo:
## What changed

- correct the OCCT geometric-tolerance enum mapping, which was misclassifying four of six CTC-01 requirements
- extract referenced shape measurements and ordered datum letters from XCAF
- preserve datum references through the raw PMI IR and generated Sheet-script round trip
- replace the stale “type only” diagnostic with field-specific partial extraction reasons

## Why

Issue #62 assumed the existing characteristic vocabulary was correct and that both values and relationships were missing. CTC-01 proves otherwise: XCAF preserves the shape and datum relationships, but the installed OCCT enum ordering differs from the hand-written table and OCCT drops all six tolerance magnitudes during translation.

This slice fixes the report-mode correctness defect without pretending the records are ready to render. All six remain explicitly `partially_extracted` until their source magnitudes can be recovered.

Closes no issue; advances #62.

## Validation

- mutation-first CTC-01 gates pin all six source IDs, characteristics, reference counts, and datum chains
- generated raw-PMI script is executed and structurally compared after round trip
- `scripts/pr-check --static` (Ruff, format, codespell, mypy)
- `pytest -q tests/test_pmi.py tests/test_sheet_emit.py -x`: 406 passed, 2 xfailed
- strengthened focused gates rerun after review: 2 passed

### DIFF
--- a/src/draftwright/model/detect.py
+++ b/src/draftwright/model/detect.py
@@ -261,6 +261,7 @@ def build_pmi_features(pmi, bbox) -> list[AuthoredDimension | PmiFeature]:
                 ref_bbox=r.ref_bbox,
                 ref_pts=tuple(r.ref_pts),
                 source_id=r.source_id,
+                datum_refs=r.datum_refs,
             )
         )
     return out
--- a/src/draftwright/model/ir.py
+++ b/src/draftwright/model/ir.py
@@ -1171,6 +1171,7 @@ class PmiFeature:
     ref_bbox: tuple[float, float, float, float, float, float] | None = None
     ref_pts: tuple[Point, ...] = ()
     source_id: str = ""
+    datum_refs: tuple[str, ...] = ()
     kind: ClassVar[str] = "pmi"
 
     def parameters(self) -> list[DimParameter]:
--- a/src/draftwright/pmi.py
+++ b/src/draftwright/pmi.py
@@ -39,6 +39,7 @@ try:
     from OCP.TDF import TDF_LabelSequence, TDF_Tool
     from OCP.TDocStd import TDocStd_Document
     from OCP.XCAFDoc import (
+        XCAFDoc_Datum,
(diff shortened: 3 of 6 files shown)


pzfreo:
## Adversarial review round 2

No blocking findings remain.

The first pass found that the extraction oracle did not prove `datum_refs` survived the record-to-IR adapter. The test now checks the exact CTC-01 datum mapping on the built model, and the raw-PMI round-trip specimen is a semantically consistent geometric tolerance.

Mutation results:

- replacing `datum_refs=r.datum_refs` with an empty tuple fails the built-model gate
- omitting `datum_refs` from `_raw_pmi_line` fails the executed structural round trip
- the pre-implementation enum and reference mutations failed the source-record gates

ADR check:

- ADR 0015: XCAF records remain below the shared PartModel waist and are adapted once
- ADR 0016: imported PMI remains materialized and generated scripts preserve it
- ADR 0017: source IDs and partial outcomes remain explicit; no incomplete tolerance is lowered or rendered
- ADR 0014: this patch adds no placement path or raw coordinates

Only follow-on nits remain for #62: bbox-derived hints are not canonical face identity, and OCCT `coaxiality` needs a lowering-vocabulary decision. Neither is used to claim completeness in this slice.

pzfreo:
## Adversarial review round 3

The canonical artifact exposed one remaining blocker: patch coverage was 84.13%, with seven untested fail-closed lines. Added direct behavior tests for:

- datum relationship lookup failure
- an unreadable datum object
- a datum with no letter
- an unknown OCCT geometric-tolerance type

All 41 PMI tests pass. A branch-aware focused run now covers 100% of changed `pmi.py` lines and reports 98% changed-line coverage across the PR (the sole missing focused line is exercised by the already-green full script-emission suite).

No further findings beyond the previously recorded lowering nits.

Issue #1095, opened 2026-08-07
Preserve AP242 geometric-tolerance modifiers through PMI lowering


pzfreo:
Parent: #62

## Evidence

The NIST CTC01 AP242 fixture contains geometric tolerance `Position surfacic profile.2` (XCAF source `geometric_tolerance:0:1:4:15`). OCCT reports:

- characteristic: profile of surface
- modifier sequence: `XCAFDimTolObjects_GeomToleranceModif_All_Around` (enum 15)
- the current `PmiRecord` / `PmiFeature` path preserves the characteristic, relationships and datum A, but drops this modifier

This was found while preparing the Part21 magnitude-recovery slice. That slice must leave the record `partially_extracted` even after recovering its 0.5 mm magnitude; a number alone is not a complete requirement.

## Acceptance

- inventory the supported OCCT geometric-tolerance modifier vocabulary and fail closed on unknown values
- preserve modifiers through XCAF → source record → IR fallback → generated Sheet script
- lower supported modifiers into the existing `ControlFrame` representation without a parallel PMI-shaped public API
- keep unsupported modifier combinations explicit and partial
- add a CTC01 oracle for `All_Around` and a mutation proving the guard fails if modifier propagation is removed

No raw placement changes; ADR 0014 remains authoritative.

pzfreo:
Implementation reconnaissance while #1096 runs CI:

- OCCT exposes 17 geometric-tolerance modifiers (enum 0–16); CTC01 exercises `All_Around` (15).
- The current `ControlFrame.modifier` is specifically a tolerance-cell material-condition modifier, so `All_Around` must not be overloaded into that field.
- The existing `build123d_drafting.Leader` already has structured `all_around=` / `all_over=` flags and renders the kink rings. No raw placement or parallel annotation API is needed.

Proposed smallest slice after #1096: add typed/purpose-specific scope flags through `PmiRecord -> PmiFeature -> ControlFrame`, lower complete geometric tolerances through the existing aspect IR, pass the flags to `Leader`, and keep every other OCCT modifier explicit/partial until its semantics have an owned representation.

pzfreo:
[...]

PR #1096, opened 2026-08-07
Recover AP242 geometric tolerance magnitudes

pzfreo:
## Summary

- add an isolated, structured Part21 adapter for AP242 geometric-tolerance facts that OCCT's XCAF transfer drops
- recover and normalize explicit SI length magnitudes only after unique semantic-name + tolerance-kind correspondence
- preserve both XCAF and Part21 provenance through raw PMI IR and executable sheet emission
- keep unsupported units, ambiguous correspondence, parser failures, and unpreserved requirement fields explicitly partial

## Evidence

On the CTC01 AP242 reference file, this recovers five geometric-tolerance magnitudes:

- two position tolerances at 0.75 mm
- surface profiles at 1.25 mm and 0.5 mm
- perpendicularity at 1.5 mm
- flatness at 0.2 mm

Five sources become complete. The 0.5 mm surface profile remains partial because XCAF reports an `All_Around` modifier that the IR does not yet preserve; that follow-up is tracked in #1095. This slice deliberately does not lower or render the recovered raw PMI records.

## Validation

- `tests/test_pmi.py`: 44 passed
- `tests/test_sheet_emit.py -x`: 367 passed, 2 expected failures
- parser and structural suite: 111 passed
- `scripts/pr-check --static`: green
- adversarial mutations proved duplicate matching, unit conversion, record-to-IR provenance, and executable emitter round-trip guards

Part of #62.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -126,6 +126,10 @@ entry. Keep `_LAYERS` and this section in step.
   detected as declared (#969); the DAG's bottom leaf (guarded
   by `test_geometry_is_a_leaf`) so the IR waist uses them without importing
   `_core`.
+- **`_pmi_part21.py`** — the rank-0 structured ISO 10303-21 adapter for AP242
+  geometric-tolerance facts that OCCT's XCAF transfer drops. It resolves explicit
+  Part21 references and SI length units, and requires unique semantic-name + kind
+  correspondence; it knows nothing about XCAF, drafting IR, or placement.
 - **`_warnings.py`** — the public warning categories (`SoftDeprecationWarning`), a
   dependency-free leaf. Separate from `_core` on purpose: `_core` imports build123d, so a
   category defined there costs the CAD kernel (~6 s) to reach, and the pytest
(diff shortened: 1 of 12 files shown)


pzfreo (commented):
Adversarial review pass: no actionable findings.

Checked:
- exact unique `(semantic_name, kind)` correspondence; duplicate and blank-name cases fail closed
- explicit Part21 reference traversal and SI metre-to-mm conversion; unsupported or malformed units remain partial
- parser failure isolation and preservation of XCAF source census
- Part21 identity propagation through record, raw IR, and executable sheet emission
- ADR 0015/0017 ownership: the Part21 parser is a rank-0 evidence adapter, while XCAF correspondence and provenance remain in the PMI orchestration
- scope: recovered records are not prematurely lowered or rendered

Remaining nits/tradeoffs are recorded: STEPutils is old but pinned and isolated; matching intentionally rejects ambiguity; conversion-based units are unsupported; `All_Around` remains partial under #1095. Four targeted mutations were killed by the relevant tests.

pzfreo (commented):
Adversarial review pass 2 addressed the sole review signal from pass 1.

Codecov showed the overlay's fail-closed decision paths were executed incompletely even though the patch threshold passed. Added a focused decision table proving that:
- a blank XCAF semantic name remains an explicit partial result
- no exact Part21 correspondence remains explicit
- a matched Part21 entity with an unusable magnitude preserves its entity identity and remains partial

A mutation that erased the semantic-name failure reason failed the new oracle exactly. Targeted tests and static checks are green. No remaining actionable findings; hosted matrix rerun is the final gate.

Issue #1130, opened 2026-08-11
Epic: requirement-driven view planning and editable sheet layout


pzfreo:
## Goal

Choose and expose a drawing's views, arrangement, standard scale, paper and fixed paper-space typography as one requirement-driven plan:

```text
recognised measurables + drawing requirements
                    ↓
          authored ViewConstraints + candidate ViewSpecs
                    ↓
 projection + complete view-block measurement
                    ↓
       page / scale / layout selection
                    ↓
       shared annotation placement solve
                    ↓
       explicit coverage and diagnostics
```

## Live execution

The umbrella is not an implementation task. Delivery is split into native child issues:

- [#1259](https://github.com/pzfreo/draftwright/issues/1259) — Phase 5.5 view-set-aware planning (**Now**).
- [#1260](https://github.com/pzfreo/draftwright/issues/1260) — typed `ViewConstraints` and Sheet verbs.
- [#1261](https://github.com/pzfreo/draftwright/issues/1261) — re-home displaced requirements.
- [#1262](https://github.com/pzfreo/draftwright/issues/1262) — automatic semantic view selection.

Project 3 is the source of live status; this body records scope and acceptance.

The automatic engine proposes a useful plan; the `Sheet` DSL and generated Python script let a user add, remove, replace, resize and arrange named **view blocks** without hand-placing feature annotations.

## Case study: thin X-axis worm planetary plate

Source supplied by a user: `worm_planetary_concept_Alimacznicy.step`.

Geometry bbox: **43 × 217 × 217 mm**. It is a thin, predominantly rotational X-axis component with two concentric hole patterns, several coaxial diameters, an axial stack, pockets and a central keyway.

Current automatic result:

| request | chosen/result | observed outcome |
|---|---|---|
| automatic | A1 landscape, 1:1 | best annotation retention, but large unused regions and an unbalanced four-view composition |
| `--page A2` | A2, 1:2 | still sparse; drops ø65/ø89 and half the axial-step definition |
| `--page A3` | A3, 1:5 | unnecessarily small geometry; loses ø65/ø89/ø189 and all axial-step lengths |
| A3 forced 1:2 | visually compact | fits visually but drops requirements, including diameter/axial definitions |
| A2 forced 1:1 | current fixed layout infeasible | one view 74 mm out of bounds; four hole callouts dropped |

[...]

PR #1131, opened 2026-08-11
docs: ADR 0018 requirement-driven view planning and editable layout

pzfreo:
## What changed

Adds proposed ADR 0018 and its architecture-index entry.

The revised decision defines:

- one shared `ViewSpec` / layout-constraint vocabulary and one planner;
- authored `ViewConstraints` as the pre-projection request;
- one immutable `ResolvedViewPlan` as the post-solve result, attached once through typed `BuildState` and exposed read-only by `Drawing`;
- requirement-aware view selection and removal;
- first-/third-angle projection convention as a hard layout input;
- joint one-sheet view/page/preferred-scale planning with fixed paper-space typography;
- deterministic terminal behavior: renderable infeasible plans return a diagnostic drawing with `plan_infeasible`, while wholly unrenderable candidate sets raise;
- relational layout and optional whole-view pins anchored at the projection origin, never feature-annotation coordinates;
- fail-closed resolved-script emission when a feature-targeted section cannot be named semantically;
- explicit automatic-semantic versus editable-resolved generated-script contracts;
- partial supersession of ADR 0004's fixed-topology assumption while retaining compose-then-pack.

## Why

The 43 × 217 × 217 mm worm-planetary case in #1130 chooses A1 at 1:1 under the fixed four-view topology. Smaller forced sheets look better but lose supported diameter and axial requirements. The engine needs to choose the semantic view set, standard scale, paper and arrangement together, while the generated Python script must let users edit the resolved view/layout intent without hand-placing annotations.

## Scope

Documentation only. Final method spelling is deliberately not frozen. Multi-sheet document planning is explicitly deferred.

## Validation

- `git diff --check`
- adversarial request/result, failure-mode, projection-convention, state-ownership and script-round-trip review
- architecture index/link review

Tracks #1130.


pzfreo (commented):
Adversarial re-review of `aa726e4`: the request/result and failure-mode fixes are sound, but I still see one blocking policy contradiction and one important lifecycle correction.

1. **Blocking — the new default objective reverses the epic’s scale/paper priority and can recreate the motivating failure.** Section 5 chooses the *smallest feasible sheet first*, then the largest scale on that sheet. Tracker #1130 phase 5 instead orders hard feasibility/legibility, then preferred standard scale, then economical paper. Under the ADR wording, an A3/A4 plan at a smaller but still “legible” scale necessarily beats A2 at 1:1, even though the case study calls the smaller geometry unnecessary and targets A2 at 1:1. Minimum projected spans do not resolve this unless they are deliberately tuned to make every smaller-sheet candidate infeasible, which would hide a policy choice inside a legibility threshold. Please choose and state the intended drafting policy consistently. My recommendation is: hard gates first; then prefer the largest appropriate/preferred standard scale; among candidates at that scale choose the smallest sheet; then redundancy/waste and conventional arrangement. If paper economy really is intended to dominate scale, update the tracker and case-study expectation and explain why.

2. **Important — the pipeline diagram still places a “resolved-plan snapshot” before candidate projection.** The normative text now correctly says `ResolvedViewPlan` exists only after projection and the exact solve, and resolved-script generation converts it into authored `ViewConstraints`. The diagram should therefore show only “authored ViewConstraints / automatic defaults” at that input boundary. Leaving a resolved snapshot there reintroduces the lifecycle ambiguity this revision otherwise fixed.

Apart from those points, the prior findings are addressed cleanly: typed request/result separation, one writer, projection convention, fail-closed script emission, pin semantics, bounded one-sheet scope, and explicit terminal behavior are now coherent.

Issue #1217, opened 2026-08-18
Epic: verify claimed representations against the drawing, once, for every coverage family


pzfreo:
Settles #1206. Successor in form to #1202 — same loop discipline, same evidence gates.

## The premise

Coverage checks answer "did this requirement reach the sheet?" by walking the registry and reading
**provenance riders** the annotations carry about themselves. That is a claim, not an observation.
An annotation asserting it carries hole 3's diameter is believed, and nothing checks that the
annotation renders that diameter — or that it renders anything at all.

The failure mode is not hypothetical. `covers_hole_representations_by_feature` is **read in two
places and populated by nothing**. That inverted the benchmark's `drawing_consumer` metric —
16/16 `unsupported` against a true 16/16 `supported` — and nobody noticed, because the only thing
that could have noticed was the metric itself.

@pzfreo's decision on #1206:

> 1. Keep the benchmark denominator independently authored.
> 2. Reuse `hole_requirement_outcomes` as the single semantic correspondence implementation.
> 3. Treat its ledger as a **pointer to the claimed representation, not final proof**.
> 4. Verify that the referenced callout, table row or leader actually exists in the built drawing
>    and carries the expected value.
> 5. Eventually verify export-visible content for especially important cases.

Scope: a **shared facility** over `(feature, parameter) → annotation`, not a hole-specific one.

## Why this is one epic and not two fixes

The consolidation (#1206's original ask) and the verification are the same piece of work in
opposite orders. Consolidating first makes the benchmark score the engine's own claims with no
second opinion — strictly worse than today's duplication, which at least reads the drawing twice.
**Verification must land before consolidation.** That ordering is the epic's main structural
constraint and the reason the slices are sequenced as they are.

## Three PRs

Originally cut into seven slices; consolidated because the fixed cost per PR — branch, 3x3 matrix,
body, review cycle — is real, and S0/S1 cannot be judged apart from S2. Splitting further buys
ceremony, not review quality.

| PR | Scope | Gate |
|---|---|---|
[...]

Issue #1206, opened 2026-08-17
Evaluation carries a second hole-correspondence implementation beside hole_requirement_outcomes


pzfreo:
Raised by adversarial review of #1205 (S1 of #1202). **Not a defect in that PR** — its
answers were cross-validated against the existing implementation and agree state-for-state
on every part built. This is about there being two.

## The duplication

`evaluation/step_analysis.py` gained `_hole_candidates` / `_table_represented` / `_consumed`
to answer "did this recognised hole's size reach the sheet?".

`linting/hole_coverage.py::hole_requirement_outcomes` already answers that, and reads
**both** representation ledgers, where the new code reads only the populated one.

The review's agreement sweep — dense/A3, dense/A4 with the table dropped, blind plate,
counterbored, side-drilled, coaxial, countersunk, bolt-circle, 4x2 grid — is really a
measurement of how closely the new copy tracks the existing one. Per CLAUDE.md
(consolidation epic #635, ADR 0007) that is the shape the project calls tech debt.

Consuming the existing ledger would also fix two known gaps for free:

* the turned-part false negative (a bore's diameter printed as a `Leader` with no
  `feature_of`, scored `unsupported` — see the module docstring's stated limits);
* the per-feature credit gap that needed a driven unit test rather than a natural one.

## The tension worth resolving first

There is a real argument *against* it, and it is the argument this whole epic rests on.

#1169's principle is that a completeness score derived from what the system reports can
validate itself. `hole_requirement_outcomes` is the engine's own answer to "did I place
this". Reading rendered annotation names is closer to "what is on the page"; reading a
placement ledger is closer to "what the engine believes it did".

Against that: #1205 already reads an engine ledger for the table route, because there is no
other way to know a table row carries a hole's diameter. So the question is not whether to
trust the engine at all, but where the line sits.

## What would settle it

State explicitly which parts of the observation may come from the engine and which must be
read off the rendered sheet, then have one implementation of the permitted part. If the
answer is "placement outcomes are engine-reported and that is acceptable because the
DENOMINATOR is independent", the duplication should go and `evaluation` should consume
`hole_requirement_outcomes`. If the answer is "the observation must be sheet-derived", then
the table-ledger read in #1205 is the thing to revisit, not the duplication.

Either way the current state — two implementations, one partially trusting the engine and
the other fully — is the least defensible of the three.

## Notes

`linting` is rank 2 and `evaluation` rank 7, so the import direction is legal.

[...]

PR #1223, opened 2026-08-18
The benchmark consumes the ledger, and follows its pointer (#1206)

pzfreo:
PR 2 of 3 for #1217. Settles **#1206**.

## What it removes

`evaluation` carried its own hole-correspondence implementation beside
`hole_requirement_outcomes`, which answers the same question. Two implementations of one
question drift — and this one was the more generous of the two: it matched recognised holes to
IR features by axis, diameter and position, then asked whether any annotation was **named**
`hc_*`.

Both halves are gone. `_hole_candidates`, `_table_represented`, `_consumed` and `_Candidate` are
deleted; provenance does not care what a renderer called its annotation.

**This does not close the turned-part false negative**, and an earlier version of this PR said it
did. Measured on `Cylinder(20,60) - Cylinder(6,70)`: `unsupported` before and after, with a
byte-identical annotation set. Name matching was never the mechanism — the bore's `ø12` is drawn
by `ldr_z0`, which carries no measurement claim at all, so the ledger reports `bore.diameter` as
`missing`. That is #754 debt, which this module's own docstring already stated correctly.

## The ledger is a pointer, not proof

Per the decision on #1206: `placed` means the engine *recorded* an annotation as carrying the
requirement. The consumer now follows that pointer through `linting/evidence` and confirms the
annotation renders the approved value. **A hole whose callout draws `⌀999` is not carried,
whatever the ledger says.**

The refutation is filtered to the **size** requirement, deliberately: a wrong *location* label is
a real defect that another check owns, and crediting it here would make this metric answer a
question it does not ask.

## Two enabling changes in `linting/`

- **`HoleRequirementOutcome` names its evidence** — the recognised hole sites it accounts for and
  the IR features it matched. The benchmark zips one observation per hole *positionally*, and a
  group outcome cannot be attributed from `source_at` and `member_count` alone.
- **`ClaimOutcome` carries its measurement**, so a consumer joins on the feature rather than only
  the parameter name.

## The measured divergence, in full

| fixture | before | after |
|---|---|---|
| `nist_ctc_03_asme1_ap203` | 15/15 supported | **10 supported, 5 unknown** |
| `nist_ctc_03_asme1_ap242` | 15/15 | **10 supported, 5 unknown** |
| `nist_ctc_04_asme1_ap203` | 54/54 | **46 supported, 8 unknown** |
| `nist_ctc_04_asme1_ap242` | 54/54 | **46 supported, 8 unknown** |
| CTC-01, CTC-02, CTC-05, `issue_915`, thumbwheel | — | unchanged |
| 5-case evaluation corpus | detection 1.0, fidelity 15/15, downstream 20/20, 5 conformant | **identical** |

**26 holes in total**, and every one of them is a hole the ledger reports `unverifiable` — it
knows they exist and declines to join them to a feature without guessing. The old code credited
them by guessing.

An earlier version of this table omitted CTC-03 entirely and called itself "the divergence in
full". It was half of it, from a fixture I had not looked at.

The eight are the holes the ledger reports `unverifiable` — it knows they exist and declines to
join them to a feature without guessing. The old code credited them by guessing. Crediting a hole
whose evidence cannot be located is the self-validation #1206 was opened to remove, so the lower
number is the more honest one.

## A false story I nearly shipped

An earlier cut derived each outcome's covered positions with `_members(source)` — but the source
object is `holes[0]`, the group's **representative**, so a group of eight registered *one*
position and the other seven scored `unknown`. On CTC-04 that manufactured 15 phantom
unaccounted-for holes, which I then explained with a plausible account of `_parameter_ids`
omitting `bore.diameter` and the ledger's denominator being narrower than the recognition set.

The parameter census I quoted was real. The conclusion was invented to fit my own defect. Sources
now carry their member sites explicitly, and the divergence is exactly the eight predicted before
implementing.

## The stubs go with the code they stood in for

`_StubDrawing` and `_StrippedDrawing` emptied `annotations_of`, which the consumer no longer
reads. A stub of an abandoned seam keeps passing while testing nothing — this file's own founding
lesson, from the #1205 review. Every surviving test drives a real build and removes a real
annotation. Tests that poked the deleted matcher's internals went with it; those guarantees are
`hole_coverage`'s now and are covered by `test_issue_1143_hole_completeness.py`.

## Verification

- **Twenty-four mutations**, each with the substitution asserted applied, **all killed** —
  including the representative-instead-of-members regression, the verifier refutation being
  ignored (which left *every* test in the file passing), and the size-parameter filter.
- Full fast tier **4,284 passed**, 5 skipped, 2 xfailed. `scripts/pr-check --static`: exit 0.
- Benchmark re-run after every change.

PR 3 adds a second feature family, whose gate is that it costs a producer and zero changes to the
facility.

---

## Review history

Two adversarial rounds. Round 1 found three blockers — a divergence table missing a whole
fixture, a claimed gate (the turned-part gap) that does not close and whose stated cause was
wrong, and a real defect in `canonical_hole_sites` where the consumer keyed a different axis
space from the ledger. Round 2 verified every fix by execution and found the code right; what it
found wrong was that the corrections lived only in the commit message, while the code comments,
the reference doc and this body still asserted the retracted claims — and that two of the
corrections had introduced new false statements of their own.

Both rounds' findings are fixed. The pattern is worth stating plainly: the implementation has
been stable since the first push; what kept failing review was the prose about it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,18 @@
 
 ### Changed
 
+- `HoleRequirementOutcome` gains `members` and `features`, `ClaimOutcome` gains `measurement`,
+  and `linting.hole_coverage.canonical_hole_sites` is public (#1217). A consumer attributing a
+  requirement outcome to a specific recognised hole needs the evidence the outcome accounts
+  for, and needs to key into the canonical space the ledger publishes it in — a through hole's
+  own-axis coordinate is zeroed there.
+- The STEP-analysis benchmark's `drawing_consumer` boundary now consumes
+  `hole_requirement_outcomes` instead of a second correspondence implementation of its own
+  (#1206), and follows the ledger's pointer through `linting.evidence` rather than trusting it.
+  Holes the ledger reports `unverifiable` now score `unknown` where the duplicate credited them
+  by guessing: 26 holes across the four CTC-02/03/04 fixtures. The five-case evaluation corpus
+  is unchanged.
+
 - `Drawing.suppressions()` rows gain a `conveyed_by` key: the dimension that states a
   withheld measurement instead, or `None` when nothing takes the fact over (#1154). It is not
   a synonym for the existing `authored` flag — that says whose decision it was, this says
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,8 +56,9 @@
   `hole_requirement_outcomes` instead of a second correspondence implementation of its own
   (#1206), and follows the ledger's pointer through `linting.evidence` rather than trusting it.
   Holes the ledger reports `unverifiable` now score `unknown` where the duplicate credited them
-  by guessing: 26 holes across the four CTC-02/03/04 fixtures. The five-case evaluation corpus
-  is unchanged.
+  by guessing: 26 holes, five on each CTC-03 fixture and eight on each CTC-04 fixture. The
+  five-case evaluation corpus is unchanged, and one lint message on CTC-03 now reports its
+  coordinates in the ledger's own canonical space.
 
 - `Drawing.suppressions()` rows gain a `conveyed_by` key: the dimension that states a
   withheld measurement instead, or `None` when nothing takes the fact over (#1154). It is not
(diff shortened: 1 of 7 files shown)


pzfreo:
Round-2 findings all addressed in a3beaf3; latent findings 13–15 filed as #1229, and the missed turned-part gate recorded on #1217 (with the real cause in #1227).

Full fast tier 4,284 passed / 5 skipped / 2 xfailed; `scripts/pr-check --static` exit 0. The two mutation gaps the review named are now killed (grouped-branch axis reverted → KILLED; `unresolved` and `no_expected_value` dropping their measurement → KILLED).

Issue #1215, opened 2026-08-18
A declared envelope tolerance is never rendered on any axis


pzfreo:
## Problem

`sheet.envelope().tolerance(0.05, on="height")` (or `width` / `depth`) records the decoration
on the model and it is **never printed**. Measured on `main`, all three axes:

```python
s = Sheet(Box(30, 20, 10)); s.envelope().tolerance(0.05, on="height"); s.auto_dimensions()
# labels: '30', '20', '10'   — no ± on any of them
```

Neither renderer appends a tolerance suffix for an overall extent:

- `annotations/from_model.py::render_envelope` builds its label without `_tol_suffix`
- `model/compiled.py::_compile_overall_height` builds the height rung with
  `rendered_label=_fmt(value)`

Both are unlike the feature-local renderers — `render_boss_heights` does
`pd.value_text + _tol_suffix(pd.tolerance, dwg.draft)` — so the same authored decoration is
honoured on a boss height and dropped on the part's overall thickness.

## Why it is filed rather than fixed in #1154

Found while reviewing #1154, which reconciles a feature-local extent against an overall one.
That slice's tolerance rule was written twice on the assumption that the receiving extent could
state the yielder's tolerance; measuring this is what showed it cannot. The rule was then
made not to depend on it — a toleranced measurement is simply never handed over — so #1154 is
correct either way and this is genuinely separate.

It is also not a small change: printing a tolerance on an overall extent alters the label of
every drawing whose author has decorated an envelope, and the overall height rides the ladder
rather than the ordinary dimension path, so the two renderers need the same answer.

## Expected

An authored envelope tolerance renders the same way a feature-local one does, on all three
extents, including the overall height when it comes from the ladder.

## Acceptance

- [ ] `envelope().tolerance(...)` renders on `width`, `depth` and `height`
- [ ] The overall-height ladder path and the ordinary envelope path agree
- [ ] A fit class (`FitClass`) renders as it does elsewhere, not only a ± pair
- [ ] Covered on the declared path (ADR 0011) and, if a decoration can reach it, the detected one
- [ ] Once this lands, #1154's tolerance rule can be revisited: with a receiving extent able to
      state a tolerance, transferring may beat refusing. The rule's comment says the refusal
      must not depend on this, so nothing breaks if it is left alone.

PR #1238, opened 2026-08-20
ADR 0019 (proposed): display-complete labels and a dimension-outcome ledger

pzfreo:
Proposes the superseding ADR that ends the per-site loop #1235 closed out.

**Why now:** #1215/#1216 took ten adversarial review rounds, and each round's fix created a new
instance of the same two defect classes. That replay — documented against the evidence in the
ADR's Context section — is the signature ADR 0016 Amendment 1 already names: *"the signature of
a missing boundary, not of missing tests."* The boundary is half-built: the compiler owns the
value text but not the tolerance text, and "approved but never drawn" has no outcome ledger.

**The decision, in three parts** (full text in `docs/adr/0019-…md`):

1. **The plan carries `display_text`** — tolerance suffix, fit class, collapse wording
   included. Renderers render, never compose; `_tol_suffix` moves below the model rank; drawn
   precision becomes a compile input; collapse honesty becomes one compiler rule. The #1215
   guard reduces to string equality against the plan.
2. **Dimension outcomes at one seam**, on both build routes, modelled on the existing
   hole/slot/channel/flat requirement-outcome ledgers — deleting the withholding/retraction
   machinery #1235 had to build three times before it was right.
3. **Per-mark identity for ladder rungs** (amending 0016 Amdt 3), so accounting is not blind
   to a partially drawn ladder.

Status **Proposed** — this PR is the review object; the three-phase work plan in the ADR is
evidence-gated and would be filed as issues on acceptance. Success criterion: a net-negative
diff.

Docs only; `pr-check --static` exit 0 (includes the mkdocs build), unit tier 193 passed.


pzfreo:
The overall direction is sound: compiler-owned display text and central outcome accounting address the repeated defect class. I think three design points need resolving in the ADR before implementation:

1. **Do not overload `DimensionId` with per-mark identity.** Lines 102–108 add `.0`, `.1`, etc. directly to `DimensionId`, but that type currently identifies the addressable dimension; a correlated ladder is deliberately one addressable set. Splitting it per rung risks breaking suppression, deduplication, lookup, and provenance equality while the ADR says addressing remains unchanged. Please introduce a separate mark identity such as `DimensionMarkId(DimensionId, discriminator)`: authored intent keeps the base ID, while registry accounting uses the mark ID.

2. **The proposed registry join cannot derive `withheld` from approved marks alone.** A compiler-withheld mark is absent from the approved plan, while a collapse-withheld tolerance can leave the underlying mark placed with only its band omitted. Neither is distinguishable from placement loss by joining approved entries to registry claims. The ADR should define an expected-mark/outcome record carrying mark identity, mandatory/optional status, the withheld component, and its reason before reconciliation. Otherwise the central seam recreates the withholding-code inference it intends to delete.

3. **Define one owner for precision and the compiled plan.** `compile_dimensions()` currently has roughly 30 call sites, including `sheet_emit`, evaluation, and lint paths with no `Draft`, and it is recomputed during rendering and later verification. Merely saying callers pass the build's precision permits different/default formatting policies to produce different `display_text` at render and reconciliation time. Please require one immutable formatting policy plus a build-owned compiled plan, or explicitly define the behavior of buildless consumers.

Small documentation correction: `docs/adr/README.md` still lists ADR 0018 as **Proposed**, while `CLAUDE.md` and `docs/architecture.md` say it was accepted on 2026-08-16.

I would request changes on the identity and outcome data models; once those are explicit, the proposed boundary looks substantially stronger.

pzfreo:
All three resolved in 5b0599e, each as you specified rather than a variant:

1. **`DimensionMarkId(dimension: DimensionId, discriminator)` as a separate type** (§3 rewritten). The base `DimensionId` stays the one addressable identity for suppression, dedup, lookup and provenance equality; the expected-mark table and registry claims use mark ids, with the dimension-level view derived from `mark.dimension`. Phase 4's gate now names those four consumer classes explicitly — they must pass unchanged with marks present.

2. **An explicit `ExpectedMark` record, emitted by the compiler** (§2 rewritten). Your diagnosis is stated in the ADR verbatim as the reason: a compiler-withheld mark is absent from the approved plan and a collapse-withheld band leaves its mark placed, so a bare claims-join would have to infer both back — recreating the withholding-code machinery this ADR deletes. The record carries mark id, `mandatory`, `display`, and per-component `expected | withheld(reason)`; refusals are written at compile time, never inferred at the seam.

3. **One plan, one policy, one owner** (new §4, now work-plan phase 1 since everything else stands on it). Measured before writing it: 32 `compile_dimensions` references across eight files, including `sheet_emit`, `evaluation` and `linting/evidence` with no `Draft` in reach. A frozen `FormattingPolicy` at the bottom rank (the engine default is `_fmt`'s current 1 dp; a `Draft` *carries* a policy); the plan compiled at most once per build, owned by `BuildState` on the ADR 0017 recognition-once pattern; `RenderableDimensionPlan.policy` makes a cross-policy join detectable instead of silently wrong; the buildless consumers compile under the stated default explicitly.

The 0018 status row in `docs/adr/README.md` is corrected to Accepted 2026-08-16 (nothing implemented yet).

Issue #1187, opened 2026-08-16
Dense sheets leave no clear leader route — escalate or replan, don't re-route


pzfreo:
> **Reframed 2026-08-16.** This was filed as "route leaders clear of the part in the semantics-bearing placers (hole callouts, boss diameters)" — a placer-refactor issue. Measurement after #798/#1188 shows that framing was wrong: there is nothing for a better router to find. The original text is in the history; the evidence that replaced it is below.

## The measurement

After #798 and #1188, ten leaders still cut the part across the fixture corpus. For each, sweeping the elbow through 64 directions × 6 shaft lengths — far more freedom than any producer offers — gives:

| part | leader | label | cut | shaft | routes clear of **material** | routes clear of **everything** |
|---|---|---|---|---|---|---|
| ctc_05 | `m_pocket_yz3` | `3.1 × 9.1 × 2.0 DEEP` | 9.9 mm | 79.8 | 333 | **0** |
| ctc_05 | `m_bossdia_z2` | `ø63.5` | 9.5 mm | 64.5 | 345 | **0** |
| ctc_04 | `hc_plan3` | `4× ⌀12 ↧ 20 ⌴ ⌀20 ↧ 20` | 2.7 mm | 32.5 | 350 | **0** |
| ctc_04 | `m_chamfer_x1` | `C5` | 1.6 mm | 27.6 | 344 | 17 |
| ctc_02 | `hc_plan8` | `5× ⌀8.4 ↧ 25` | 5.6 mm | 45.9 | 344 | **0** |
| ctc_02 | `hc_plan12` | `5× ⌀8.4 ↧ 25` | 0.6 mm | 45.5 | 310 | **0** |
| #881 flange | `hc_front0` | `4× ⌀4 THRU EQ SP ON ø50.9 BC` | 27.6 mm | 49.0 | 146 | **0** |

Hundreds of routes avoid the part. Essentially none avoid everything else — they collide with committed annotation ink, the page margin, or the silhouette.

**The part's shape is not what traps these leaders. The sheet being full is.**

## Why the original framing was wrong

[...]

Issue #798, opened 2026-07-21
Route ⌀/feature leaders clear of the part silhouette (design)


pzfreo:
Phase 2 follow-up to #796. Phase 1 (the `leader_crosses_silhouette` info-level lint) landed the observability net; this issue is the *routing* design work to actually keep leader shafts clear of the part body.

## Why it's a design task, not a nudge

Investigating #796 established that the crossings/near-misses aren't a free elbow choice:

1. **Row-packing forces the diagonals.** In `_diameter_row_below`/`_diameter_column_left`, when the ⌀ labels don't fit at their natural x's they spread by `min_gap`, so the elbow moves off the tip and the shaft goes diagonal. E.g. GRM-03's ⌀6 elbow is at x≈39.4 (tip at 36.9) purely because the row is full — it can't be "pulled left" without dropping a label or restructuring the row. (GRM-03's ⌀6 is a *near-miss*, not a true crossing.)
2. **Nested features can't be routed clear at all.** A ⌀6 boss stub protruding from a ⌀30 flange must cross the flange body to reach any row-below/column-left label — no elbow position avoids it. The honest fix is a **detail view** (the existing #304/#307 escalation) or a new **leader-to-part-end** placement mode, not a nudge.

## Scope to design

- Decide the policy per case: (a) de-diagonalise where the row has slack; (b) escalate genuinely-nested features to a detail view; (c) optionally a leader-to-end placement mode for external stubs.
- A **general** `leader_crosses_silhouette` geometry test (Phase 1 is scoped to profile-view ⌀/feature leaders via a ≥2-crossing bbox heuristic + hole-callout exclusion; a fully robust version wants a filled point-in-polygon / crossing-parity test over the projected silhouette, incl. curved end-view outlines).
- Fits ADR 0014 (collect-then-solve): side/elbow choice as part of the solve.

## Acceptance
- The nested-boss and thin-neck fixtures either route clear or escalate to a detail view; the Phase-1 notice goes quiet where a clear route exists.

Related: #796, #304/#307 (detail escalation), ADR 0014.

Issue #740, opened 2026-07-19
#735 child: collect-then-assign for leader-callout corner choice


pzfreo:
Child of #735 (5). Within _leader_callout_pass each job greedily takes its first clear corner in job order, so one callout can take another's only viable spot (within-pass order sensitivity — the last of its kind after #733 fixed the cross-pass ordering). Collect all jobs' candidate corners first, then assign jobs→corners maximising placements (then minimising total leader length); _assign_balloon_bands (layout.py) is the in-repo precedent for exactly this shape. Post-drain decoration only, so worst case equals today's greedy result. Medium PR.

PR #1252, opened 2026-08-21
ADR 0018 slices 1+2: the view topology becomes a value, and views report what they carry (#1130)

pzfreo:
The first slice the ADR names: "an explicit ViewSpec / ResolvedViewPlan
representation reproducing current behaviour". No behaviour changes; what
changes is that the topology stops being implicit.

Before, "which views exist and where they go" was spread across three places
that did not know about each other: `builder` named front/plan/side with literal
camera positions, their page coordinates lived in `Analysis` as `FV_X`/`PV_X`/
`SV_X`, and `compose.choose_scale` knew the arrangement only through a sentence
in its docstring ("Layout columns: [front(x×z)] [side(y×z)] [iso] [title
block]"). The topology was real, but there was nothing to read or replace.

Now `view_plan.py` (rank 0 — it describes views, it cannot reach what draws
them) holds `ViewSpec` (a request, in model terms, saying nothing about the
page) and `ResolvedViewPlan` (the immutable result). The split is ADR 0018 §1,
and it is a split rather than one mutable object because a resolved plan that
can be edited in place is indistinguishable from a request — which is how a
layout comes to be silently relaxed. `builder` creates its views FROM the plan,
so the drawing and its record of the topology cannot disagree; `compose` reads
its layout map from it. One typed `BuildState` attachment, a read-only
`Drawing.view_plan` with no setter, and ADR 0005's encapsulation guard is what
keeps it to one writer.

Also included, because it is what the ADR asks for first: a synthetic thin
X-axis rotational plate reproducing the motivating failure without the
proprietary STEP the ADR was proposed from. It hits the case study exactly —
bbox 43 x 217 x 217, A1 landscape at 1:1, four fixed views — and shows WHY that
is wrong rather than merely that it is big: `plan` is the same 42.5 x 217
projection as `front` and carries 1 annotation to front's 8.

Three things this turned up that are worth recording.

**A mutation I could not kill, and what it took to fix.** Re-routing
`compose._view_geom` was, at first, unverifiable: emptying its result left every
golden and every repack-seam test green, because it feeds only the
measure-and-repack loop and no fixture there triggers a repack. Emptying it also
raises `KeyError` inside `_measure_blocks`, which proves the function is CALLED,
not that its answer is used — the first version of the test mistook one for the
other. Halving the half-extents keeps every consumer working and changes only
what the repack measures, and on CTC-03 AP203 that changes the drawing. So the
consumer is load-bearing there, and the re-routing was verified by comparing
that fixture against main: page, scale, every annotation name and label, every
view's bounds — byte-identical, 3093 bytes of signature.

**A coupling I introduced and backed out.** Routing `_view_geom` through the
full resolver made a narrow consumer depend on `SCALE`, `PAGE_W` and `PAGE_H`,
and broke a repack test that passes a minimal stub carrying only the position
fields. The stub was right; `principal_placements` is the narrow function that
consumer should have been asking for.

**A ratchet I would have grown.** The bridge test read `drawing._analysis`,
which #741 caps and only allows to shrink. It captures the `Analysis` at the
seam that is handed it instead.

Mutations killed: a plan that disagrees with the analysis it was resolved from;
mutable placements behind a frozen dataclass; an open `kind` string.

4407 passed (fast), 46 passed (slow), pr-check --static exit 0, goldens
unchanged.

### DIFF
--- a/src/draftwright/builder.py
+++ b/src/draftwright/builder.py
@@ -75,6 +75,7 @@ from draftwright.projection import (
     _fit_iso_view,
     _project_iso,
 )
+from draftwright.view_plan import resolve_from_analysis
 
 # A view centre must move by more than this (mm) for the measure-and-repack
 # pass to re-assemble.  Below it, the estimate already matched the measured
@@ -448,11 +449,28 @@ def _assemble(
     # loop's intermediate assemblies will pass a cheap stand-in and only the final one the real
     # solid (#1137). Every caller currently takes the default, so nothing has changed yet.
     part_s = (a.part if shape is None else shape).scale(a.SCALE)
-    dwg._add_view(
-        "front", part_s, (cxs, cys - dist, czs), (0, 0, 1), (a.FV_X, a.FV_Y), scaled=True
-    )
-    dwg._add_view("plan", part_s, (cxs, cys, czs + dist), (0, 1, 0), (a.PV_X, a.PV_Y), scaled=True)
-    dwg._add_view("side", part_s, (cxs + dist, cys, czs), (0, 0, 1), (a.SV_X, a.SV_Y), scaled=True)
+
+    # ADR 0018: the views this drawing has, and where they go, come from ONE resolved plan
+    # instead of three hardcoded calls whose cameras, page fields and layout meaning were
+    # spread across this function, `Analysis` and `compose.choose_scale`'s docstring. The plan
+    # still describes exactly the third-angle front/plan/side set the engine has always built —
+    # this slice is representation, not selection — but it is now a value that a later slice can
+    # vary, and the projection convention is stated where the views are named rather than
+    # implied by three camera literals.
+    #
+    # Cameras are attached here because they need the scaled part's centre and the projection
+    # distance, which the plan deliberately knows nothing about: a `ViewSpec` is a request in
+    # MODEL terms, and a camera position in page-scaled coordinates is not one.
+    _CAMERAS = {
+        "front": ((cxs, cys - dist, czs), (0, 0, 1)),
+        "plan": ((cxs, cys, czs + dist), (0, 1, 0)),
+        "side": ((cxs + dist, cys, czs), (0, 0, 1)),
+    }
+    dwg._build.view_plan = view_plan = resolve_from_analysis(a)
+    for spec in view_plan.of_kind("principal"):
+        camera, up = _CAMERAS[spec.name]
+        place = view_plan.placements[spec.name]
+        dwg._add_view(spec.name, part_s, camera, up, (place.cx, place.cy), scaled=True)
     _project_iso(dwg, a, a.SCALE, shape_s=part_s)
 
     _diagnostics = None  # the audit ledger; filled once at the end of this function (#996)
--- a/src/draftwright/compose.py
+++ b/src/draftwright/compose.py
@@ -54,6 +54,7 @@ from draftwright._core import (
 )
 from draftwright.layout import fit_box
 from draftwright.model.callout import hole_callout_spec
+from draftwright.view_plan import principal_placements
 
 _log = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


pzfreo:
Re-reviewed after the fixture update. No remaining findings from me.

The overall approach looks right: separating the requested view vocabulary (`ViewSpec`), the immutable resolved result (`ResolvedViewPlan`), and finished-sheet measurement evidence (`ViewCoverage`) gives later view selection a clean compiler seam. In particular, treating “carries nothing exclusively” as a candidate signal rather than proof that a view is redundant—and failing closed where `DimensionId` cannot distinguish marks—is the correct level of caution.

The updated synthetic plate now forms one connected solid on both supported build123d branches, the Python 3.10 regression test passes locally, and the full CI matrix is green. Ready to approve.

pzfreo:
> [!CAUTION]
> **Correction — the comment below is wrong about this PR, and I am leaving it visible rather than deleting it.**
>
> I wrote it believing #1252 was open with seven commits awaiting review. **This PR merged at 15:43 UTC on `f75ffcb`**, four hours before I posted, and **none of the commits I describe were ever in it.** They were pushed to the branch *after* the merge, which is also why no CI ran on them — a merged PR's branch fires no `pull_request` workflow.
>
> So everything below is misdescribed: what merged here is slices 1–2 plus the fixture fix, fully CI'd, with **no behaviour change and no gate retirement**. The auto-merge fired correctly on a green matrix.
>
> The work it actually describes — arrangement selection, the requirement gate, the golden-gate retirement, view routing and view selection — is now in **#1258**, where it can get the CI it has never had. The two decisions I raised are real and belong there, not here.
>
> My error throughout: I queried this PR's `mergeStateStatus` and never its `state`.

---

Thanks — but flagging that this approval covers `f75ffcb` and the branch is now at `d8bf6bd`, **seven commits later**. Two of them change things I don't think an approval of slices 1–2 should carry, so please don't treat "ready to approve" as covering the current head.

### Needs your eye specifically

**1. User-visible output changed.** Slice 5 landed arrangement selection (ADR 0018 §5's fourth dimension) behind a requirement gate. Three corpus parts — `chamfered`, `filleted`, `prismatic_ladder` — now compose under a new `stacked-iso` arrangement and come out on **A4 instead of A3 at the same scale**, with identical annotations, labels and build issues, lint clean. `centered_rebate` and `scattered_plate` try the smaller sheet, lose a dimension, and are rejected back. That rejection costs a **second compile** for those parts, recorded in a new `Drawing.arrangement_decision`. Real drawings get smaller paper — that's a product call, and it's one line to defer.

**2. I retired a gate whose authority had expired.** `test_refactor_golden` asserted byte-exact coordinates and failed with *"NOT to be re-blessed away"*. Its own header said it was retained "until that epic concludes", naming #602/#638/#639 — all closed 2026-07-17, #740 on 2026-08-15. I'd been treating it as binding across six slices; worse, in slice 5 I edited that very header to add a re-bless justification without checking whether its premise still held.

The standing contract is now: no annotation vanishes, no label changes, no build issue gets worse (counted), and the sheet never grows — a ratchet, so footprint may shrink freely and may never creep back. Byte-exact survives as `DRAFTWRIGHT_GOLDEN_EXACT=1` and still passes on this tree. Swept the rest of the suite for the same pattern: 47 conditional-retention sites, 23 referencing closed issues, all but one historical narrative or deliberately permanent. The exception is fixed here — `holes.py` deferred to "#798 owns robust silhouette-aware routing", but #798 closed *without* delivering it; it now points at #1187.

### On your specific design point

You wrote that treating *"carries nothing exclusively"* as a candidate signal rather than proof of redundancy was the right level of caution. That was right, and slice 7 measured something stronger: **the signal is never positive.** Every principal view carries an exclusive requirement even on a featureless box, because the three envelope extents are distributed one per view. So it would never have proposed dropping anything.

The working criterion turns out to be whether what a view carries can be carried by a view that *remains*, which needed a map the engine didn't have — `_END_ON` answers "which single view does this read face-on in", not "which views could carry this". `VIEW_AXES` is that primitive now.

### Where the epic actually is

The case study moves for the first time: four views give **A1 at 1:1**; dropping the redundant plan gives **A2 at 1:1** — the ADR's stated target. It costs six annotations and gains `callout_dropped`, so a gate weighing it would reject it, and nothing selects automatically yet. `_views` is an engine seam, not a public option. Remaining work is re-homing those callouts to the axial view; the layout side is done.

Gates on `d8bf6bd`: 4473 passed, `pr-check --static` exit 0, goldens byte-identical under `EXACT`. Seven mutations killed by named tests. One earlier surviving mutation (the repack seam) is now covered; two mutation runs were invalid harness runs and were redone rather than counted.

Issue #28, opened 2026-06-15
feat: dwg.view_bounds(view) — page bounding box of a named view


pzfreo:
## Problem

`dwg.at(view, x, y, z)` converts a 3D world point to a page coordinate, but there is no way to query the extent of the projected view on the page. Scripts that place `Note`, `Leader` elbows, or other free-form annotations outside a view have to guess offsets from `dwg.at()` results.

This leads to labels placed on top of projection lines, inside the view boundary, or off the sheet edge.

## Proposed API

```python
x_min, y_min, x_max, y_max = dwg.view_bounds("front")
# page coordinates (mm from sheet bottom-left) of the tight bounding box
# of the projected geometry in this view
```

Usage:

```python
bounds = dwg.view_bounds("front")
# place a note just to the right of the front view
note_x = bounds[2] + 5   # 5 mm right of rightmost projection line
note_y = (bounds[1] + bounds[3]) / 2  # vertically centred
dwg.add(Note("SEE NOTE 1", position=(note_x, note_y), draft=dwg.draft))
```

## Notes

- Should return the bounds of the **projected geometry** (silhouette lines), not the allocated annotation zone.
- Returning `None` for unknown view names is safe.
- View centre is already calculable via `dwg.at(view, 0, 0, 0)` but that only gives the origin, not the extent.
- This pairs well with `dwg.features(view)` (#26) for placing callouts reliably outside the view boundary.

PR #1258, opened 2026-08-21
ADR 0018 slices 3-7: arrangement selection, the requirement gate, and a chosen view set (#1130)

### LINKED ISSUE #1250: The automatic path emits a sheet the engine refuses when asked for it explicitly, and reports passed: True
## The inconsistency

```
```

```
```

and reports:

```
```

## Reproduction

## Shape of a fix

pzfreo:
Continues ADR 0018 (#1130) from where #1252 merged. Seven commits: slices 3–7, the repack coverage, and the golden-gate retirement.

**Why a new PR:** #1252 auto-merged at 15:43 on `f75ffcb` (slices 1–2 + the fixture fix), correctly and fully CI'd. These seven commits landed on the branch *after* that, so no `pull_request` workflow could fire and none of them has been through the matrix. That is the main thing this PR is for — the 3.10/3.11/3.12 shards run build123d 0.10, which is exactly where the earlier fixture defect surfaced.

## The headline

ADR 0018's motivating case moves for the first time. The thin rotational plate: four views give **A1 at 1:1**; dropping the redundant plan view gives **A2 at 1:1** — the ADR's stated target.

It costs six annotations and gains `callout_dropped`, so a requirement gate weighing that candidate rejects it. Nothing selects view sets automatically yet and `_views` is an engine seam, not a public option. Remaining work is re-homing those callouts to the axial view; the layout side is done.

## Behaviour change — approved

Three corpus parts now compose under a new `stacked-iso` arrangement (the isometric sharing the title block's column rather than claiming its own) and come out **A4 instead of A3 at the same scale**: `chamfered`, `filleted`, `prismatic_ladder` — identical annotations, labels and build issues, lint clean. `centered_rebate` and `scattered_plate` try the smaller sheet, lose a dimension, and are rejected back unchanged.

Approved by @pzfreo in conversation, 2026-08-21. Real drawings get smaller paper. Cost is a second compile for parts that both depart from the default *and* lose something (2 of 14). Recorded in a new `Drawing.arrangement_decision`.

Two constraints keep it safe: an alternative arrangement may only compact a scale the default arrangement already chose (packing must not bid up legibility — unconstrained, the dense plate reached 2:1 and starved its location dims), and every candidate is proved by a real compile.

## Golden gate retirement — approved

`test_refactor_golden` asserted byte-exact coordinates and failed with "NOT to be re-blessed away", while its own header said it was retained "until that epic concludes" — #602/#638/#639 closed 2026-07-17, #740 on 2026-08-15. A byte-exact gate on layout the engine is being taught to improve fails every improvement and passes everything that improves nothing.

Standing contract is now: no annotation vanishes, no label changes, no build issue gets worse (counted), the sheet never grows (a ratchet). Byte-exact survives as `DRAFTWRIGHT_GOLDEN_EXACT=1` and passes on this tree. Approved by @pzfreo in conversation, 2026-08-21.

## What else is in here

- **Slice 3** — the layout choice becomes a value carrying all four ADR 0018 §5 dimensions.
- **Slice 4** — the second arrangement implemented, plus the measurement of why it could not then be selected.
- **Slice 5** — the decision carried once (`ScalePick`, `Analysis.arrangement`), and ADR 0018's first hard gate applied to it. That gate already existed for explicit *scale* and never ran on the automatic path; applying it there closes #1250.
- **Slice 6** — one owner for view routing. It was a bare literal at eleven sites across seven modules (two distinct tables, one of which documented its own duplication in a comment). `Drawing.at` now raises `ViewNotPlanned` naming the view and what the sheet has.
- **Slice 7** — a chosen view set is buildable and refusable; the layout reclaims a dropped view's paper.

Also fixes two defects found by running rather than reading: `_diameter_row_below` unpacked `view_bounds("front")`, which has documented `None` for an absent view since #28; and the arrangement gate's fallback rebuild silently reverted a two-view request to four views on a larger sheet.

## Evidence

- 4473 passed, `pr-check --static` exit 0, goldens byte-identical under `EXACT` (local, Python 3.14 only — hence this PR).
- Mutations killed by named tests across the slices; one earlier survivor (the repack seam) is now covered by a test pinning both arrangements, and two invalid harness runs were redone rather than counted.
- ADR 0018 delivery-gate items answered: the asymmetric counterexample (same box, same hole, Z vs X — dropping plan refused for one, clean for the other), and a truly redundant view dropped retaining every requirement.


pzfreo:
Review findings:

1. **The requirement gate compares only blocker counts, so it can accept a different loss.** `_preserve_requirements_under_arrangement` keeps the alternative whenever `len(preferred_blockers) >= len(blockers)`. If the preferred layout drops requirement B and the alternative drops requirement A, both have one blocker and the alternative wins—even though A was preserved by the default. That contradicts the stated “preserve every supported requirement” contract. Please compare blocker/requirement identities as a multiset (and reject any new blocker relative to the preferred result), rather than comparing cardinality alone.

2. **Reduced view sets are only reflected in layout for a missing plan view.** `_layout_geometry` checks `has_plan`, but always includes front and side widths, footprints, obstacles and fit accounting. `_views` can omit any principal view, and the resolver/builder honor that set, so omitting front or side creates a sheet without the view while still reserving its paper. The candidate also always records `third_angle_view_names()` rather than the requested `views`, making its first-class infeasibility data disagree with what `_fits` actually evaluated. Please derive all packed blocks and `LayoutCandidate.views` from the requested set, or explicitly constrain this seam to plan removal only.

The focused ADR 0018 tests pass locally on Python 3.10; these are uncovered semantic cases rather than current test failures.

pzfreo:
Both confirmed by execution before fixing, and both real. Fixed in `d0b6b4b`.

**1. Blocker counts → identity multiset.** Exactly as described: `len(preferred) < len(alt)` accepts a *different* loss. `_blocker_identity` now keys on what was lost — code plus measurement/requirement/source ids — and the alternative is rejected if it introduces any blocker the preferred result did not already have.

Deliberately **not** keyed on the message: messages carry positions and sheet sizes, so two layouts of the same defect read differently, and keying on them would make every blocker unique and reject every alternative always. There's a test pinning that.

The rule is one-sided, which I want to be explicit about since it isn't purely "fewer is better": the alternative does **not** have to beat the default on volume. The default is the baseline every drawing had before this choice existed, so an alternative earns its place by costing nothing *new*. An alternative that introduces nothing and preserves strictly more still wins.

**2. Only `has_plan` reached the layout.** Confirmed — `views=('front','plan')` reported `row_w=313.00`, identical to the full three, so it built a sheet without the side view and kept its paper.

Generalised rather than constrained to plan-removal, because the tests that prove refusal and reporting need to drop front and side. Every term is now conditioned on presence. The asymmetry is worth stating: front and plan **share** the column and both project x across the page, so dropping one reclaims *height* and only dropping both reclaims the column's width. Measured — `('front','plan')` reclaims the side's 60 mm, `('side',)` reclaims the column's 90 mm, full set unchanged at 313.00 with goldens byte-identical under `EXACT`.

`LayoutCandidate.views` now carries the set actually judged, observed at the predicate since rejected candidates aren't exposed on a result yet.

**On the tests.** Six mutations, all killed by a named test. Three survived a first pass and I fixed the tests rather than reporting the result: the gate test recomputed the multiset difference itself instead of calling `_preserve_requirements_under_arrangement`, so it passed whatever production did — the same fake-guard shape as an earlier repack test in this branch; no test dropped both column views; and the candidate test asserted only that `choose_scale` returned non-None.

4484 passed, `pr-check --static` exit 0. Your read that these were uncovered semantic cases rather than failures was right — nothing in the suite was catching either.

pzfreo:
Re-reviewed through `232bb44`, including ADR 0018 Amendment 1. Both earlier implementation findings are addressed correctly: blocker comparison is now identity-aware and multiset-based, and layout/candidate accounting now follows the actual requested view set. The focused ADR 0018 tests pass locally on Python 3.10.

Two API-model issues remain in the amendment:

1. **The proposed `ViewPlanIncomplete` example contradicts itself.** It declares the authored set as `("front", "side")`, then says `envelope.depth` reads horizontally only in `side` and recommends `add view("side")`; the following provenance line also confirms that `side` was already declared. Either the set should be `("front", "plan")`, or that first missing requirement/recommendation must change. This example defines the public diagnostic contract, so it should be internally executable rather than merely illustrative.

2. **The replacement section/detail verbs leave no specified augmentation path.** The amendment says `section_view(...)` / `detail_view(...)` join the authored view set, while `add_view(...)` is the only verb that augments `auto_views()`. As written, a user cannot force a targeted section/detail while retaining automatically selected principal views (and, transitively, automatic dimensions); they must switch the entire sheet to authored views/dimensions. That removes the useful capability currently provided by `Sheet.section()`—forcing a section on an otherwise automatic drawing. Please specify an augmenting form for derived views (for example `add_section_view` / `add_detail_view`, or define these verbs as augmenting when the source is automatic), including how it interacts with the two-source rule.

The overall request/result lifecycle and the ban on authored principal views + automatic dimensions are otherwise coherent.

pzfreo:
Both real, both fixed.

**1. The example contradicted itself.** Exactly as described — set declared as `("front", "side")`, then `add view("side")`, with the provenance line below confirming `side` was already there. Now `("front",)`, with two requirements giving two *different* answers: `envelope.depth` needs `side`, `hole_1.location` needs `plan`, and `envelope.width` resolves to `front` so it is correctly absent from the report.

Derived from the maps rather than sketched, and the ADR now says so: with `("front",)` authored, `views_showing("y", horizontal=True)` is `None` with only `side` qualifying, and `_END_ON["z"]` is `plan`. Your framing is the right one and I've adopted it — a block defining a public diagnostic contract has to be executable, not illustrative.

**2. The augmentation hole was a real capability regression**, and worse than "unspecified": as written it would have made `section_view(...)` select the authored source, trigger the ban, and cost a user their automatic principal views *and* their automatic dimensions in order to force one section. Replacing a verb with something strictly less capable is not a reshape.

Derived views are now their own set with the same three verbs, and I've stated the general rule rather than leaving it implicit in the principal-view case:

- an **authoring** verb defines a set, so omission is significant, so it can strand a planner-chosen dimension in a view that is not there — banned with `auto_dimensions()`;
- an **augmenting** verb is purely additive, removes nothing, can strand nothing — legal with `auto_dimensions()`.

The principal-view ban is now one case of that rule rather than a rule of its own. `Sheet.section(feature)` gets a like-for-like replacement:

```python
s = Sheet(part).auto_views().auto_dimensions()
s.add_section_view("A", through=bore)
```

And `plan_sections` stays the automatic derived set, so omission in an authored derived set means something: no `section_view(...)` line is how you say "no section, even though the counterbore would fire one".

---

Note on CI: this branch has had six pushes in quick succession, each cancelling the previous run, so **no matrix run has completed since `f75ffcb`** — slices 3–7 have only ever been verified locally on Python 3.14. The branch is now rebased onto main (`625f8d4` → this commit) and I'm holding off further pushes so the run can finish. The shards that matter are 3.10/3.11/3.12 on build123d 0.10, which this layout code has never run against.

Issue #1262, opened 2026-08-21
ADR 0018 Phase 6: automatically select the smallest complete semantic view set


pzfreo:
## Outcome

Select a conventional view set from drawing requirements, jointly with arrangement, preferred ISO scale and standard paper, while treating supported-requirement survival as a hard gate.

Parent tracker: #1130. Depends on the view-set-aware planner and complete re-homing of eligible requirements.

## Scope

- Generate dominant-axis axial/longitudinal candidates for rotational parts.
- Remove redundant views only when every supported requirement remains covered.
- Consider longitudinal sections for internal axial/depth requirements.
- Size isometric orientation views independently as bounded NTS aids.
- Optimise feasible candidates lexicographically as specified by ADR 0018.

## Acceptance evidence

- The synthetic worm-style plate selects a complete A2 1:1 plan when feasibility confirms it.
- A purely axisymmetric stepped shaft does not receive three redundant orthographic views.
- A turned part with an asymmetric end-face feature receives the additional view required for it.
- GRM-01 either fits A4 at 2:1 through a coherent arrangement/view set or refuses the authored constraint explicitly.
- GRM-04 never accepts a smaller/sparser plan that drops the authored Ø2.4 callout.
- Selected view IDs and requirement coverage are asserted independently of final appearance.

## Non-goals

- Multi-sheet planning.
- Dimensioned isometric drafting.
- Reduced paper-space typography.

PR #1265, opened 2026-08-21
Give the automatic path the completeness the explicit path already had (#1250)

### LINKED ISSUE #1250: The automatic path emits a sheet the engine refuses when asked for it explicitly, and reports passed: True

pzfreo:
Fixes #1250 — reopened after checking, because it was auto-closed by a keyword when #1258 merged and the defect still reproduced on `main`.

## The defect

```
build_drawing(part)                       -> A1 @ 1:1, passed: True, 0 lint errors
build_drawing(part, page="A1", scale=1.0) -> ScaleIncompatibilityError
```

`_scale_blockers` ran on the explicit-scale policy loop and on nothing else, so how the caller phrased the request decided whether the same drawing was acceptable. The automatic A1 sheet carries a slot's length and position dropped and reported success.

## Resolution

The automatic path now applies the recognition-free required-outcome gate to its settled drawing. If required placement outcomes were lost, it records `plan_incomplete` at error severity, sets `scale_decision.status` to `incomplete`, and returns the diagnostic drawing. `lint_summary().passed` can therefore no longer certify a drawing the explicit path rejects.

`plan_incomplete` is an unscored summary: the underlying drops remain the score owners, so the new error changes the verdict without double-counting quality. The summary preserves every existing provenance channel — `measurement_ids`, `hole_requirement_ids`, and `source_ids` — so source-only PMI/GD&T and recognition-owned physical requirements remain addressable.

## Why this does not search smaller scales

The first implementation rebuilt at smaller scales and accepted a clean candidate when its registry measurement-id set covered the incumbent's. Independent review found that comparison was only a lower bound: unknown, source-only, and physical-hole requirements could disappear while both compared sets remained empty. It also introduced a second scale-selection policy ahead of ADR 0018's joint page/scale/view planner and increased the fast tier from roughly 190s to 279s.

The final implementation fails closed instead. It performs no speculative rebuild and leaves candidate search to #1262, where the planner can compare the full semantic requirement inventory under the ADR 0018 ordering.

## Results

| part | before | after |
|---|---|---|
| #1130 case study | A1 1:1, required drops, `passed: True` | A1 1:1, `plan_incomplete`, `passed: False` |
| source/physical-only synthetic drop | unaddressed summary / unsafe candidate acceptance | all provenance preserved; no candidate accepted |
| plain box | 1:1 clean | unchanged |

Three goldens were re-baselined with annotations, labels, page, and scale unchanged: one honest build issue was added and none removed.

## Review evidence

- Independent review reproduction against the original head: `accepted_unknown_loss True fallback`; summary identities `() () ()`.
- Focused source-only + physical-requirement regression added.
- `scripts/pr-check --static`: clean.
- Affected cluster: 74 passed, 1 deselected.
- Focused PMI/footprint cluster: 12 passed.
- Full cross-version CI and platform canaries are the merge gate.


pzfreo (commented):
Independent review found two related substantive provenance gaps in the new completeness gate:

1. `_placed_requirements()` compares only `registry.measurement_of(...)`. Its own premise says an empty tuple means unknown, so that set is a lower bound and makes the acceptance test weaker, not stricter: candidate content with hole/source-only or not-yet-threaded identity can disappear without participating in `>= baseline_placed`. Please either compare the full semantic coverage inventory used by the gate or state and fail closed on unverifiable coverage.

2. The `plan_incomplete` summary copies only `measurement_ids`, even though `_scale_blockers()` deliberately treats `hole_requirement_ids` and `source_ids` as first-class blocker identities. A placement-stage PMI/GD&T loss may therefore produce an unaddressed summary. Please preserve all three identity channels and add a focused source-only/hole-requirement assertion.

I’m otherwise still reviewing the control flow and tests; these are non-nit correctness findings because ADR 0018/0017 require requirement survival by semantic identity rather than the currently migrated subset.

pzfreo:
Concrete reproduction for the two review findings, run against `38c5ab5` with a duck-typed drawing (the helper itself is duck-typed):

```
accepted_unknown_loss True fallback
summary_ids () () ()
```

The baseline had one annotation whose registry measurement identity was unknown and a placement-stage `pmi_dropped` carrying only `source_ids=('pmi-7',)`. A clean candidate with that annotation absent was accepted as the fallback because both `_placed_requirements` sets were empty. When every candidate retained the source-only blocker, the final `plan_incomplete` carried none of `measurement_ids`, `hole_requirement_ids`, or `source_ids`.

This confirms both are reachable control-flow defects, not documentation concerns.

pzfreo:
Resolved both substantive review findings in `09b54d3` and re-reviewed the full diff against ADR 0017/0018.

The fix removes the speculative smaller-scale rebuild loop entirely. Registry measurement provenance is still intentionally partial, so no candidate can be proven to preserve source-only/physical/unknown requirements here; the automatic path now fails closed on its settled drawing and leaves joint page/scale selection to #1262. `plan_incomplete` now copies `measurement_ids`, `hole_requirement_ids`, and `source_ids` from every required placement drop.

Added a focused source-only + physical-requirement regression test. Local evidence: static preflight clean; affected cluster 74 passed, 1 deselected; focused PMI/footprint cluster 12 passed. Second review found no remaining substantive issues or nits.

pzfreo:
CI follow-up `88416e9`: Python 3.14 shard 2 exposed two identical side-drilled fixtures whose old assertions called the drawing lint-clean while both Z-height requirements were recorded as `off_axis_location_dropped` at info severity. Both X offsets and both callouts land; the height companions genuinely do not.

Updated those subject-specific tests to preserve their placement assertions and pin the honest `plan_incomplete` error plus both underlying drops. Focused tests pass 2/2 and static preflight remains clean. This is an in-scope #1250 expectation correction, not a gate exemption.

pzfreo:
Additional final-head evidence: the complete local fast tier exits clean, and the full `tests/test_make_drawing.py` module is 731 passed, 3 skipped, 1 xfailed. No further review findings.

PR #1267, opened 2026-08-21
Stabilize the slow repack seam test

### LINKED ISSUE #1266: Slow repack seam test repeatedly times out under post-merge xdist load
## Problem

The post-merge slow tier is unreliable because `tests/test_view_plan.py::test_the_repack_loop_really_consumes_the_plan` inherits the global 300 s timeout while doing two full CTC-03 builds under `pytest -n auto` resource contention.

This is not a one-off runner failure: the same test timed out on three consecutive `main` runs after it was introduced:

- run 32534063339 (merge `661b27d`)
- run 32527329601
- run 32493712059

There was one intervening successful run (32470057086), so the failure is load-sensitive rather than deterministic. The timeout lands inside OpenCascade text-shape transformation during the second perturbed build. An isolated local run passes in 79.69 s, confirming the assertion itself is healthy while the suite-level budget is not.

The triggering merge #1263 changed only project audit/docs files, so this is independent of that PR's product scope.

## Required outcome

- Preserve both semantic claims of the seam test: `_view_geom` remains load-bearing, and perturbing its measured extents changes the repacked result.
- Give this intentionally heavy two-build test an honest execution envelope or isolate/restructure it so normal xdist contention cannot exceed its budget.
- Do not weaken or skip the assertion.
- Demonstrate a green post-merge slow tier.

pzfreo:
## Outcome

The intentionally heavy repack seam test now has a 600-second per-test timeout, matching the execution envelope already used by other full CTC builds. It still performs both real builds and keeps the load-bearing perturbation assertion intact.

This fixes a repeated post-merge CI failure: the test exceeded the global 300-second cap on 3 of 4 recent `main` runs while contending with the other xdist worker, despite passing in isolation.

## ADR assessment

No compiler, placement, view-plan, or public-surface behavior changes. ADR 0018's seam evidence is preserved exactly; this change only makes the test's declared resource budget honest under the slow-tier execution model.

## Verification

- `scripts/pr-check --static`
- `uv run pytest tests/test_view_plan.py::test_the_repack_loop_really_consumes_the_plan -m slow -q` (1 passed in 71.61 s)

Fixes #1266

### DIFF
--- a/tests/test_view_plan.py
+++ b/tests/test_view_plan.py
@@ -165,6 +165,7 @@ class TestTheRepresentationChangedNothing:
 
 
 @pytest.mark.slow  # a CTC fixture build, twice (#153)
+@pytest.mark.timeout(600)  # Two OCC builds exceed the global 300 s cap under xdist load (#1266).
 def test_the_repack_loop_really_consumes_the_plan():
     """The one consumer whose re-routing the golden corpus does NOT cover.
 


pzfreo:
Independent re-review: clean beyond nits. The diff is one test-only timeout override on the already-`slow` two-build CTC seam; it preserves both ADR 0018 assertions, does not skip or xfail evidence, and matches existing 600 s CTC budgets. Local static checks and the exact slow test pass (71.61 s in isolation). Auto-merge is enabled pending CI.

Issue #1268, opened 2026-08-22
Make the CTC automatic corpus complete instead of diagnostic


pzfreo:
## Problem

After #1250 made automatic completeness fail closed, the slow CTC corpus revealed that 9 of its 10 automatic drawings are diagnostic rather than complete:

- AP203 01–05: `plan_incomplete`
- AP242 02–05: `plan_incomplete`
- AP242 01: complete

The underlying outcomes include `callout_dropped`, `chamfer_dropped`, `fillet_dropped`, `hole_pattern_dim_dropped`, `location_ref_dropped`, `off_axis_location_dropped`, `pocket_dim_dropped`, `pocket_dropped`, and `slot_dim_dropped`.

This was previously hidden because the end-to-end standards tests treated warning-severity placement drops as lint-clean. #1250 correctly changed the verdict to `passed: False`; that diagnostic must not be downgraded. An explicit A1/0.5 probe of AP203-01 exhausted the current fallback scale sequence without finding a complete plan, so this is not merely an automatic scale-selection typo.

## Dependencies and scope

Build on the view-set-aware planner (#1259), requirement re-homing (#1261), and joint candidate selection (#1262). For each remaining CTC fixture, classify whether the blocker is view coverage, ownership, placement, or an explicitly unsupported capability, then make supported requirements survive the selected plan.

## Acceptance

- Each of the ten AP203/AP242 CTC variants either produces a complete automatic drawing or reports a named unsupported capability rather than a placement loss.
- No fixture is certified complete while any supported requirement outcome is dropped.
- `DimensionId`, hole-requirement, and source provenance survive any re-homing or candidate change.
- The slow corpus removes each `expect_incomplete=True` only when the corresponding drawing is genuinely lint-clean.
- Mutations that remove a required outcome or suppress the terminal diagnostic fail.

Issue #1259, opened 2026-08-21
ADR 0018 Phase 5.5: make dimension planning view-set aware


pzfreo:
## Outcome

Make the compiler plan dimensions against the selected view set before rendering, so an authored or candidate view set is either coherent or rejected with a diagnostic naming the unmet requirement.

Parent tracker: #1130.

## Scope

- Pass the planned view set into `planner._group_view`.
- Re-plan requirement ownership when a view is removed instead of rendering a fixed-topology plan into fewer views.
- Move view-set feasibility to the planner boundary defined by ADR 0018 §5.
- Add a pre-projection coherence check for authored view sets.
- Preserve ADR 0016 dimension identity and ADR 0017 requirement outcomes; do not use labels or page coordinates as correspondence.

## Acceptance evidence

- A requirement whose preferred view is absent is assigned to another eligible planned view or reported as uncovered before rendering.
- A Sheet-authored incoherent view set produces a diagnostic naming the authored constraint, not a late `ViewNotPlanned` exception.
- The asymmetric Z-hole/X-hole counterexample retains the necessary view and permits removal only where coverage is unchanged.
- Mutations that restore fixed-topology routing or silently discard an unassigned requirement fail.
- Focused, static, full fast-tier and applicable slow checks are green.

## Non-goals

- Public Sheet view verbs.
- Automatic semantic view selection.
- Annotation coordinate control.

pzfreo:
Delivered and verified.

- PR: #1271
- Merge commit: `339d37f79957f9aab59ff9fb58029afe5b4128bd`
- PR CI: all Python 3.10–3.14 shards, macOS/Windows canaries, lint, `ci-ok`, Codecov patch, and Codecov project green.
- Local evidence: full fast suite `4492 passed, 5 skipped, 3 xfailed`; applicable slow cases `2 passed`; branch-aware changed-line coverage 97% after the final guard expansion.
- Post-merge evidence: CI run https://github.com/pzfreo/draftwright/actions/runs/32548399927 succeeded, including `test-slow`; Documentation and Publish also succeeded.
- Review: no unresolved or outdated review threads; final independent ADR 0016/0017/0018 review found no non-nit issues.
- Tracker: Done, horizon/iteration cleared; `scripts/project-audit 3 pzfreo draftwright` green.

PR #1269, opened 2026-08-22
Make the slow CTC contract honest about incomplete plans

### LINKED ISSUE #1250: The automatic path emits a sheet the engine refuses when asked for it explicitly, and reports passed: True

pzfreo:
## Outcome

The slow CTC end-to-end corpus no longer calls diagnostic drawings standards-clean after #1250 correctly makes required placement losses fail the lint gate.

The shared helper is split into:

- an export/structure contract that applies to both complete and diagnostic drawings;
- the original lint-clean standards contract for ordinary examples;
- a CTC diagnostic contract that pins the current truth.

All five AP203 fixtures and AP242 02–05 must now contain exactly one error, `plan_incomplete`, with `scale_decision.status == "incomplete"` and `lint_summary()["passed"] is False`. AP242-01 remains required to be lint-clean and passed. This is deliberately not a filter, warning downgrade, skip, or xfail.

The actual CTC placement debt is tracked separately in #1268, depending on #1259/#1261/#1262.

## ADR assessment

ADR 0018 §6 says a renderable infeasible candidate is returned as a diagnostic drawing and that lint/quality callers must fail it. The old slow helper contradicted that contract by requiring every CTC diagnostic to have zero lint errors. This change preserves diagnostic exportability while mutation-proofing the failing verdict. ADR 0019's outcome ownership is unchanged.

No production code, annotation coordinates, placement policy, or public API changes.

## Verification

- `scripts/pr-check --static`
- fast end-to-end tier: 4 passed
- complete slow CTC file: 18 passed in 98.09 s
- final exact-error representatives: 2 passed in 82.40 s
- project audit: green

Fixes #1250

### DIFF
--- a/tests/test_e2e_standards.py
+++ b/tests/test_e2e_standards.py
@@ -1,7 +1,6 @@
-"""End-to-end tests: build a drawing from STEP or a build123d object and assert
-it meets the mechanically-checkable drawing standards.
+"""End-to-end tests: build and export drawings, then inspect their honest result.
 
-These check the parts of "meets standards" a machine can verify:
+The small representative parts check the parts of "meets standards" a machine can verify:
 
 - no error-severity lint violations (axis swaps, label mismatches, page bounds,
   view overlaps) — warnings are tolerated;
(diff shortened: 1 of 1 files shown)


pzfreo:
Independent re-review: clean beyond nits. The published diff does not suppress `plan_incomplete`: it makes each known diagnostic fixture prove exactly one terminal error, `status=incomplete`, and `passed=False`, while the clean fixture and ordinary examples still require zero errors. This matches ADR 0018 §6 and preserves ADR 0019 outcome ownership. Full slow file 18/18; exact-error representatives 2/2; static checks clean.

PR #1273, opened 2026-08-22
Lower AP242 hole tolerances onto Sheet features

### LINKED ISSUE #1116: Lower AP242 hole tolerances into declarative Sheet features
## Problem

A real AP242 benchmark session using:

`nist_ctc_01_asme1_ap242-e1.stp`

showed that generated Sheet scripts can split one semantic hole tolerance across two representations:

- the primary hole/pattern declaration (for example, `sheet.hole(...)`); and
- fallback raw `PmiFeature` / measured-dimension records carrying diameter limit or bilateral tolerance data.

This produces duplicate competing callouts. In the observed A2 drawing, raw `ø35 -0.2` and `ø35 +0.2` PMI collided with the primary hole-pattern diameter callout and were dropped. An angular fallback (`60 ±0.5`) also reached linear measured-label comparison and produced `label_vs_measured`.

Manually lowering the diameter tolerances onto the corresponding declarative holes with `.tolerance(...)`, then removing the redundant fallback PMI, changed the lint result from 0.85 with warnings to 1.0 with no warnings.

This is an emitter fidelity problem, not merely a placement problem: one authored semantic fact has multiple downstream owners.

## Desired behavior

During Sheet script emission, lower supported AP242 dimensional tolerances into the canonical declarative feature/dimension that owns the measured characteristic.

For a hole or hole pattern, supported diameter limit/bilateral PMI should produce one declaration equivalent to:

```python
sheet.hole(...).tolerance(lower, upper)
```

The same source PMI must not also be emitted as a fallback raw callout.

Unsupported or ambiguous PMI must remain explicit and provenance-preserving; it must not be silently discarded or guessed.

## Scope

- Map AP242 diameter limits/bilateral tolerances to the matching hole or hole-pattern diameter parameter.
- Preserve authored units, sign, nominal value, and asymmetric limits.
- Consume a source PMI record exactly once when lowering succeeds.
- Keep unmatched, ambiguous, or unsupported PMI in the explicit fallback path with a reason.
- Route angular PMI through angular semantics; never compare an angular label as a linear path length.
- Preserve provenance so consumers can distinguish authored PMI from inferred/default manufacturing information.

## Acceptance criteria

- [ ] A fixture based on NIST CTC 01 lowers the ⌀25, ⌀35, and ⌀20 tolerances onto their canonical feature declarations.
- [ ] Regenerated scripts contain no duplicate raw diameter PMI for successfully lowered tolerances.
- [ ] Direct and regenerated models agree on nominal diameter, lower/upper tolerance, feature identity, and pattern membership.
- [ ] Lint reports no `pmi_dropped` caused by duplicate lowered/fallback representations.
- [ ] Angular tolerances do not trigger linear `label_vs_measured` comparison.
- [ ] Ambiguous feature matching fails closed into a documented fallback rather than attaching tolerance to the wrong feature.
- [ ] Tests cover symmetric, asymmetric, unilateral, limit-dimension, pattern, ambiguity, and unit-conversion cases.
- [ ] The generated script identifies whether each tolerance is authored AP242 PMI or inferred/default data.

## Context

This was the highest-leverage manual correction in an iterative Draftwright session: layout selection improved presentation, but semantic lowering removed the duplicate annotations at their source.

Related: #964 (semantically faithful generated Sheet scripts).

pzfreo:
Closes #1116

## Summary

- correlate supported AP242 diameter tolerances with their canonical hole or pattern owner at the completed IR waist
- preserve source provenance in `ToleranceDecoration` and emitted `.tolerance(..., source=..., source_ids=...)` declarations
- fail closed to provenance-preserving measured dimensions for unmatched, ambiguous, conflicting, partial-pattern, or unsupported requirements
- reconcile lint stage accounting with lowered feature ownership and document the narrowed ADR 0016 rule

## Verification

- `scripts/pr-check --full` — 4,523 passed, 5 skipped, 3 xfailed; 99% diff coverage
- focused #1116 suite after independent-review addition — 18 passed (slow CTC fixture excluded)
- `scripts/pr-check --static` — clean
- NIST CTC01 slow fixture verifies all seven hole-tolerance sources are consumed exactly once, the angular source remains angular, generated Sheet provenance round-trips, and no duplicate `pmi_dropped` remains

## Review notes

The lowering pass is semantic only: it does not place annotations or introduce coordinates. The existing planner and placement solve remain the sole rendering/placement owners, consistent with ADRs 0004, 0011, 0014, 0015, and amended 0016.

### DIFF
--- a/docs/adr/0016-declared-dimensioning-intent.md
+++ b/docs/adr/0016-declared-dimensioning-intent.md
@@ -838,7 +838,8 @@ constraining planner internals that are still moving. The gaps are known and fin
 | Correlated sets, per member | `step_height` / `step_position` ladders and rotational bores are one `AddressableDimension` holding N members | **One** line per set; suppress the set, not a member |
 | Location dimensions, per member | Addressable per FEATURE since #925 (`dimension(hole, "location")`); whether a patterned hole's position splits per member is **#883** | **One** line per feature |
 | Inter-feature spans and angles | No `(feature, role)` form — needs `RelationDimensionId`, whose selector spelling is still open | Comment floor until the relation selector lands |
-| Imported AP242 PMI | Materialized: carries `ref_pts` / `ref_bbox` / `at`, so there is nothing to reference | `sheet.measured_dimension(...)` — still one editable line |
+| Supported AP242 feature enrichment with no proven canonical owner | Materialized: its source geometry cannot safely identify a feature parameter | `sheet.measured_dimension(...)`, carrying a structured lowering blocker |
+| Supported AP242 feature enrichment that proves a canonical parameter | Referential enrichment: the source geometry uniquely identifies the feature/member and parameter | The feature line's aspect, e.g. `hole.tolerance(..., source="ap242_pmi", source_ids=...)` |
 | Low-level furniture | Centre marks, section arrows, hatching, the NTS caption carry no editable intent | Engine-automatic, by decision |
 | Anything the emitter cannot re-solve | The fidelity floor `emit_sheet_script` already holds for features | Self-describing comment |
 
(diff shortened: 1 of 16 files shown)


pzfreo:
Independent review completed against #1116 and ADRs 0004, 0011, 0014, 0015, and 0016.

One substantive finding was resolved before push: pattern lowering had direct-IR coverage but did not prove emitted Sheet reconstruction preserved `PatternFeature` identity/membership. The test now executes the emitted declaration and verifies the pattern members, bore diameter, tolerance value, AP242 source, and source IDs round-trip.

No unresolved non-nit findings remain. Local gates: 4,523 passed / 5 skipped / 3 xfailed, 99% diff coverage; focused suite after the review change 18 passed; static/docs clean.

Issue #1190, opened 2026-08-16
Section A–A withheld with ample room, non-monotonic in scale, and inconsistently reported


pzfreo:
# Section A–A is withheld when there is ample room, and its presence is non-monotonic in scale

**Draftwright 0.4.8** · build123d 0.11.1 · Python 3.14.7 · macOS (arm64)

Three related defects in automatic section placement. The first two make
`section_aa` appear or disappear for reasons unrelated to whether the drawing
needs it; the third makes the omission undetectable. This matters because the
section is the only view that shows internal profile, and nothing in
`lint_summary` reports that it was dropped.

---

## 1. The room check refuses a section with ~3.6× the needed space free

`GRM-01` is an 80 × 13.55 × 11 mm shank with a Z-axis counterbored hole
(`⌀5 THRU ⌞⌟ ⌀6.3 ↓ 5.5`), so the automatic section is warranted. On **A4 at 1:1
the section is omitted**; on A3 at 1:1 it is placed.

Measured on the A4 sheet:

| quantity | value |
|---|---|
| page width | 297.0 mm |
| side view silhouette | x 116.3 … 127.3 |
| iso view silhouette | x 250.3 … 263.4 |
| gap between side and iso | **123.0 mm** |
| free band in the view row after annotations (measured from the render) | **49.6 mm** (x 195.2 … 244.8) |
| section silhouette width (as placed on A3) | **13.6 mm** |

The section needs 13.6 mm and 49.6 mm of the view row is genuinely empty ink —
the free band sits exactly where A3 places the section (x 207.7 … 221.2).

The refusal is not caused by annotation load. Removing the pocket callout, the
hole location dimensions, and the envelope dimensions individually and together
leaves the section omitted in every case.

Sweeping page width at 1:1, height fixed at 210 mm:

```
297x210  no      340x210  no      380x210  no
310x210  no      360x210  no      400x210  YES
320x210  no                       420x210  YES
330x210  no
```

So placement requires ≥ 400 mm of page width for a 13.6 mm view.

`GRM-01` itself logs nothing about the section (see defect 3), but the same
refusal on another part reports:

```
Section A–A skipped (no room right of the side view;
a wider step-dimension corridor may have reduced the available space)
```

If that is the same cause here, it fits: `GRM-01` declares a `step_level`, and
the step-dimension corridor appears to reserve far more width than it draws —
the reserved band is empty ink in the render.

### Reproduce

```python
from build123d import import_step
from draftwright import Sheet

[...]

PR #1290, opened 2026-08-23
Fix sparse-sheet replanning and coverage gaps

### LINKED ISSUE #443: needs repro: turned step-length chain phantom sub-mm step (GRM-03 #298 secondary)
Split out from #298 (the diameter half is fixed in PR #442). This is the **secondary symptom**: a spurious sub-millimetre step that poisons the turned step-**length** chain.

## Symptom
On a turned part with a narrow band adjacent to a larger OD (GRM-03: ø6 boss at axial 0.0–0.5, ø10 disc at 0.8–2.2), `find_turned_steps` can manufacture a phantom ~0.5 mm step. Too fine to dimension at any practical scale, it forces the whole axial chain to skip → `axial_length_missing` (the #293 cram-guard).

## Root cause
`find_turned_steps.local_od(pos)` (`recognition/turned.py:118-124`) uses `_OD_SPAN_PAD = 0.7` mm-padded band spans for **three** purposes:
1. shoulder-face detection (a chamfered shoulder at a band edge should still see its band's OD — the pad is *needed* here),
2. the step **diameter** at a segment midpoint (line ~158),
3. the phantom-drop `> 0` filter (line ~159, #279).

Because the pad (0.7) exceeds the sliver width (0.5), a sub-mm segment created by the nested band's edges acquires a neighbour's OD via the pad, survives the `> 0` filter, and enters `prof.steps` → the length chain (`render_step_lengths`) and the `axial_length_missing` count (`coverage.py`).

## Directions
- Use **unpadded** band spans when assigning a segment's *diameter* (use #2 exact), keeping the pad only for shoulder-face detection (#1) — a sliver between bands then reads OD 0 and is dropped by the `> 0` filter.
- Or **dedup shoulders closer than the pad** (merge two shoulder planes within `_OD_SPAN_PAD`), so no sub-pad segment is created.
- Or (the legibility answer) a **detail view** of the turned head when steps are below the chain's legibility floor (see #306/#56/#54) — dimension the fine features at magnification instead of skipping.

## Verification
Needs the real GRM-03 length chain — reconstruct from the `pzfreo/gramel` `thumbwheel_drive_screw` params (journal ø4, boss ø6 ×0.5, disc ø10, unthreaded ø5 ×3, M3 thread ×20) and assert no `axial_length_missing` after the fix, with the true step lengths dimensioned. The diameter half (#442) already reconstructs cleanly; the phantom step did **not** reproduce in that reconstruction, so a faithful length-chain repro is the first task here.

Follow-up to #298 / PR #442.

### LINKED ISSUE #1155: GRM-04 auto layout leaves most of A4 unused while dropping the Ø2.4 callout
## Problem

For the compact `GRM-04_drive_plate.step` part, Draftwright 0.4.2 automatically selects A4 landscape at 2:1 but uses only a small portion of the available drawing field. Despite the large unused region, placement drops an authored hole callout:

```
plan/side right strip: 1 of 2 bore callouts skipped (strip full)

callout_dropped: hole callout ø2.4 dropped from the side view
(right strip full)
```

The same output also places a `ø8` callout into a crowded dimension region while the right half of the sheet remains largely empty. This demonstrates that per-view strip capacity can declare failure without feeding back into scale, view assignment, or whole-sheet repacking.

Observed through pzfreo/draftwright-io#452 and reproduced directly with:

```python
generate_sheet_script(
    "GRM-04_drive_plate.step",
    "drawing.py",
    title="GRM 04 drive plate",
    formats=("pdf", "png"),
)
```

Automatic result: A4 landscape, 2:1, front/plan/side/iso plus detail A at 4:1.

Source used locally: `/Users/paul/steps/GRM-04_drive_plate.step`.

## Expected

An automatic plan must not drop a supported authored callout while a feasible reflow, larger standard scale, alternate eligible view, or detail assignment exists on the selected sheet. Strip-local failure must participate in the joint planning/repack decision described by #1130.

## Acceptance

- [ ] Add a GRM-04-derived synthetic or authorised fixture.
- [ ] Measure occupied-area ratio and supported requirement survival for the automatic candidate.
- [ ] Preserve the Ø2.4 callout through scale/reflow/view reassignment on A4 when feasible.
- [ ] Do not solve the case by shrinking typography or silently enlarging authored paper constraints.
- [ ] Keep every view/annotation block in bounds without new collisions.
- [ ] If no complete plan is feasible, report `plan_infeasible` with the uncovered requirement rather than exporting a completed-looking drawing.
- [ ] Link the regression into #1130's joint sheet-planning phase.

### LINKED ISSUE #1187: Dense sheets leave no clear leader route — escalate or replan, don't re-route

### LINKED ISSUE #1275: sheet.notes() cannot render U+2300 ⌀, so diameters print as □ (the ISO square symbol)
Draftwright 0.4.9. This one changes meaning rather than looking untidy, hence filing it separately from #1274.

## Symptom

Text passed to `sheet.notes([...])` containing `⌀` (U+2300 DIAMETER SIGN) renders as a missing-glyph box. On a GRM-03 drawing the notes block reads:

```
2  AXIAL PROFILE, LEFT TO RIGHT: □4 x 3.2; □6 x 0.5; TAPER □6 TO □10 OVER 0.3; □10 x 1.4.
4  □1.6 BLIND x 8.0 DEEP FROM LEFT END, COAXIAL WITH PART.
```

**`□` is the ISO 1101 symbol for a square section.** A machinist reading `□4 x 3.2` sees a 4 mm square, not a Ø4 turned diameter. Every one of those features is a plain cylinder.

## The source is correct

The script passes the right character, 19 times:

```
U+2300  DIAMETER SIGN  x19
```

```python
sheet.notes([
    'AXIAL PROFILE, LEFT TO RIGHT: ⌀4 x 3.2; ⌀6 x 0.5; TAPER ⌀6 TO ⌀10 OVER 0.3; ⌀10 x 1.4.',
    '⌀1.6 BLIND x 8.0 DEEP FROM LEFT END, COAXIAL WITH PART.',
    ...
], prefer='tr')
```

## Two code paths, one glyph

On the **same sheet**, U+2300 renders correctly outside the notes block:

- feature leader callouts render `ø4`, `ø6`, `ø10`, `ø5`, `ø3`
- the end view renders `⌀1.6 ↓ 8` correctly, diameter sign and depth symbol both

So the font used for annotations has the glyph and the font used for the notes block does not — or the notes path takes a different text pipeline that loses it.

## Why it matters

Notes are where a drawing puts what the dimensions cannot carry, so they attract exactly this kind of content — profiles, bore callouts, thread specifications. Silently turning every diameter in that block into a square symbol is the worst available failure: the drawing looks complete and reads wrong.

Nothing in lint flagged it.

## Suggested handling

Use the same font/text path for `sheet.notes()` as for feature annotations. Failing that, refuse to render a glyph the notes font lacks rather than substituting the box — a visible error beats a plausible wrong symbol.

Happy to supply the full script and the rendered PDF/PNG.

pzfreo:
## Summary

- measure sparse A4 recovery-detail sheets at the next preferred scale and re-home GRM-04 side-hole Y locations into the plan corridor when needed
- replan GRM-03 around optional ISO content until all true axial step lengths are present
- render U+2300 diameter signs in notes/tables with the supported drafting glyph while preserving source rows
- close the remaining dense leader cases as deliberate Policy-B retention, with a fast release gate alongside the exhaustive slow characterization

Closes #1155
Closes #1187
Closes #1275
Closes #443

## Verification

- repository-wide non-slow suite: 4,640 passed, 5 skipped, 48 deselected, 2 xfailed; sole expectation drift corrected and rerun
- focused issue, arrangement, and emitted-Sheet parity gates: 6 passed
- Ruff and git diff checks pass
- wheel and sdist build successfully at 0.4.11.dev0
- GRM-03 and GRM-04 PNG outputs visually inspected

The slow-marked corpus tests were intentionally not run.


src/draftwright/linting/coverage.py
```diff
@@ -418,6 +424,10 @@ def lint_location_coverage(
         if (
             HoleRef.of(h.location) not in patterned
             and not coaxial
+            and not any(
+                all(abs(actual - expected) <= tol for actual, expected in zip(point, (x, y, z)))
+                for point in structured_locations
+            )
             and not any(
                 abs(vx - px) <= tol or abs(vy - py) <= tol for vx, vy in dim_verts.get(view, ())
             )
```
pzfreo, line 488: **[confirmed — reproduced on the PR head]** The structured-fact acceptance silences `feature_not_located` axis-blind and feature-blind: it matches only `fact[2]` (the point) and discards `fact[0]`/`fact[1]` — the feature and *which axis the dim covers*. Replaying the arrangement-gate mutation: holes at (-25,15,6)/(-15,15,6) have **no X-location dimension anywhere** on the mutated sheet, yet the lint is silent because side-view dim `m_locy3` (a Y ordinate) carries `('location.location.y', point)` facts for them. `hole_coverage.py:581` consumes these same facts `(feature, parameter)`-keyed; this consumer discards the parameter.

Two compounding problems: (a) `tol=0.6` is applied in **model mm** here while every other tolerance in this function is page mm — at 5:1 (the GRM-04 population this PR targets) that's 3 page-mm, so two micro-part holes within 0.6 model-mm cross-match and an unlocated hole goes unreported; (b) the function's own docstring sells it as "derived from the drawing itself (not a build-time side channel — so it judges any producer)", and `covers_hole_locations` is precisely an engine-attached side channel — the producer now partially self-certifies the physical critique. Worse, the only remaining signal (`location_ref_dropped`) is documented to be *cleared* by the hole-table escalation (holes.py:708-711), after which a sheet stating a hole's X position nowhere passes physical critique fully clean.

Minimum fix: key the match on `(fact[0], fact[1])` like the sibling consumer, and use the compiler's stated stability (~1e-6, per model/compiled.py) or `HoleRef.of(point) == HoleRef.of(h.location)` rather than a loose per-axis tol.

tests/test_adr0018_arrangement_gate.py
```diff
@@ -151,10 +154,7 @@ def resolving(*args, **kwargs):
 
         drawing = build_drawing(_dense_plate())
         assert (drawing.page_w, drawing.page_h) == (baseline.page_w, baseline.page_h)
-        assert _lint_codes(drawing) - _lint_codes(baseline) == {
-            "location_ref_dropped",
-            "feature_not_located",
-        }
+        assert _lint_codes(drawing) - _lint_codes(baseline) == {"location_ref_dropped"}
```
pzfreo: **[confirmed]** The shrunken delta `{'location_ref_dropped'}` is real on the PR head, but the comment's justification is factually wrong: the silencing evidence is **not** "#1155 structured location evidence across view reassignment" (a `dim_loc_plan_y*` re-home) — it is the pre-existing side-view Y-ordinate dim `m_locy3`, whose fact the new coverage code now accepts cross-view. Under old code `feature_not_located` fired precisely *because* that Y dim sits outside the holes' end-on view.

The test used to require the physical critique to corroborate the compiler-side requirement loss; it now requires the opposite — it accepts "requirement lost but physical critique went blind" as the passing state, so a future over-silencing regression in the structured acceptance keeps this green.

src/draftwright/builder.py
```diff
@@ -1603,6 +1652,89 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
...
+        # the same read-back check confirms every shoulder is located.
+        original_has_axial_gap = bool(
+            latest_analysis is not None
+            and lint_axial_coverage(
+                latest_analysis.part,
+                drawing,
+                prof=latest_analysis.prof,
+            )
+        )
+        if original_has_axial_gap and _include_iso and views_are_automatic:
```
pzfreo: **[confirmed — measured on the PR head]** The #443 gate checks `_include_iso and views_are_automatic` but not `auto_dims` (unlike the #1155 block above). `build_drawing(turned_part, auto_dims=False)` now runs **two full builds** (`include_iso` `[True, False]`): a build that draws no automatic dimensions always has an axial gap, and removing the iso can never close it — the candidate is guaranteed-futile, so every dimension-free or authored-omission turned build (ADR 0016: omission means suppression) silently pays ~2× build time on every invocation. Same waste when the settled drawing already has no iso view.

Also note `lint_axial_coverage` fires at severity *info* for assemblies, which this gate doesn't distinguish.

src/draftwright/builder.py
```diff
@@ -1603,6 +1652,89 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
...
+            without_iso = _build(None, views=_views, include_iso=False)
```
pzfreo: Two robustness gaps in the #443 rebuild, both absent from the sibling #1155 loop:

1. **No exception guard.** The #1155 loop anticipates `Standard_Failure` on speculative candidates; this rebuild re-runs scale selection from scratch (reaching much larger scales per the changelog) with no guard, so a candidate-build failure aborts the whole `build_drawing` call — throwing away the working original drawing already in hand. A speculative candidate's failure should reject the candidate, not the build.
2. **No acceptance constraints.** It is accepted on `not candidate_has_axial_gap and not _automatic_blockers` alone — no same-page rule, no structural-error check. It can silently change the paper size relative to the settled result, and when it follows an accepted #1155 upscale it re-runs the estimate-driven compose that #1155 exists to correct, discarding the same-page/no-detail guarantees enforced moments earlier (a sparse no-iso A4 with a recovery detail can get its small scale back).

src/draftwright/builder.py
```diff
@@ -1612,7 +1744,21 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
             policy="automatic",
             requested=None,
             effective=drawing.scale,
-            status="automatic",
+            status="automatic_replanned" if replanned else "automatic",
+            attempted=(original_scale, drawing.scale) if replanned else (),
```
pzfreo: **[confirmed by trace]** When both replans fire in one build (#1155 accepts 2:1→5:1, then #443 replaces that with an iso-less build), the `scale_decision` record is wrong on both axes: `attempted=(original_scale, drawing.scale)` drops the accepted intermediate 5:1 entirely, and `replan_reason` holds only the *last* reason, so the 2:1 original is labeled `axial_coverage_incomplete` even though it was actually replanned away for the detail reservation. The machine-readable audit trail — the record's whole purpose per the surrounding comment — misstates both what was tried and why.

Related shape issue: building the attempts tuple at each win site (instead of the two mutable flags + the reason→status string dispatch 90 lines away) removes the failure mode where a third replan cause forgets the mapping and the fallback else-arm silently records the wrong status.

src/draftwright/builder.py
```diff
@@ -1603,6 +1652,89 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
...
+            and _occupied_rectangle_ratio(drawing) < 0.15
+        ):
+            for candidate_scale in sorted(item for item in _SCALES if item > original_scale):
+                try:
+                    candidate_drawing = _build(
+                        candidate_scale,
+                        views=_views,
+                        page_override=original_page,
+                    )
+                except Standard_Failure as exc:  # pragma: no cover - OCC/platform-specific
```
pzfreo: This handler guards the wrong exception class: `Standard_Failure` has no observed occurrence in this codebase (this is its only use, and it's `pragma: no cover`), while the engine's one precedented build-failure mode is the `ValueError` ("drawing geometry degenerates") that the sibling downscale-retry at ~1835 catches — and this loop does **not** catch it. A candidate build failing the way builds actually fail aborts the whole call, discarding the good original drawing. Suggest catching the precedented class (or both).

src/draftwright/builder.py
```diff
@@ -1603,6 +1652,89 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
...
+        if (
+            auto_dims
+            and views_are_automatic
+            and any(name.startswith("detail_") for name in drawing.views)
+            and original_page == (297.0, 210.0)
+            and _occupied_rectangle_ratio(drawing) < 0.15
+        ):
+            for candidate_scale in sorted(item for item in _SCALES if item > original_scale):
+                try:
+                    candidate_drawing = _build(
```
pzfreo, line 1747: An accepted #1155 candidate is built with `arrangements=None` and never passes through `_preserve_requirements_under_arrangement`, so `drawing = candidate_drawing` reverts `arrangement_decision` to the constructor default — the arrangement audit record the discarded original carried (chosen/rejected attempts when the initial pass departed from the preferred arrangement) is erased from the drawing that's actually returned. Requirement *loss* is covered by `_automatic_blockers`, but the decision record is not.

pzfreo (commented):
Automated adversarial review (8 parallel finders + cross-file tracing, empirical probes run against the PR head). Inline comments carry the concrete defects; two broader points belong here:

**Architectural fit (CLAUDE.md: weigh against the ADRs, amend in the same PR):**
- The #443 replan makes the engine drop the iso view on its own. ADR 0018's status header explicitly records "automatic view SELECTION … does not exist (nothing drops a view on its own)" and gates it behind the ViewSpec/ResolvedViewPlan lifecycle. This PR ships that capability as an ad-hoc builder retry, and `docs/` is untouched — the exact "silently diverging" the project rule forbids. Either amend ADR 0018 (status, reasoning, date) or reshape the change.
- The #1155 measured upscale selects scale from **built, bbox-measured geometry** (`_occupied_rectangle_ratio`), a second selection mechanism outside `choose_scale`'s candidate loop. ADR 0004 says footprints are box layouts, "never bbox-measured geometry", and its 2026-07-18 amendment permits measurement for *post-build verification* only. compose.py's own ADR 0018 §5 comment says new decision dimensions should be "an addition to the generator below, not a rewrite of the loop" — this adds the decision outside the loop, gated on a hardcoded `(297.0, 210.0)` (with `_PAGE_SIZES` already imported) and a magic 0.15 with no principled source. A sparse A3 sheet with the identical symptom silently never re-measures.

**Minor / efficiency bundle** (none blocking on their own):
- Each rejected #1155 candidate pays 1–2 redundant full `lint(physical=False)` passes (eager `structural_error` + `_automatic_blockers` recompute), and the loop never breaks on the monotone overflow failure class a pinned page guarantees — a sub-1:1 original could burn double-digit futile full builds.
- `same_page` can never be false (the candidate is built with `page_override=original_page` and `choose_scale`'s explicit-scale-plus-page branch returns it as-is) and the two `latest_analysis is not None` guards are always-true; if these are meant as invariants, assert them so drift fails loudly instead of silently vetoing the replan.
- `lint_axial_coverage` runs before the cheap `_include_iso and views_are_automatic` gates; a turned part with iso disabled or authored views pays a full drawing coverage walk whose verdict is then discarded.
- coverage.py's new fact parse accepts only the 3-tuple form; the sibling consumer `hole_coverage.py:581` also handles the documented legacy 2-tuple, so the two hole lints can disagree about identical evidence.
- The ⌀→ø substitution lands only at the table measure/render seam; `Drawing.note`/`Sheet.note`/title-block text render the same missing glyph. If the premise is "the pinned fonts lack U+2300", it wants one shared user-text normalisation choke point.

src/draftwright/annotations/holes.py
```diff
@@ -854,8 +856,54 @@ def _locate_across(dwg, ctx, a: Analysis, off):
...
+
+    def _fallback(name):
+        # Every queued side candidate is created in the same block as its plan
+        # alternate above; a missing key is an internal invariant violation.
+        alt = plan_alternates[name]
+        alt_name = alt[0]
+        if not _off_axis_emit(
+            dwg,
+            tier,
+            a.pv_zones.right,
```
pzfreo: Two issues with the plan-view fallback:

1. **No plan-view presence check.** `a.pv_zones` is built unconditionally from `_layout_geometry` (compose.py:1323), so on a restricted/authored view set that omits `plan`, `pv_zones.right` is a live Strip even though no plan view is drawn — the re-homed `dim_loc_plan_y*` lands at plan-view page coordinates as a floating dimension for a view not on the sheet, instead of the honest `off_axis_location_dropped` record.
2. **The comment oversells the solve.** It calls the shared `place_strip_candidates`, but as a solitary single-candidate emit executed synchronously inside the *(side, below)* corridor's drain — never a candidate in the (plan, right) batch's joint solve, and bypassing `ctx.post_drain` (the sanctioned deferral `drain_corridors` documents for exactly this). Depending on drain order it either consumes capacity/corners the (plan, right) joint solve was entitled to, or lands outside that ladder's crossing-free order — the same defect the `_locate_heights` comment at 1047-1075 records, and whose capacity pre-check this copy lacks. This is now the third per-pass copy of the alternate-view-on-drop pattern; ADR 0014 Amendment 2's collected late stage was built for this class.

tests/test_issue_1155_grm04_sheet_use.py
```diff
@@ -0,0 +1,74 @@
+"""#1155 — GRM-04 must use A4 and retain both Ø2.4 requirements."""
+
+from pathlib import Path
+
+from draftwright import build_drawing
+
+_FIXTURE = Path(__file__).parent / "fixtures" / "grm04_drive_plate.step"
+
+
+def _rect_union_area(rectangles):
```
pzfreo: `_rect_union_area`/`_occupied_ratio` are a ~40-line verbatim copy of `builder._occupied_rectangle_ratio` — same algorithm, same rectangle-harvesting loops. The copy buys no oracle independence (a shared bug in the sweep yields the same wrong ratio on both sides), and it drifts silently: when the production trigger's rectangle collection changes, this keeps asserting `>= 0.25` against a stale mirror while measuring something different from the `< 0.15` gate it exists to pin. Import `_occupied_rectangle_ratio` from `draftwright.builder` (this test already reaches private surfaces, e.g. `drawing.registry.measurement_of`).

pzfreo:
Addressed the review findings on final head `ed10e84`:

- location coverage now keys structured facts by exact IR feature, axis parameter, and member position; compiler-tagged dimensions cannot self-certify through geometric fallback, and legacy two-tuple facts remain supported;
- the side→plan hole-location fallback is deferred through `post_drain` and refuses a missing plan view;
- #1155 no longer uses post-build occupied-area/bbox heuristics or A4 literals; corrective trials retain the settled page/arrangement, catch fixed-page failures, lint once, and audit every attempt;
- #443 is gated by automatic dimensions/views and an actual ISO + axial gap, discovers a no-ISO scale then rebuilds it on the settled page/arrangement before acceptance, with structural/blocker/detail/coverage gates and complete audit records;
- incomplete automatic plans append to, rather than overwrite, corrective history;
- U+2300 diameter signs render through the supported `ø` glyph in tables, `Drawing.note`, declared `Sheet.note`, and title blocks while semantic source data remains unchanged;
- ADRs 0004 and 0018 now document the bounded corrective policies and the truthful `feature_not_located` expectations are restored.

Validation: focused review/regression sets passed (72 + 121 + 35 tests), the private-access ratchet passes, and `scripts/pr-check --static` passes (Ruff check/format, mypy, codespell, docs). Slow tests were not run. An independent re-review confirmed all three final edge-case findings are resolved (6 focused tests, no slow tests).

src/draftwright/recogniser_contract.py
```diff
@@ -442,16 +428,17 @@ def validate_recogniser_capabilities(
         "version": distribution_version("draftwright"),
     }:
         raise RecogniserCapabilityError(
-            "consumer identity/version is stale; update it with the Draftwright release"
+            "consumer identity/version does not match installed Draftwright metadata"
         )
+    installed_package_version = distribution_version(_RECOGNISER_DISTRIBUTION)
```
pzfreo, line 433: **[confirmed]** The version leg of the fail-closed join is now tautological. The declaration builds `"version": f"=={distribution_version('b123d-recognisers')}"` (line ~287) and this validator compares it back against the same installed metadata — installed vs installed, same call, same process. The manifest's `package.version` is also the same package's `__version__`. So no runtime or test check can detect a drifted install anymore: a venv where someone ran `pip install -U b123d-recognisers` past the pyproject pin (or `--no-deps`, or a stale lock) with unchanged family ids/schemas passes `validate_recogniser_capabilities()` at every call site — including the evaluation harness's truth gate — where pre-PR every validation failed loudly with "does not satisfy b123d-recognisers==0.3.0". Geometry recognition then differs silently from the reviewed pin.

One-line fix: compare the installed version against draftwright's own installed `Requires-Dist` pin (importlib.metadata), keeping one independent truth on each side of the join. Side effect worth knowing: `distribution_version("draftwright")` also makes the module raise `PackageNotFoundError` from a PYTHONPATH-only checkout, which the old hardcoded strings tolerated.

tests/test_recogniser_capabilities.py
```diff
@@ -121,11 +111,8 @@ def _emitter_literal_kinds() -> set[str]:
 
 def test_installed_package_contract_validates_without_a_sibling_checkout() -> None:
     distribution = importlib.metadata.distribution("b123d-recognisers")
-    assert distribution.version == EXPECTED_PACKAGE_VERSION
-    if CANDIDATE_VERSION is None:
-        assert distribution.read_text("direct_url.json") is None
-    else:
-        assert distribution.read_text("direct_url.json") is not None
+    assert distribution.version == INSTALLED_PACKAGE_VERSION
```
pzfreo, line 114: **[confirmed]** Dead assertion: `INSTALLED_PACKAGE_VERSION` is `importlib.metadata.version("b123d-recognisers")` (line 35), so this is literally `x == x`. The line it replaced compared the installed distribution against the checked-in release record — the only test that proved the environment actually contains the reviewed release. The surviving pin test in `test_external_recognition_boundary.py` cross-checks pyproject *text* against uv.lock *text* and never touches the installed environment. Fix: compare `distribution.version` to the version parsed out of pyproject's `==` pin.

src/draftwright/builder.py
```diff
@@ -1603,6 +1646,263 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
...
+                )
+            )
+            for candidate_scale in sorted(item for item in _SCALES if item > original_scale):
+                try:
+                    candidate_drawing = _build(
+                        candidate_scale,
+                        arrangements=(settled_arrangement,),
+                        page_override=original_page,
+                    )
+                except ValueError as exc:
```
pzfreo: **[confirmed]** This catch (and the two in the iso pass at ~1808/~1841) takes bare `ValueError` around the *entire* build pipeline and treats it as a monotone fixed-page fit failure. Two problems:

1. It absorbs `annotations/leaders.py`'s `_FeatureLeaderInvariantError` — a ValueError subclass the annotation layer deliberately re-raises so compiler invariant errors are never swallowed. A regression that raises only at the candidate scale gets logged at INFO as "fixed-page fit failed", breaks the ladder (so a feasible larger scale is never tried), and ships as a quietly-never-improving sheet.
2. The exception it names may not exist on this path: with both scale and page fixed, `compose.choose_scale` only *warns* on a misfit (compose.py:616-618) and returns — there is no fit ValueError to catch, so the guard exists mainly to absorb other bugs.

The pre-existing sibling loop (~1990) encodes the opposite policy explicitly: it filters ValueError by the known "drawing geometry degenerates" message and re-raises everything else, with a comment saying "Do not hide any unrelated build error." A typed fit-failure exception from compose (the mechanism's real owner), or the sibling's filter-and-reraise, is the right shape.

src/draftwright/builder.py
```diff
@@ -1589,6 +1629,9 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
     if scale is None:
         if scale_policy != "fallback":
             raise ValueError("scale_policy applies only when an explicit scale is supplied")
+        views_are_automatic = _view_constraints is None or (
```
pzfreo, line 1637: `views_are_automatic` uses `ViewConstraints.is_empty`, which goes False as soon as `principal_source` is set — and `Sheet.auto_views()` sets `principal_source="automatic"` (and is *required* before `add_view()`/`add_section_view()`). So both replans are silently disabled for any Sheet whose principal views are explicitly AUTOMATIC: the same GRM-03 part keeps iso at 2:1 with `axial_length_missing` via `Sheet(part).auto_views()...` while `build_drawing(part)`/the CLI replans to 5:1 without iso — a #707-class same-part surface divergence, and the ADR 0014 Amendment 4 pattern (a gate silently disabling the capability it protects on ordinary input). The gate should treat `principal_source == "automatic"` with no authored view entries as automatic.

src/draftwright/builder.py
```diff
@@ -1603,6 +1646,263 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
...
+
+        # #443: a pictorial view is useful context, but it cannot outrank the
+        # dimensions needed to manufacture a turned profile.  GRM-03 selected 2:1
+        # with ISO, collapsed its 0.5 + 2 mm head steps into an unowned 2.5 mm
+        # block, then had no room for the recovery detail.  Re-plan once without
+        # the optional ISO; scale selection reaches 5:1 and the two truthful step
+        # dimensions fit inline.  This is deliberately a coverage comparison, not
+        # suppression of ``axial_length_missing``: the candidate wins only after
+        # the same read-back check confirms every shoulder is located.
+        if auto_dims and _include_iso and views_are_automatic and "iso" in drawing.views:
```
pzfreo: **[confirmed]** Both replan gates test the `auto_dims` kwarg, which stays `True` for authored-dimension builds — `Sheet.authored_dimensions()` carries the set on `model.authored_dimensions` and never touches `auto_dims`. That contradicts the ADR 0018 Amendment 3 text this PR adds ("both views and dimensions are automatic"), and the guard test only exercises `auto_dims=False`.

Consequence: an authored turned Sheet that deliberately suppresses a step length (ADR 0016: omission = suppression; `lint_axial_coverage` is documented to still fire on authored suppression, #955) deterministically triggers this replan on every build — two extra full OCC builds from the *same* authored model, guaranteed rejection ("axial_coverage_incomplete"), plus noise in `scale_decision.attempts`. Pure cost, no possible win. The gate should also require `model.authored_dimensions is None`, and the guard test needs an authored-set sibling.

src/draftwright/builder.py
```diff
@@ -1603,6 +1646,263 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
...
+            replan_attempts.append(
+                _scale_attempt(
+                    drawing.scale,
+                    "detail_reservation_conservative",
+                    reason="measured_upscale",
+                    views=drawing.views,
+                    page=original_page,
+                )
+            )
+            for candidate_scale in sorted(item for item in _SCALES if item > original_scale):
```
pzfreo: **[confirmed]** The upscale ladder has no work budget, and its intended natural bound is dead code: the only early exit for oversized scales is the ValueError break, but with scale and page both fixed `choose_scale` never raises on a misfit (it warns and returns), and the one downstream explicit-scale raise fires only for too-*small* scales. So an oversized candidate completes a **full build**, is rejected via lint, and the loop continues to the next larger scale.

On ordinary dense parts whose recovery detail is genuinely needed — the common case the reservation exists for — every `build_drawing` pays 2-3 extra rejected full build+lint cycles (settled 2:1 or 1:1), and a reduction-scale part with a needed detail can walk 6+ rungs. ADR 0014 Amendment 4 (cited in this repo's CLAUDE.md) records exactly this class: a budget must bound measured work. A cap on the lookahead (next one or two preferred scales) or a live counter would bound the loss; the attempts are at least recorded in `scale_decision`, so the work is auditable after the fact.

src/draftwright/builder.py
```diff
@@ -1603,6 +1646,263 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
...
+                        _scale_attempt(
+                            without_iso_proposal.scale,
+                            "scale_proposal",
+                            reason="remove_optional_iso",
+                            views=without_iso_proposal.views,
+                            page=proposal_page,
+                        )
+                    )
+                    try:
+                        without_iso = _build(
```
pzfreo: Efficiency: the iso replan always runs this second full build at fixed page even when the proposal build already landed on the original page — for the GRM-03 target case itself (`proposal_page == original_page`, same arrangement, same fixed scale it just chose) that's a deterministic byte-equivalent second full OCC pass. `if proposal_page == original_page: without_iso = without_iso_proposal` skips it.

Two related redundancies: the full `_automatic_assessment` lint runs before the cheap `has_detail`/axial-gap gates that can reject alone (both passes — evaluate the cheap gates first); and an accepted winner is fully re-linted from scratch by `_complete_automatic_plan` despite `_scale_blockers_from_issues` existing precisely to share one lint pass — thread the winner's already-materialised issue tuple through. Together that's roughly one wasted full build plus two wasted full lint passes per successful replan.

src/draftwright/builder.py
```diff
@@ -1603,6 +1646,263 @@ def scale_blockers_for(built: Drawing) -> tuple[dict, ...]:
...
+        # those larger candidates only when the settled result actually contains
+        # that semantic recovery artifact: post-build occupied rectangles are not
+        # a scale-selection input.  A candidate may win only on the same sheet and
+        # settled arrangement, with no recovery detail or required placement loss.
+        if (
+            auto_dims
+            and views_are_automatic
+            and any(name.startswith("detail_") for name in drawing.views)
+        ):
+            replan_attempts.append(
```
pzfreo: The two replan passes are ~220-260 lines of near-copy-paste: ten near-identical `_scale_attempt(...)` blocks repeating `reason=`/`views=`/`page=original_page`, plus the same build → retain → assert-page → assess → reject-reason cascade duplicated with one extra condition — and the third/fourth hand-rolled copies of the trial-build loop the explicit-scale fallback already has (which the copies already disagree with subtly: `scale_blockers_for` vs `lint(physical=False)`, different exception policy). The `any(name.startswith("detail_") ...)` predicate now appears four times (1686, 1742, 1861, projection.py:490) — a rename that misses one site makes the upscale gate silently never fire.

ADR 0018's own rationale names the next corrective case (#1190's section), which will be copied again. A local `record(...)` closure, one shared qualify-candidate helper, and a tiny `has_detail_view(views)` would collapse this before pass #3 arrives.

src/draftwright/linting/coverage.py
```diff
@@ -398,7 +403,36 @@ def lint_location_coverage(
...
+                elif len(fact) == 2:  # compatibility with legacy/external structured facts
+                    measurement, point = fact
+                    feature = getattr(measurement, "feature", None)
+                    parameter = getattr(measurement, "parameter", None)
+                else:
+                    continue
+                if getattr(feature, "kind", None) == "hole" and parameter is not None:
+                    structured_locations.add((feature, str(parameter), HoleRef.of(point)))
+
+    model_features = tuple(getattr(dwg.model(), "features", ()))
```
pzfreo: Two contract regressions against this function's own (rewritten) docstring, which still promises the duck-type contract is `at()`/`iter_annotations()`/`view_of()` and that external producers are judged from placed witness geometry:

1. `dwg.model()` is now called unconditionally — an external producer's drawing implementing exactly the documented contract crashes with `AttributeError` on any part with a hole, instead of being judged. (`lint_location_coverage` is exported public API.) A guarded `getattr` restores the contract.
2. A dimension carrying *any* `covers_hole_locations` fact is wholly removed from the geometric witness pool (the `if not facts:` at ~410), so a legacy/external fact shape the 412-421 parser drops (2-tuple whose measurement lacks `.parameter`, or any other arity) loses both semantic and geometric standing — a false `feature_not_located` for a hole its printed dimension truthfully locates. Engine paths are safe (verified: `render_locations`/`_locate_across` accumulate facts for every collapsed hole), so the fix is just falling back to witness geometry when no fact parses.

src/draftwright/linting/coverage.py
```diff
@@ -398,7 +403,36 @@ def lint_location_coverage(
...
+
+    def _feature_for(hole):
+        ref = HoleRef.of(hole.location)
```
pzfreo: Efficiency: `_feature_for` rescans every model feature and rebuilds `{HoleRef.of(point) for point in members}` per (hole, feature) pair, and is called eagerly for every hole — before the patterned/coaxial exemptions that discard most holes, and even when `structured_locations` is empty (any external drawing), where its result cannot matter. On a pattern-heavy plate that's thousands of set constructions per lint pass, and this lint runs on every physical lint/export. Build one `dict[(axis, HoleRef) -> Feature]` before the hole loop and look up lazily after the exemptions.

Also: the fact-normalisation block at 412-421 is a verbatim copy of `hole_coverage.py:581-588` (down to the comment), with a third partial variant in `annotations/_common.py:131` — one shared decoder would keep the three from drifting.

pzfreo (commented):
Second-round review of the updated head (`ed10e84`) — 8 finder angles plus dedicated verification agents (~20 candidates checked; refuted along the way: a `_views`-dropped claim, an analysis-staleness claim, a greedy-retry ADR 0014 claim, and a cross-hole fact-exclusion claim on engine paths — the `_build` views default and the fact-accumulating dedup paths are sound).

**Addressed since round 1 (credit where due):** ADR 0004/0018 amendments now ship in the PR; the plan-view-absent branch was added to the holes fallback; exception guards were added to the candidate loops; the arrangement/`_views` threading is correct.

**The one review-blocking process issue — undisclosed scope:** commit `d29639e` ("chore: simplify recogniser dependency integration") deletes the entire recogniser-dependency delivery machinery — bump workflow, check/update scripts, `recogniser-release.json`, workflow docs, ~370 test lines, 1,176 lines across 23 files — none of which traces to #1155/#1187/#1275/#443, and the PR body never mentions it. Two enforced invariants have no replacement: the `RELEASE_HASHES` assertion that cross-checked uv.lock's sha256 artifact hashes against an independently checked-in record (uv.lock is now self-certifying), and the review gate for major/minor pin bumps. Verified consequence (see the inline `recogniser_contract.py` comment): the "fail-closed" version join is now tautological. This removal deserves its own PR, or at minimum explicit disclosure and sign-off in this one.

**Minor notes** (no inline comment): `coverage.py` ~471 states the same 1.0 mm threshold twice (`coaxial` == `not required_axes`, making one disjunct dead — one constant, please); `_complete_automatic_plan`'s `getattr`/`hasattr` guards exist only for a test's SimpleNamespace fake — fix the fake instead; user-supplied `label=` overrides via `place_dim`/`dimension(...)` still reach Plex Mono with raw U+2300 while every other seam now substitutes ø; the covers_hole_locations fact decoder now has three copies (coverage.py, hole_coverage.py:581, _common.py:131) — worth one shared helper before they drift.

Issue #736, opened 2026-07-19
#735 child: solve-trace / explain mode for strip placement


pzfreo:
Child of #735 (1). A debug flag (env var or build_drawing(trace=...)) that records, per strip solve: the candidate set (name, priority, size, force, anchored), the carved free segments, the obstacles that carved them (by annotation name), each drop with its reason, and the final positions — dumped as JSON beside the drawing. Zero output change. Also enrich the placement_unsatisfiable lint message to name what filled the strip (the #733 diagnosis took a custom script + two slow CTC builds; this makes it a glance). Wire the same data into Escalation where cheap.

PR #1294, opened 2026-08-23
Fix DRAFTWRIGHT_TRACE ownership after corrective replans

pzfreo:
## Problem

The post-merge slow tier has been failing on `main` since the #1290 merge:

```
FAILED tests/test_issue_798_silhouette_lint.py::...[nist_ctc_05_asme1_ap242-18]
AssertionError: the leader floor placed 16 callouts, not 18
```

#1290's automatic branch may now run corrective replan candidates (the #1155 measured-upscale ladder and the #443 remove-optional-ISO replan). Each candidate is a full `_build_drawing_once` build, and **every build resolves `DRAFTWRIGHT_TRACE` to the same path and atomically rewrites it at build end**. When a candidate is rejected, the file on disk describes the *rejected* build rather than the drawing the caller receives.

Per-build probe on `nist_ctc_05_asme1_ap242` (current `main`): the ladder runs 3 builds placing **18** (0.2, shipped) / 19 (0.5, rejected) / **16** (1.0, rejected) — and the last writer wins, so the trace file says 16. The #798 cardinality canary reads the file, hence 16 ≠ 18. (CI saw 17 before the recognisers-0.3.1 pin, 16 after — same mechanism, shifted inventory.)

## Fix

After the replan loop settles, give the settled drawing's own tracer the last write — the same reason a successful `finalize` re-writes (#736). One guarded call in `builder.py`:

- candidate **accepted** → its tracer re-writes (idempotent, no behaviour change);
- all candidates **rejected** → the original's tracer reclaims the file;
- **no replan fired** → no-op.

The explicit-scale fallback loop is untouched: there the accepted candidate is always the last build, so its final write is already correct.

## Verification

- The failing canary now passes: `nist_ctc_05` trace records 18 (the shipped build), `nist_ctc_04` unaffected (single build, 11 before and after; my probes confirm 11 is the truthful shipped value on both sides of #1290).
- Probe evidence: builds place [18] / [19] / [16,16]; shipped scale 0.2; with the fix the final file matches the shipped build.
- Lint quartet clean (ruff check, ruff format, mypy, codespell); full fast tier run locally.

## Not addressed here (separate fallout)

The #1293 recognisers-0.3.1 pin introduced two additional slow failures that are unrelated to trace ownership and pre-date nothing in this PR: `test_issue_1187_unroutable_leaders` (`m_pocket_yz3` no longer cuts) and `test_issue_1219_location_claims_its_own_axis` (fixture compiles no slot position). Those need their own triage against the recogniser behaviour change.

### DIFF
--- a/src/draftwright/builder.py
+++ b/src/draftwright/builder.py
@@ -1911,6 +1911,14 @@ def build_drawing(
             attempted=tuple(item["scale"] for item in replan_attempts),
             attempts=replan_attempts,
         )
+        if replan_attempts and drawing.solve_trace is not None:
+            # Every corrective candidate was a full build, and each build writes the
+            # one shared trace path, so the file on disk may describe a *rejected*
+            # candidate rather than the drawing returned.  The settled drawing's own
+            # recorder holds the shipped build's records; give it the last write so
+            # DRAFTWRIGHT_TRACE always describes the drawing the caller receives
+            # (#736 — the same reason a successful finalize re-writes).
+            drawing.solve_trace.write()
         return _complete_automatic_plan(drawing, issues=settled_issues)
 
     requested_scale = float(scale)
--- a/tests/test_issue_1187_unroutable_leaders.py
+++ b/tests/test_issue_1187_unroutable_leaders.py
@@ -124,7 +124,6 @@ def _clear_routes(dwg, name, *, directions=64, reaches=(0.6, 0.8, 1.0, 1.3, 1.7,
 @pytest.mark.parametrize(
     ("fixture", "name"),
     [
-        ("nist_ctc_05_asme1_ap242", "m_pocket_yz3"),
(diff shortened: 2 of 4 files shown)


pzfreo:
Follow-up on recognisers 0.3.1 fallout (supersedes the final “Not addressed here” section of the PR body):

- Removed `m_pocket_yz3` from the #1187 retained-crossing canary because its measured material cut is now `0.0`; it has a clear route and is no longer a Policy-B exception.
- Removed the vacuous CTC-02 slot-position fixture assertion because 0.3.1 intentionally recognises no slot there; the synthetic fast #1219 cases still non-vacuously guard slot claim ownership.
- Added a fast unit regression proving rejected corrective builds cannot leave their trace as the shared final trace.

Focused result: 11 passed, 3 slow cases deselected. Static preflight passed.

Issue #1300, opened 2026-08-23
Plan: make PMI-enabled GRM-03 truthful and manufacturing-complete


pzfreo:
## Outcome

`build_drawing(GRM03_AP242, pmi="annotate")` must produce a drawing whose imported dimensions, threads, tapped-hole details and knurl are accounted for, whose annotation geometry agrees with its labels, and whose five axial stations remain manufacturable on the selected sheet.

Real source: `GRM-03_thumbwheel_drive_screw_AP242_PMI.step`, SHA-256 `4b6462b9cc9f0d419250933bd77fb305f9cfebb7ec2b3f377008732876010a21`.

## Work items

- #1209 — linear PMI relationship direction / truthful witnesses (existing issue, now has GRM-03 evidence).
- #1296 — single-cylinder-face diameter PMI axis, canonical ownership and rendering.
- #1297 — fail-closed semantic manufacturing-requirement inventory.
- #1298 — typed external-thread, tapped-hole and knurl lowering/rendering.
- #1299 — bounded automatic sheet escalation for complete axial coverage.

### 0. Freeze the evidence

Add the exact file as a test fixture if redistribution is permitted; otherwise add minimal AP242 fixtures for the same entity patterns and keep a locally runnable real-file integration probe keyed by SHA-256. Record the current extraction report, model features, stage summary, lint codes and selected scale/page so every later slice proves a specific improvement.

### 1. Make source accounting fail closed — #1297

Extend `PmiSourceEntity`/the report taxonomy and Part 21 overlay to inventory semantic manufacturing requirements independently of whether they can yet lower. Preserve source entity IDs, full authoritative text and associated representation item/shape aspect IDs. Land this first: later lowering must consume one source-accounting seam, and unsupported requirements should become visible immediately.

Gate: the GRM-03 report inventories external thread, internal thread, knurl and surface texture, with no silent category loss.

### 2. Correct reference geometry semantics — #1209 and #1296

Refactor extraction around per-reference geometry facts (surface kind/axis, centroid, bbox) rather than deriving semantic direction from one merged envelope.

- Linear: select the measurement direction from the relationship between reference stations, retaining bbox only for witness support.
- Diameter/radius: accept one cylindrical reference and obtain its axis from cylindrical topology.
- Correlate imported size requirements with canonical feature/dimension owners and attach source provenance; do not draw a second callout when the automatic one already expresses the requirement.

[...]

Issue #1298, opened 2026-08-23
Lower AP242 thread, tapped-hole and knurl requirements into typed drawing intent


pzfreo:
## Summary

Once AP242 semantic manufacturing requirements are inventoried, Draftwright needs to lower geometry-associated thread/tap and knurl requirements into the existing typed model and render them through the shared annotation solve. They must not be emitted as raw page-coordinate notes or duplicate unrelated automatic annotations.

Real case: `GRM-03_thumbwheel_drive_screw_AP242_PMI.step`, SHA-256 `4b6462b9cc9f0d419250933bd77fb305f9cfebb7ec2b3f377008732876010a21`.

## Required mappings in GRM-03

- External `M3 x 0.5-6g RH, full available length` is associated with the nominal ⌀3 X-axis shaft region.
- Internal `M2 x 0.4-6H RH, 6 MIN FULL THREAD` plus ⌀1.6 × 8 drilling details is associated with the blind axial hole.
- Straight knurl, 1.0 pitch, full width between C0.3 chamfers, ⌀10 maximum after knurling, is associated with the head's nominal ⌀10 region.

The imported B-rep deliberately keeps thread and knurl envelopes smooth; these requirements cannot be reconstructed from recognition alone.

## Design constraints

- Correlate at the model/lowering seam using the source geometry association and canonical feature identity.
- Reuse or extend typed feature/aspect concepts (`HoleFeature.thread`, turned-step thread/knurl intent, decorations) rather than preserving these as opaque raw PMI strings indefinitely.
- Render through existing feature-callout candidates and ADR 0014 placement. Never hand-place the imported callouts.
- One requirement has one owner. If an automatic callout already expresses the same parameter, enrich its label/provenance rather than duplicate it.
- Preserve full source text/provenance for round-trip and diagnostics even when a typed subset drives rendering.

## Acceptance

- GRM-03 renders truthful callouts for the M3 external thread, M2 tapped hole/drill requirement, and straight knurl.
- Each callout is owned by the associated canonical feature and claims its imported source identity.
- No duplicate plain-diameter callout competes with a richer thread/knurl callout for the same requirement.
- Unsupported/conflicting associations remain explicit lowering failures; they never attach by value-only guessing.
- `drop(feature)`, emitted Sheet scripts, and declared-vs-detected builds preserve the typed requirements and provenance.
- Lint/stage summary reports each imported source as rendered or specifically dropped.
[...]

PR #1301, opened 2026-08-23
Inventory AP242 semantic manufacturing requirements

### LINKED ISSUE #1297: AP242 semantic manufacturing requirements are absent from the PMI source inventory
## Summary

`extract_pmi_report()` can currently inventory only `dimension`, `geometric_tolerance`, and `datum`. Semantic AP242 manufacturing requirements represented as shape aspects/descriptive representation items are not inventoried at all, so `pmi="report"` and `pmi="annotate"` can appear complete while material source intent is silently absent.

Observed on `main` (`36c6cd05`, 0.4.12.dev0) with `GRM-03_thumbwheel_drive_screw_AP242_PMI.step`, SHA-256 `4b6462b9cc9f0d419250933bd77fb305f9cfebb7ec2b3f377008732876010a21`.

## Source intent currently invisible to Draftwright

The Part 21 file contains semantic `SHAPE_ASPECT` / `GEOMETRIC_ITEM_SPECIFIC_USAGE` links and authoritative `DESCRIPTIVE_REPRESENTATION_ITEM` values for at least:

- external thread: `M3 x 0.5-6g RH, full available length on nominal DIA 3 region`;
- internal thread: `M2 x 0.4-6H RH, 6 mm minimum full thread; DIA 1.6 tapping drill x 8 mm full-diameter depth; conventional 118 degree drill point`;
- knurl: `Straight knurl, 1.0 mm pitch, full width between C0.3 chamfers, DIA 10 mm maximum after knurling; cut or formed process permitted`;
- surface texture: `Ra 3.2 um unless otherwise specified`;
- material, general-tolerance and chamfer requirements.

The current extraction report instead contains only 18 dimension sources (10 semantic dimensions and 8 presentation-only dimensions). None of the requirements above appears as extracted, partially extracted, unsupported, or failed.

## Why this is a trust defect

The source-inventory stage is the fail-closed boundary for imported manufacturing intent. A category the extractor never inventories cannot be protected by lowering/rendering diagnostics, and the stage summary overstates completeness by omission.

## Acceptance

- Extend the source inventory with a stable category/identity for semantic manufacturing requirements and notes.
- Inventory source requirements even before every subtype can be lowered; unsupported entries must have an explicit `not_extracted`/unsupported outcome and reason.
- Read authoritative semantic text from the Part 21 representation rather than OCR/tessellated presentation glyphs.
- Preserve the shape-aspect/geometry association needed by downstream typed lowering.
- The GRM-03 report names the external thread, internal thread, knurl and surface-texture requirements at minimum.
- PMI stage summaries and lint account for inventoried-but-unlowered requirements instead of silently passing.
- Add a compact AP242 fixture covering one associated manufacturing requirement and one unassociated general requirement.

Typed lowering/rendering is deliberately tracked separately so source accounting can land fail-closed first.

pzfreo:
## Summary

- inventory Part21 `manufacturing requirement` property chains with stable source identities and authoritative descriptive text
- preserve semantic-callout → shape-aspect → representation-item associations through `PmiRecord`, raw typed IR, and emitted Sheet scripts
- fail closed on malformed/unreadable requirement graphs while avoiding a full Part21 parse for ordinary STEP files with no requirement marker

Closes #1297. Part of #1300.

## GRM-03 evidence

Tested against `/Users/paul/steps/GRM-03_thumbwheel_drive_screw_AP242_PMI.step` (SHA-256 `4b6462b9cc9f0d419250933bd77fb305f9cfebb7ec2b3f377008732876010a21`).

The source census changes from 18 dimension entities to 26 accounted sources:

- 10 extracted semantic dimensions
- 8 presentation-only dimensions
- 8 extracted manufacturing requirements

The extracted requirements are external thread, internal thread, knurl, surface texture, general tolerances, datum scheme, chamfers, and model-representation intent. External/internal thread, knurl, texture, general tolerances, and chamfers retain their source shape-aspect/reference-item associations.

`pmi="report"` now reports the eight requirements as explicit raw typed IR and emits eight informational `pmi_not_lowered` outcomes. `pmi="annotate"` raises those outcomes to errors until #1298 supplies typed lowering/rendering. They no longer vanish before the source denominator.

## Tests

- `scripts/pr-check --static`
- `uv run pytest -q tests/test_pmi_part21.py` — 29 passed
- `uv run pytest -q tests/test_pmi.py` — 55 passed
- `uv run pytest -q tests/test_sheet_emit.py` — 392 passed, 1 deselected, 2 xfailed
- exact GRM-03 `off` / `report` / `annotate` census and lint probe

No slow-marked tests were run locally.

### DIFF
--- a/src/draftwright/_pmi_part21.py
+++ b/src/draftwright/_pmi_part21.py
@@ -1,13 +1,15 @@
 """Structured Part21 facts that OCCT's XCAF PMI transfer currently drops.
 
 This is deliberately a leaf adapter: it understands ISO 10303-21 entities and units, but
-knows nothing about XCAF labels, drafting IR, or placement.  ``pmi.py`` owns the overlay and
-may use a fact only after :func:`match_geometric_tolerance` proves exact correspondence.
+knows nothing about XCAF labels, drafting IR, or placement.  ``pmi.py`` owns the overlay.
+Geometric-tolerance facts require exact XCAF correspondence; manufacturing-requirement facts
+retain their own complete property -> representation -> shape-aspect source chain.
 """
 
 from __future__ import annotations
 
 import math
+import re
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -53,6 +55,13 @@ _SI_METRE_TO_MM: dict[str, float] = {
     ".ATTO.": 1e-15,
 }
 
+# Entity names are invariant ASCII Part21 syntax; string parameters are not.  Keying this
+# guard on the literal category text silently missed valid ``\X2\...\X0\`` encodings before
+# steputils had a chance to decode them.  A file with no property definitions cannot carry
+# the requirement chain, while any file that does is parsed conservatively (#1297).  Match
+# only the invariant token: Part 21 comments may legally separate it from the opening ``(``.
+_PROPERTY_DEFINITION_MARKER = re.compile(rb"\bPROPERTY_DEFINITION\b", re.IGNORECASE)
+
 
 @dataclass(frozen=True)
 class GeometricToleranceFact:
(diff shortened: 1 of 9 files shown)


pzfreo:
Independent review of the first revision found four fail-closed gaps. Commit `c3f256f` fixes all four:

- the cheap guard keys on the invariant `PROPERTY_DEFINITION` entity token, so escaped Part21 category strings cannot disappear;
- completeness is validated per matched callout and per shape aspect, not after flattening the aggregate;
- dangling representation references retain their own source identity/reason without hiding valid siblings;
- the Part21-only projection is collected before XCAF, so source IDs/text survive unavailable GDT support or read/transfer failure alongside the global error.

Added focused regressions for each case. Updated full `tests/test_pmi.py`: 56 passed. Static remains green; exact GRM-03 still reports all eight manufacturing requirements as fully extracted.

pzfreo:
Final independent-review edge fixed in `37eb159`: the cheap Part 21 guard now keys only on the invariant `PROPERTY_DEFINITION` token, so a valid comment between the entity name and `(` cannot suppress manufacturing-requirement inventory. Added the exact regression.

Revalidated:
- `tests/test_pmi_part21.py`: 33 passed
- `scripts/pr-check --static`: passed
- exact GRM-03: 26 sources (10 extracted dimensions, 8 presentation-only dimensions, 8 extracted manufacturing requirements); no extraction error; every requirement has a blank reason

Issue #1299, opened 2026-08-23
Automatic axial-coverage recovery cannot escalate the sheet after same-page replan fails


pzfreo:
## Summary

The optional-ISO corrective pass introduced for #443 is confined to the initially selected page. A larger five-step turned part therefore returns an incomplete A4 2:1 drawing even though a complete no-ISO A3 candidate exists.

Observed on `main` (`36c6cd05`, 0.4.12.dev0) with `GRM-03_thumbwheel_drive_screw_AP242_PMI.step`, SHA-256 `4b6462b9cc9f0d419250933bd77fb305f9cfebb7ec2b3f377008732876010a21`.

This reproduces with `pmi="off"`; it is independent of PMI rendering.

## Evidence

The AP242 model is materially different from the older AP203 GRM-03 fixture:

- bbox X: -3.2 to 25.5 mm instead of 0 to 23.5 mm;
- five axial steps instead of four, adding a 3.2 mm-long ⌀4 region;
- terminal ⌀3 region is 20 mm instead of 18 mm;
- blind ⌀1.6 bore is 8 mm deep instead of 3.8 mm.

Automatic result:

```
page=A4, scale=2:1, views=front/plan/side/iso
step labels: 3.2, 2.5, 3, 20
lint: axial_length_missing (5 steps, only 3 step lengths)
```

The no-ISO same-page proposal remains A4 2:1 and is rejected for `axial_coverage_incomplete`, after which the original incomplete drawing is returned.

A no-ISO A3 candidate at either 4:1 or 5:1 renders all five truthful step labels (`3.2, 0.5, 2, 3, 20`) with no axial-coverage issue.

## Expected

Page preference is subordinate to manufacturing completeness. When removing an optional view cannot produce complete axial coverage on the original page, automatic recovery should evaluate the next standard sheet size rather than return a known-incomplete drawing.

## Acceptance

- After the fixed-page no-ISO candidate fails axial coverage, evaluate bounded larger standard pages under the settled arrangement and automatic view policy.
- Accept only a candidate with complete axial shoulder coverage and no required placement drops or structural errors.
- Keep the existing A4 5:1 result for the older four-step GRM-03 fixture; do not escalate drawings that already have a complete same-page solution.
- The AP242 five-step case chooses a complete no-ISO A3 candidate; exact preferred scale follows the existing scale/page policy rather than being fixture-hardcoded.
- `scale_decision.attempts` records page as well as scale and gives a distinct page-escalation reason.
[...]

PR #1303, opened 2026-08-23
Fix linear PMI reference direction and witnesses

### LINKED ISSUE #1209: CTC-04 linear PMI dimensions render ~90% too long
Found by adversarial review of #1207. **Not introduced there** — identical on `main`, and
outside #1177's scope, which is about dimension *categories* the linear renderer cannot
express. These are linear records rendered by the linear renderer, simply wrong.

## Problem

Building NIST CTC-04 AP242 with `pmi="annotate"` reports:

```
Dim '20 ±0.2': label value 20.000 differs from measured path length 52.000
               (÷0.2 = 260.000) by 92.3% — possible axis swap or wrong endpoint
Dim '25 ±0.2': label value 25.000 differs from measured path length 52.000
               (÷0.2 = 260.000) by 90.4% — possible axis swap or wrong endpoint
Dim '75':      label value 75.000 differs from measured path length 17.800
               (÷0.2 = 89.000) by 15.7%
```

Two records with different values (20 and 25) both draw the **same** 260 mm span, which
suggests the witness endpoints are taken from something other than the record's own
reference points — the "wrong endpoint" half of lint's own suggestion.

## Why it matters

This is the #1153 class: **lint detects the contradiction and the drawing exports
anyway.** The label asserts 20 mm, the geometry asserts 260 mm, and a reader has no way to
know which is right.

It also undermines a guard added in #1207. That PR introduces a fail-closed partition over
`AUTHORED_DIMENSION_KINDS`, classifying `linear` as a kind the linear renderer states
truthfully — true of the *category* (a linear value is a straight projected span) but not
of today's rendering on this fixture. The classification is right; the renderer needs to
earn it.

## Related

* #1208 — `radius` renders a full-diameter-length line. Same shape: correct category,
  wrong span.
* #1153 — GRM-04 labels a 27 mm path as `4`; lint catches it, the PDF ships it.
* #1207 — refuses `angular`/`curve_length`/`curved_dist`/`oriented`, which are wrong for a
  different reason (the renderer cannot express their measurement basis at all).

## Expected

A linear PMI record draws a span equal to its value at the drawing scale, or the omission
is reported and the record not drawn — never a line whose length contradicts its own label.

pzfreo:
Closes #1209

## Outcome

Linear AP242 dimensions now derive semantic measurement direction and witness stations from the two authored XCAF reference groups, rather than the largest extent of their merged face envelope.

- group stations are segmentation-stable merged-envelope centroids;
- only a principal-axis relationship whose station span agrees with the nominal is renderable;
- oblique, missing-side, coincident, or contradictory relationships remain typed and source-owned but fail closed with an explicit geometry reason;
- geometry rendering blockers are typed separately from canonical-correlation/lowering blockers, preserving #1116 standalone fallback dimensions;
- the combined bbox remains transverse witness support only;
- short truthful dimensions use external-arrow/text layout instead of disappearing below an arbitrary 3 page-mm threshold;
- AP242 thickness records use the same conservative two-group proof and otherwise fail closed.

Both blocker classes, one-sided references, typed intent, and source IDs survive emitted Sheet-script round-trip. No nominal-only correspondence or raw annotation coordinates are introduced.

## Real-file evidence

Exact GRM-03 source:
`/Users/paul/steps/GRM-03_thumbwheel_drive_screw_AP242_PMI.step`
SHA-256 `4b6462b9cc9f0d419250933bd77fb305f9cfebb7ec2b3f377008732876010a21`

All five axial values resolve from independent authored reference groups:

| value | axis | station span | blockers |
| ---: | :---: | ---: | :---: |
| 3.2 | X | 3.2 | none |
| 0.5 | X | 0.5 | none |
| 2.0 | X | 2.0 | none |
| 3.0 | X | 3.0 | none |
| 20.0 | X | 20.0 | none |

An annotate build produces no `label_vs_measured`. Four of five currently place on A4; the remaining source is a structured `pmi_dropped` capacity outcome tracked by #1299.

NIST CTC-04 now:

- renders the 75 mm record from Y=285 to Y=210, not the old 89 mm bbox span;
- preserves and renders unmatched hole-correlation fallbacks :23, :24, :25, and :28;
- omits only the oblique :26 and one-sided :29 relationships with exact typed reasons;
- produces no `label_vs_measured`.

NIST CTC-03 source :47 no longer draws a roughly 383 mm path labelled 0.8; its one-sided thickness relationship is explicitly refused and produces no `label_vs_measured`.

## Independent review

The first review found two P1 regressions: correlation blockers suppressed valid fallbacks, and thickness lacked proven endpoints. Both are fixed by the typed rendering/lowering blocker split and the thickness station proof. Re-review is requested on current head.

## Validation

- relevant PMI/render/round-trip selection: 567 passed, 2 deselected, 2 xfailed;
- platform canary selection: 203 passed;
- `scripts/pr-check --static`: passed;
- Ruff and diff check: passed;
- exact GRM-03, CTC-04, and CTC-03 extraction/render probes: passed as described above.

The repository's slow-marked suite was not run.

### DIFF
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -5088,7 +5088,7 @@ def _authored_with_usable_references(record) -> bool:
         record.kind == "authored_dimension"
         and record.value > 0
         and len(record.ref_pts) >= 2
-        and not getattr(record, "lowering_blockers", ())
+        and not getattr(record, "rendering_blockers", ())
     )
 
 
@@ -5099,18 +5099,18 @@ def _blocked_authored_dimension_records(records):
         for record in records
         if record.kind == "authored_dimension"
         and record.value > 0
-        and getattr(record, "lowering_blockers", ())
+        and getattr(record, "rendering_blockers", ())
     ]
 
 
 def _record_blocked_authored_dimension(ctx, rec):
     source_id = getattr(rec, "source_id", "")
-    reasons = "; ".join(rec.lowering_blockers)
+    reasons = "; ".join(rec.rendering_blockers)
     ctx.record_issue(
         "error" if source_id else "warning",
         "authored_dim_source_unresolved",
         f"authored dimension {getattr(rec, 'label', '')!r} is not drawn because its source "
-        f"intent is unresolved: {reasons}",
+        f"reference geometry is unresolved: {reasons}",
         source=source_id,
         outcome_stage="validation",
     )
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -4979,7 +4979,7 @@ def _record_pmi_drop(ctx, dwg, ax, label, rec):
 
 def _record_pmi_unrenderable(dwg, label, rec, *, ctx):
     """Record an authored dimension whose reference geometry can't form a witness (fewer
-    than two distinct reference points, or a sub-legible span). Distinct from
+    than two distinct reference points, or a zero span). Distinct from
     ``pmi_dropped`` (a *placement* failure): this is a *validation* failure, so a caller
     sees a specific reason instead of a misleading "no room" — an authored dim is only
     ``pmi_dropped`` after a real candidate reaches the corridor solver and cannot fit (#562)."""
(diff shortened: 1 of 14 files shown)


pzfreo:
Independent re-review is clean on `56fb566`. The review covered the typed lowering/rendering blocker split, CTC-04 fallback preservation, CTC-03 thickness refusal, partial-reference fail-closed behavior, emitted-Sheet round-trip, and omission/quality accounting. No findings remain.

pzfreo:
The current-head CI ratchet failure is fixed in `2b79748`: `authored_dim_source_unresolved` is now explicitly classified as an unscored omission, consistent with the adjacent degenerate/unsupported authored-dimension outcomes. Local evidence on this head: the targeted quality classification test (8 cases), static checks, the 203-test smoke/unit canary, and diff-check all pass. Slow tests remain intentionally skipped.

Issue #1310, opened 2026-08-23
Tests build a whole drawing to read .model(), which detect_part_model returns 80x faster


pzfreo:
Part of #1307.

## Problem

Many tests need a `PartModel` and get one by building an entire drawing and
throwing the drawing away.

`builder.detect_part_model(part)` returns the same model directly:

| | |
|---|---|
| `build_drawing(part, title="T", number="N").model()` | **0.83 s** |
| `detect_part_model(part)` | **0.01 s** |

~80×, on `counterbored_plate` from `test_issue_1215`.

The two results are equivalent for this purpose. Across all 12 parts in that
module's `_PARTS` table, `tuple(a.features) == tuple(b.features)` holds in every
case. The `PartModel` objects compare unequal, but the sole differing field is
`bbox`, a `BoundBox` with no value equality — the numbers print identically.

The clearest instance is `test_issue_1215`'s `_sweep`, which runs

```python
base = build_drawing(_PARTS[part_name](), title="T", number="N")
model = base.model()
```

once per `(part, feature, param, mode)` combination, using `base` for nothing but
`.model()`. Its sibling `test_the_sweep_actually_decorates_something` does the
same in a nested loop. That module is the single largest reclaimable block in the
suite (137.7 s), and this pattern is most of it.

Scale of the pattern: **523 `.model()` occurrences across 59 test files**, against
**11 files** that use `detect_part_model` today.

`detect_part_model` is not a test-only backdoor — it is the seam `sheet.py`
already imports and uses, so this moves tests onto the supported path rather than
off it.

## Scope

1. Audit the `.model()` call sites and identify those where the built drawing is
   used for nothing else. Grep finds candidates; only reading each one confirms it.
2. Replace those with `detect_part_model(part)`.
3. Where a module needs the same model repeatedly, hold it in a module-scoped
   fixture (overlaps the sibling duplicate-build issue — do that module once,
   covering both).

## Gates

- Before converting a site, assert the equivalence *for that part*:
  `tuple(build_drawing(p).model().features) == tuple(detect_part_model(p).features)`.
  Do not generalise from the 12 parts checked here — ADR 0017 makes recognition
  lifecycle-sensitive, and a declared build recognises nothing until asked, so a
  model-bearing `model=` round-trip is not the same question as a detected one.
- Each converted module records measured before/after wall time.
- Any site where the drawing *is* also used keeps its build; say so in review
  rather than converting it and weakening the assertion.

[...]

Issue #1307, opened 2026-08-23
Epic: build_drawing is three quarters of the suite's cost, and its largest single cost is avoidable


pzfreo:
# Epic: `build_drawing` is three quarters of the suite's cost, and its largest single cost is avoidable

## Problem

Nothing in this repository measures where build time goes, so every performance
judgement to date has been a guess. This epic starts from measurement.

Instrumenting the whole fast tier (a pytest plugin wrapping `builder.build_drawing`,
`sheet.build_drawing` and `builder._build_drawing_once`, one JSON per xdist worker):

| | |
|---|---|
| fast tier, `-n auto --dist loadscope` | 4681 passed, **216 s** wall |
| aggregate test CPU (sum of call+setup+teardown ≥ 5 ms) | ~2320 s |
| time inside `build_drawing` | **1735 s over 2548 calls — ~75 %** |

So the suite *is* the engine. Any engine speedup is a CI speedup of roughly the
same proportion, and the converse holds: tuning the test harness alone cannot
reach the three quarters that is real drawing construction.

Within a build, one stage dominates. `leaders._measure` — the per-candidate
measurement in the shared feature-leader solve — is **552 s, 32 % of all build
time**, across 47 532 calls. Of those, **37 223 (78 %) have no
`analytical_geometry`** and account for **551 s of the 552 s**. The analytical
path costs ~0.1 ms per candidate; the OCC path ~14.8 ms. Roughly 150×.

That is not a micro-optimisation. It is one seam, already built and already
proven for ordinary holes, that the machined-feature callouts never got.

## Outcome

A build spends its time on geometry the drawing keeps, not on geometry it
constructs to measure and throws away; the test suite stops paying for identical
drawings many times over; and the numbers above are reproducible on demand rather
than rediscovered each time someone wonders why CI is slow.

## Work items

Ordered by what they cost to do, not by size of prize — the two test-side items
are pure harness work with no engine risk, and land first.

| | issue | measured prize | risk |
|---|---|---|---|
| 1 | #1311 — `pr-check` uses `--dist loadscope` | 216 s → 152 s local (**−30 %**); **nothing at CI's 4 cores** | one line |
| 2 | #1310 — tests build a drawing to read `.model()` | 0.83 s → 0.01 s per site (~80×) | low |
| 3 | #1309 — tests rebuild identical drawings in one module | **506 s** reclaimable (lower bound) | low |
[...]

PR #1317, opened 2026-08-24
Add reproducible build profiling harness

pzfreo:
## Outcome

The #1307 performance claims now have a reproducible, fail-closed measurement path instead of an ad hoc local plugin. `scripts/profile-builds` runs an explicitly enumerated pytest selection, writes one profile per xdist worker, verifies every worker's collection census, and emits an aggregate summary.

Part of #1307. Stacked on #1314; independent of the #1310 conversion stack.

## Coverage of the real call paths

The plugin installs before test-module collection and wraps:

- `draftwright.builder.build_drawing` and the public lazy `draftwright.build_drawing` binding;
- `draftwright.sheet.build_drawing`, which `sheet.py` binds at import time and the original census missed;
- `draftwright.builder._build_drawing_once`.

Calls are aggregated by test node id and seam. Pytest setup/call/teardown phases taking at least 5 ms are retained for attribution. Public/Sheet times and `_build_drawing_once` are intentionally separate nested measurements rather than summed as unrelated work.

## Honesty properties

- Pytest modules/options remain distinct argv entries; no shell string is constructed, so zsh word splitting cannot silently collect zero tests.
- `--expect-collected` is required and positive.
- Every worker must report exactly that selected count; zero, partial, missing, or inconsistent profiles fail the runner even when pytest's exit status is zero.
- Each xdist worker writes a distinct JSON file. The controller does not emit an empty pseudo-worker.
- An output directory containing worker profiles is rejected rather than mixed with a new run.

## Verification

- Unit guards: 6 passed (literal argv, worker aggregation, zero/partial/missing census rejection, JSON evidence).
- Focused harness + workflow tests: 16 passed.
- Serial runner integration: one selected test produced `profile-main.json` and `summary.json` with collected count 1.
- Sheet integration: recorded one `sheet.build_drawing` and one `_build_drawing_once` call for the selected test.
- Public-builder integration: recorded five `builder.build_drawing` and five `_build_drawing_once` calls for the selected test.
- xdist integration (`-n 2 --dist loadscope`): two workers each reported the literal two-test collection; controller emitted no profile.
- `scripts/pr-check --static`, focused Ruff check/format, `git diff --check`, and executable mode check: green.
- Slow tests not run.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -54,6 +54,7 @@ Compact map, bottom to top:
 - **Facades / top layer** — `make_drawing.py` + `annotate.py` (thin compat),
   `sheet.py` (the fluent `Sheet` facade, ADR 0011), `sheet_emit.py` (the `--script`
   emitter), `cli.py` (Typer; engine imported lazily inside command bodies, #313),
+  `_build_profile.py` (developer-only pytest/runner profiling support),
   `evaluation/` (the versioned STEP-analysis benchmark — production code must never
   depend on benchmark expectations or scores), `recogniser_contract.py` (the
   fail-closed cross-repository capability join).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -151,6 +151,20 @@ checks. Target is 100% passing. Tiers (#153):
   not mint a new dense fixture.
 - **`-m slow`** (CTC fixture builds) — CI-only.
 
+For reproducible build-cost profiling, use a fresh output directory and state the expected
+collection census explicitly:
+
+```bash
+scripts/profile-builds --output /tmp/draftwright-profile \
+  --expect-collected 4740 -- tests/ -n auto --dist loadscope
+```
+
+The runner passes every module/option as a literal argv entry, writes one JSON file per xdist
+worker, and refuses to report success when any worker's collected count differs. It times the
+public builder binding, `Sheet`'s import-time builder binding, and `_build_drawing_once`, and
+records pytest phases of at least 5 ms for attribution. Do not reuse an output directory that
+already contains worker profiles.
+
 Coverage is kept out of the default addopts (it adds ~13% locally); the CI
 workflow passes the `--cov` flags. CI runs the full fast tier (3×3 OS/Python
 matrix, parallelised with `-n auto`) on every PR; the **slow tier runs post-merge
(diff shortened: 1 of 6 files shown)


pzfreo:
CI finding fixed at `e00ba3d`: both platform canaries correctly rejected the new source module as absent from the machine-enforced architecture DAG. `_build_profile` is now rank 7 (a developer-only top-layer consumer), CLAUDE and `docs/architecture.md` are synchronized, and the lazy upward package-root import was removed by using the already-loaded package from `sys.modules`.

Post-fix verification:

- exact hosted-canary command: 203 passed;
- import-boundary + profiler guards: 17 passed;
- full static preflight: green;
- slow tests not run.

The push restarted required CI. Independent review of the new exact head is in progress; the earlier acceptance applies only to superseded `e91d5b5`.

pzfreo:
Independent exact-SHA review: **ACCEPTED** at `e00ba3d6cbdde9da5a1c263f8c97739555c97d72` against base `e1a8f02feb351dcbe8dc8c148144894fa8995291`, with no findings. The reviewer revalidated architecture rank/docs, serial and two-worker profiler output, all three call seams, collection censuses, 17 import-boundary/harness tests, architecture guards, static checks, and a clean diff. Required replacement CI is running; slow remains skipped.

pzfreo:
Independent exact-SHA re-review after rebasing onto merged current main: **ACCEPTED with no findings**.

- head: `0b20db0dbafa2fe4ef0262130c4ba7d2e96b01c1`
- base: `ff5f4304f860d61d6ce1a98ee76faf2fe309dad0`
- both commits are patch-identical by range-diff to the previously accepted profiler + architecture fix
- import-boundary and harness suite: 17 passed
- serial/xdist seam, worker-file, stale/census/zero-count, literal-argv and layer-DAG behavior were re-probed successfully
- `scripts/pr-check --static` and `git diff --check`: passed

Replacement hosted CI is still required before merge. Slow tests remain skipped.

pzfreo:
The first complete run exposed a real merge blocker: Codecov patch coverage was **59.37%** (target 93.37%), even though the regular matrix and `ci-ok` passed.

Fixed at exact head `4bb773e1e8129bbc1b4c6ae57143590dec1541f0` by exercising the plugin lifecycle rather than suppressing coverage:

- output-option registration;
- inactive/no-output and xdist-controller paths;
- both worker and non-xdist main paths;
- all three wrapped build bindings;
- collection census, slow-phase filtering, node transitions, and session write;
- missing-profile rejection.

Local targeted result: **11 passed**; static preflight green; changed-line coverage is now **100% (96/96 lines)** via the downloaded canonical coverage configuration. Fresh exact-SHA independent review and replacement CI are running. Slow tests remain skipped.

PR #1319, opened 2026-08-24
perf(tests): avoid repeated model-only cross-pass builds

pzfreo:
## Summary

Fourth focused #1310 slice, stacked on #1318.

- split the reusable narrow cross-pass solid factory from its detected-model helper;
- replace six full `build_drawing(..., auto_dims=False).model()` executions inside that helper with `detect_part_model`;
- add a permanent fresh-solid `tuple(features)` equivalence gate;
- leave every downstream rendered feature-leader drawing and assertion unchanged.

One equivalence build replaces six model-only builds: five net full builds removed.

## Measurement

Same 18-core macOS/Python 3.14.7 host.

Whole module, serial:

```console
/usr/bin/time -p uv run pytest tests/test_issue_1166_cross_pass_feature_leaders.py -q
```

- base: 143 passed, 42.38 / 42.71 s wall;
- head: 144 passed, 42.71 s warm wall (an initial 85.02 s wall / 49.18 s CPU host stall is disclosed);
- result: no whole-module wall change distinguishable from noise—the layout-heavy downstream builds dominate.

Isolated helper, ten calls in one process (geometry construction included on both sides):

- base: 1.049729 / 1.089552 s;
- head: 0.305086 / 0.327216 s;
- midpoint: 1.069641 -> 0.316151 s, **70.4% faster**.

## Validation

- changed module: 144 passed;
- permanent fresh-part feature-equivalence gate;
- `scripts/pr-check --static`;
- `git diff --check`.

Slow tests were not run.

Part of #1310 and #1307. The audit remains open.

PR #1320, opened 2026-08-24
perf(tests): avoid a model-only slanted profile build

pzfreo:
## Summary

Fifth focused #1310 slice, stacked on #1319.

- extract a reusable factory for the expensive slanted/blind solid;
- use `detect_part_model` in the recognition-only test instead of building a full detail drawing solely for `.model()`;
- prove exact `tuple(features)` equivalence against a separately constructed fresh solid inside the existing full rendered-drawing test;
- retain that full build and every annotation/view/lint assertion.

The other three model-only geometries in this module are deliberately unchanged: each is a singleton with no existing detected full build, so the mandatory equivalence gate would replace one full build with one full build and save nothing.

## Measurement

Same 18-core macOS/Python 3.14.7 host and serial command:

```console
/usr/bin/time -p uv run pytest tests/test_slanted_blind_step.py -q
```

- base: 12 passed, 16.97 / 17.12 s wall;
- head: 12 passed, 12.98 / 13.06 s warm wall;
- midpoint: 17.045 -> 13.020 s, **23.6% faster**.

The first head run stalled at 61.81 s wall with 18.04 s user CPU. It is disclosed and excluded from the midpoint.

## Validation

- changed module: 12 passed on both retained head runs;
- fresh-part exact feature-equivalence assertion;
- `scripts/pr-check --static`;
- `git diff --check`.

Slow tests were not run.

Part of #1310 and #1307. The remaining audit stays open.

### DIFF
--- a/tests/test_slanted_blind_step.py
+++ b/tests/test_slanted_blind_step.py
@@ -32,11 +32,11 @@ from build123d import (
 )
 
 from draftwright import build_drawing
+from draftwright.builder import detect_part_model
 from draftwright.model.declare import envelope
 
 
-@pytest.fixture(scope="module")
-def slanted_blind_step():
+def _slanted_blind_step():
     align_min = (Align.MIN, Align.MIN, Align.MIN)
     with BuildPart() as part:
         with BuildSketch(Plane.XZ):
@@ -59,6 +59,11 @@ def slanted_blind_step():
 
 
 @pytest.fixture(scope="module")
+def slanted_blind_step():
+    return _slanted_blind_step()
+
+
+@pytest.fixture(scope="module")
 def bounded_slanted_recess():
     """A partial-width ramp whose ends are defining profile transitions."""
     align_min = (Align.MIN, Align.MIN, Align.MIN)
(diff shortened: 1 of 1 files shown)


pzfreo:
Independent exact-SHA review: **ACCEPTED with no findings**.

- head: `7999f3b532c63bc83f9224a90f5b833b0ada9d62`
- base: `213732e80a352ad0b7b3d149babfc3c6116e6392`
- only the expensive recognition-only slanted build was converted;
- the retained rendered-detail build compares exact features against direct detection of a separately constructed fresh solid;
- annotation names, detail labels/views, and lint assertions are unchanged;
- all three zero-gain singleton sites were confirmed and retained;
- changed module: 12 passed in 12.60 s;
- static checks and `git diff --check` passed.

The 23.6% timing arithmetic and disclosed stalled run were independently checked. Hosted CI and stack dependencies remain required; slow tests stay skipped.

pzfreo:
Independent exact-SHA re-review: **ACCEPT** at `8ca39ab900ff8e8bd5775af6af9df84de7b0fbff` against current main `3c5219d591f6e6cf7419f94a4d915ea9acfc9a04`. No findings.

The reviewer verified that the replaced drawing was consumed only through `model().features`; the retained full rendered test proves exact `tuple(features)` equivalence against detection on a separately constructed solid; detail-view, annotation, label, lint, and placement assertions remain unchanged; and no mutable fixture sharing or private shortcut was introduced.

Validation: focused non-slow module 12 passed; private-access canaries 4 passed; Ruff, formatting, diff, and clean-worktree checks passed. Slow tests were not run.

PR #1322, opened 2026-08-24
Lint the line-work drawn through a label, not the ink two annotations share

### LINKED ISSUE #1321: annotation_overlap compares label boxes, so ink drawn over a label goes unreported
`annotation_overlap` compares **label boxes**, so it is silent whenever something
other than a label is drawn over a label — a dimension line, an arrowhead, a
leader. On GRM-03 it fires **zero** times while three pairs of annotations share
ink in a region a reader can see.

Split out of #1274 because that thread is hard to read: it contains two
corrections of my own mistaken evidence before it reaches anything usable. This
is the settled version, with a working implementation behind it.

## The measurement

GRM-03 thumbwheel drive screw, A3 at 4:1. Lint reports six
`label_centerline_overlap` and **no** `annotation_overlap`:

```
FINDINGS 3   line_crossings 7   exempt 3   examined 13
   0.9444 mm²  m_env_width x m_steplen0     place 8.98 x 5.00 mm
   0.1258 mm²  m_env_width x m_steplen4     place 0.75 x 1.49 mm   (two arrowheads)
   0.0669 mm²  m_steplen1  x m_steplen2     place 2.10 x 0.60 mm
```

## Why the obvious fix does not work

Two things have to be got right, and I got both wrong first.

**Ink area alone reports ordinary drafting as a defect.** Strokes are filled
faces about 0.1 mm wide, so everything crosses everything. Measured on the same
sheet, the *largest* shared area was two collinear extension lines:

```
m_steplen0 x m_steplen1    1.1004 mm²    region 0.100 x 11.000 mm
m_steplen3 x m_steplen4    1.1004 mm²    region 0.100 x 11.000 mm
```

That is the stacked-dimension case `structural.py`'s own comment calls out as
the reason full bounding boxes were rejected — reproduced exactly, from the
other direction. Seven such pairs on this sheet.

**The union of everywhere a pair meets is not the region to test.** Two
stacked dimensions sharing left *and* right extension lines give two thin places
forty millimetres apart; unioned they are 40 × 11 mm with 2.2 mm² of ink and
look like the worst collision on the drawing.

A value-overlap test separates nothing either: chamfers share 1 of 10 values
with what is drawn, risers 1 of 4, holes 3 of 7 — and holes *are* dimensioned
while chamfers are not.

## What does work

Group the shared ink into **places** by proximity, and treat a place as a
collision only if it has width **and** height:

- collinear line-work is thin in one axis however long it runs
- a single crossing is thin in both
- ink shared across a region with both is something a reader sees

`0.5 mm` in both axes is your own figure, from `annotation_overlap`'s
`ox > 0.5 and oy > 0.5`, applied to ink instead of boxes.

### Two cases any such rule must not break

Both from this sheet:

- **must stay excluded** — two stacked dimensions sharing left and right
  extension lines: two places of 0.1 × 11 mm, forty apart. Baseline dimensioning.
- **must stay reported** — a place made of faces none of which is itself square:
  `m_steplen1`/`m_steplen2` meet in 0.10 × 0.10 and 0.10 × 0.60 two millimetres
  apart, one place of 2.10 × 0.60.

Per-face tests break the second. Union tests break the first.

Also worth exempting anything flagged `is_centerline`, as `annotation_overlap`
already does — three on this sheet, and a locating dimension reaching over a
centre mark is not a defect.

## Two traps in the build123d API

Both cost me a debugging session:

- `Shape.intersect` returns a **`ShapeList`**, not a shape, so
  `getattr(result, "area", 0.0)` — the natural spelling — silently answers `0.0`
  for every pair and reports every sheet clean.
- It returns **`None`**, not an empty list, for disjoint shapes.

## Known blind spot, which is yours already

A line drawn *through* text shares ink only as tall as the line, so no test on
the shape of the shared region will ever call it a collision.
`leader_line_through_text` covers that directly — the two checks are
complementary rather than overlapping.

## Implementation

Running in `pzfreo/draftwright-io` since #531, on pinned public API only
(`iter_annotations`, `view_of`, `page_w`/`page_h`, and build123d's
`bounding_box`/`intersect`/`area`). A PR against this repo follows.

Supersedes the ink-related half of #1274; the `label_centerline_overlap`
empty-label half is unaffected and still reproduces on 0.4.11.

pzfreo:
Closes #1321. Stage 1 of #1331; the design change is #1332.

`annotation_overlap` compares labels with labels, so it cannot see a dimension line, extension line, or leader shaft drawn over text. This adds `annotation_ink_overlap` to measure that defect directly.

## What it measures

Arithmetic clipping of each annotation's located `segments` against the other label's region, in both directions:

- axis-aligned labels use `label_bbox`;
- rotated dimensions use an exact transform-aware `label_polygon` (added at the source by build123d-drafting-helpers PR #182);
- duck-typed diagonal dimension objects without exact polygon metadata are classified from public geometry and excluded explicitly, rather than measured against a known-inflated AABB.

The longest connected stroke path wins. Adjacent renderer segments join only when their shared endpoint lies inside the label; disjoint corner grazes do not sum. A 0.5 mm floor excludes the one confirmed 0.308 mm corner graze while retaining the smallest confirmed defect (0.854 mm).

No CAD boolean is used. The previous shared-ink implementation was removed because `Shape.intersect` decomposed identical drawings differently across supported build123d releases and cost +79% on the fixture corpus.

## Latest review fixes

- Removed the dependence on optional private `_dw_spec` for diagonal classification.
- Added exact convex-polygon clipping for rotated labels, eliminating both the diagonal false-negative and the direct-helper AABB false-positive.
- The dense-plate regression now names all six real `10/20/30/40/50/60 → 2× 14.1` crossings (1.5 mm each).
- Connected pieces of one rendered stroke now combine across an in-label endpoint, while independent grazes remain separate.
- build123d-drafting-helpers 0.14.2 is now the minimum and authoritative transform-aware polygon source; Draftwright’s temporary compatibility backfill has been removed.

## Behaviour and scoring

`annotation_ink_overlap` is a legibility code and is aggregated against the crossed label, so several strokes obscuring the same text remain one primary defect. If two labels already overlap and their line-work also crosses, the surviving `annotation_overlap` message carries the crosser, crossed label, and length instead of suggesting a text-only move.

This deliberately does not diagnose line-on-line defects with no text involved. `leader_line_through_text` remains separate because it checks an annotation's own leader against its own text; this check only compares distinct items.

## Validation

- focused latest-head suite against published helpers 0.14.2: **129 passed, 0 skipped**;
- dense diagonal regression: six named crossings detected;
- Ruff lint and formatting: passed;
- slow tests deliberately not run; the normal fast matrix is owned by CI.

The branch retains the earlier kernel-agreement corpus, named fixture-pair assertions, and defensive duck-typed metadata tests.


pzfreo:
## Measured against the whole STEP corpus

All 23 STEP fixtures in `tests/fixtures` (11 `*.step` + 12 `*.stp`, both protocols of NIST CTC-01…05), built through `build_drawing` — so `analysis._import_step`, not `build123d.import_step`. Nothing reimplemented the gating: the real `worst_shared_place`, `places_where_ink_meets` and `lint_drawing` were wrapped, and the harness verdicts match `lint_summary()['by_code']` on **23/23**. (First attempt tripped its own self-check — `repair=True` lints 3–7 times per build and pairs were accumulating across passes.)

### At 0.3 the check reported nothing, and cost 79%

`by_code` came back **identical to the pre-#1321 baseline on every fixture**. Zero findings. Total build time **264.6s → 473.9s (+79%)**. Confirmed uninstrumented on the worst three, serialized:

| fixture | baseline | branch |
|---|---:|---:|
| nist_ctc_05_asme1_ap203 | 24.8s | 46.5s |
| nist_ctc_01_asme1_ap203 | 6.7s | 21.3s |
| issue_915_case_study_2 | 16.7s | 22.9s |

### The five places it was suppressing are all real

7 of 1908 places pass the shape+label test; all 7 sit under 0.3. Rendered at 600 dpi and looked at:

| area | fixture | what the render shows |
|---:|---|---|
| 0.2755 | grm03 `0.5`×`2` | arrowhead driven through the `5` |
| 0.2124 | ctc-02 ap203 | dimension line struck through a hole-table row, arrowhead on the `1` of `10.1` |
| 0.1828 | ctc-02 ap242 | line through the row, arrowhead covering the `⌀` |
| 0.1312 | issue_915 | arrowhead on the `×` of `50 × 120 × 5 DEEP` |
| 0.0944 | ctc-02 ap203 | arrowhead on the `⌀`, extension line down through `22` |

All ink over label text. The floor is back to the sibling's 0.05.

### One of the three anchors does not reproduce

Two do, exactly: #915 at 0.1312 (claimed 0.13) and the flange at 1.4499 (claimed 1.45).

The third — *"dimension line through a label (GRM-03) — 0.94 mm², region 9.0 × 5.0"* — has no counterpart on the drawing this engine builds for GRM-03. Every place there:

```
 1.1008 mm²  0.100 x 11.000  on_label=False   '3'   x '18'    <- the excluded collinear case
 1.1008 mm²  0.100 x 11.000  on_label=False   '2'   x '3'
 0.5917 mm²  0.100 x  5.917  on_label=False   '0.5' x '2'
 0.2755 mm²  1.311 x  1.958  on_label=True    '0.5' x '2'     <- largest on-label place
 0.0220 mm²  0.263 x  0.100  on_label=True    '0.5' x '2'
```

Largest region by extent is 1.31 × 1.96 mm. The GRM-03 numbers the module *excludes* do reproduce (1.1008 against the documented 1.1004), which is what makes this one specific rather than a whole-fixture mismatch. That measurement belongs to the sibling's AI-authored sheet of the same part.

And the remaining anchor was the argument for the raise — rendered, the #915 "graze" is the same defect as the blatant cases, smaller, not a different kind.

### The budget is reached, and stopped silently

Not a pathological-sheet guard in practice:

| fixture | annotations | lint passes | booleans (last pass) |
|---|---:|---:|---:|
| nist_ctc_02_asme1_ap203 | 162 | 3 | **250 — saturated** |
| nist_ctc_02_asme1_ap242 | 165 | 3 | **250 — saturated** |
| nist_ctc_04_asme1_ap203 | 144 | 3 | **250 — saturated** |
| nist_ctc_04_asme1_ap242 | 126 | 3 | **250 — saturated** |
| nist_ctc_05_asme1_ap242 | 77 | 6 | 162 |
| nist_ctc_01_asme1_ap203 | 54 | 6 | 138 |

GRM-03 spends 6, not ten. Four sheets abandon the check part-way and say nothing — the sheets it matters most on. It now emits `annotation_ink_overlap_truncated`, in `_UNSCORED_CODES` because it is the linter's ceiling, not the drawing's defect. Also: it is a **per-`lint_drawing` ceiling** and a build lints 3–7 times, so a saturating sheet spends up to 3× it per build.

### The flange failure is kernel-dependent, not just fixture-dependent

Worth separating from the threshold question. `pyproject.toml:18-19` pins `build123d<0.11` below Python 3.13 and `>=0.11` above, so the PR matrix spans two kernels — **OCCT 7.8.1 / build123d 0.10.0** on 3.10–3.12 and **OCCT 7.9.3 / build123d 0.11.1** on 3.14. Same commit, same shard (`--splits 2 --group 2`, confirmed the test collects into group 2), 2546 tests each: green on 3.10/3.11/3.12, red only on 3.14.

I ran the full suite under both. On **3.12 / build123d 0.10.0** the flange fixture has no `annotation_ink_overlap` at all — not at 0.3, and not at 0.05 either. On **3.14 / 0.11.1** it has 1.4499 mm² across 5.13 × 1.15 mm. That is not a threshold artifact; the two kernels draw that step-length chain differently. Which one is right is your call, and the assertion is left untouched.

### Suite

- py3.14 / build123d 0.11.1 — **4792 passed, 3 failed**: the two #915 lint-clean assertions this branch now rewrites, plus the pre-existing flange case.
- py3.12 / build123d 0.10.0 — **4798 passed**, no related failures. The lower floor introduces nothing on the other kernel.

Both new rules are mutation-checked: reverting the floor to 0.3 fails three named tests, disabling the truncation report fails two.

pzfreo:
## Correction: the flange disagreement is not a layout difference

My earlier comment said the two kernels "draw that step-length chain differently" and that this "is not a threshold artifact". Both halves are wrong, and the second one mattered — it pointed at the wrong fix.

**The two kernels lay this fixture out identically.** Same label extents to 2dp — `'8'` at `(173.16, 117.92, 174.63, 120.08)`, `'4× 2'` at `(182.44, 117.93, 189.35, 120.07)` — and the same full extents. Rendered at 600 dpi, the merged arrowhead blob is plainly present on **both**.

What differs is only what `Shape.intersect` returns for the same pair. Raw pieces of shared ink, before any grouping:

**build123d 0.10.0 — 3 pieces**

```
 x  177.84 ->  177.94   (0.10 wide, 0.10 tall)   0.0100 mm²
 x  179.39 ->  180.40   (1.01 wide, 1.15 tall)   0.8508 mm²   <- the blob
 x  182.68 ->  182.97   (0.30 wide, 0.10 tall)   0.0235 mm²
```

**build123d 0.11.1 — the same 3, plus 3 hairlines, each exactly one stroke-width tall**

```
 x  177.89 ->  180.28   (2.38 wide, 0.10 tall)   0.2231 mm²
 x  179.24 ->  180.44   (1.20 wide, 0.10 tall)   0.1196 mm²
 x  179.51 ->  181.89   (2.38 wide, 0.10 tall)   0.2231 mm²   <- this one decides it
```

Those hairlines are where the two dimension lines lie along each other. Both versions *draw* those lines identically; 0.10 simply does not return them as intersecting.

The verdict then turns entirely on `CLUSTER_GAP_MM = 2.0`:

- **0.10** — blob ends at 180.40, next piece starts at 182.68. Gap **2.28 mm**, wider than 2.0, so they stay two places. The blob alone stops short of the label box (starts 182.44). Silent.
- **0.11** — the third hairline reaches 181.89, leaving only 0.79 mm to the next piece. Everything links into one place spanning 177.84 → 182.97, which reaches the label. Reported.

So the whole disagreement is one hairline, turning a 2.28 mm gap into a bridged one — a **14% margin on the joining distance**, not a difference of drawing.

This also explains why dropping the ink floor to 0.05 did not change it: the blob was always far above any floor under discussion. What decided the outcome was whether the pieces counted as one place or two.

Swept on the fixture itself, re-running the real grouping and the real verdict at each value:

| joining distance | 0.10.0 | 0.11.1 |
|---|---|---|
| 2.00 mm | 2 places, silent | 1 place, reported 1.4499 mm² |
| 2.25 mm | 2 places, silent | reported |
| **2.30 mm** | **1 place, reported 0.8842 mm²** | reported |
| 3.00 mm | reported | reported |
| 10.00 mm | reported | reported |

At 2.30 and above both kernels agree on the *same place* — 5.13 × 1.15 mm — differing only in how much ink they attribute to it.

The defect is real in both renders, so 0.10 is the one getting it wrong. I am widening the joining distance rather than leaving the assertion to a which-kernel-is-correct judgement, and measuring the corpus on both kernels for false positives before settling the value. Correction to the PR description to follow.

pzfreo:
Fixed the two latest-head review findings in `5bf794c`:

1. Diagonal-label crossing no longer depends on optional private `_dw_spec`. Lint consumes exact `label_polygon` geometry, classifies older direct dimensions from public segment metadata, and declines inflated-AABB guesses. The dense-plate regression now names all six real crossings through `2× 14.1`.
2. Crossing length now combines connected rendered segments when their shared endpoint is inside the label, while keeping disjoint corner grazes separate.

The authoritative transform-aware polygon API is in pzfreo/build123d-drafting-helpers#182; draftwright backfills it while 0.14.1 remains supported. Validation: 127 focused tests passed against 0.14.1 (one new-helper-only skip), all 120 pass with #182 overlaid, and `scripts/pr-check --static` is green. Slow tests were not run.

pzfreo:
helpers 0.14.2 is published and now consumed by this PR. Head 7f83eb6 raises the minimum to >=0.14.2, refreshes the lock, removes Draftwright’s temporary polygon backfill, and makes the exact diagonal-polygon regression mandatory. Focused suite: 129 passed, 0 skipped. Static/type/format/docs checks are clean; slow tests were not run.

PR #1323, opened 2026-08-24
Finish model-only drawing audit

### LINKED ISSUE #1310: Tests build a whole drawing to read .model(), which detect_part_model returns 80x faster

pzfreo:
Closes #1310
Part of #1307.

## What changed

The earlier five #1310 slices left a real audit tail. This finishes the AST-assisted and manual audit of all current test-suite `.model()` call sites:

- converts 19 remaining model-only full-build executions across seven modules;
- adds/retains fresh-solid exact `tuple(features)` gates for every distinct converted geometry;
- removes 15 net full builds after accounting for four new full-build equivalence cases;
- preserves public `build_drawing` recognition/lifecycle tests, foreign-Drawing ownership tests, and one-off sites whose mandatory gate would cancel the only saving.

No mutable drawing state is shared. Direct detection is used only where the discarded Drawing was consumed solely through its model/features.

## Measured evidence

Honest profiling-harness runs over all seven affected modules, serial on the same host:

| | base | head |
|---|---:|---:|
| selected tests | 310 | 314 |
| `builder.build_drawing` calls | 223 | 208 |
| `builder._build_drawing_once` calls | 241 | 226 |
| public-build time | 88.303 s | 82.045 s |
| pytest wall | 119.79 s | 113.19 s |

That is 15 fewer complete builds, 7.1% less measured public-build time, and 5.5% less wall while adding four permanent equivalence cases.

Repeated warm per-module wall was mostly flat/noise at individual-module scale; the clear individual movement was `test_issue_885_prismatic_coverage.py`, 15.66 -> 13.20 s. Cold/stalled first runs are disclosed in the issue and not used as gains.

## Gates

- affected non-slow suite: 314 passed, 2 deselected;
- private import/attribute canaries: 4 passed;
- static, typing, docs, Ruff, formatting and diff hygiene: green;
- independent pre-rebase review: ACCEPT, no findings;
- fresh exact-head review and regular CI are pending;
- slow tests were not run and will not be awaited.

### DIFF
--- a/tests/test_adr0018_view_selection.py
+++ b/tests/test_adr0018_view_selection.py
@@ -24,15 +24,31 @@ import pytest
 from build123d import Box, Cylinder, Pos, Rot
 
 from draftwright import ViewPlanIncomplete, build_drawing
+from draftwright.builder import detect_part_model
 from draftwright.compose import _layout_geometry
 from draftwright.view_plan import VIEW_AXES, VIEWS_SHOWING, views_showing
 
 ALL_THREE = ("front", "plan", "side")
 
 
+def _plain_box():
+    return Box(90, 60, 25)
+
+
 def _z_hole():
     """A part whose only feature reads face-on in PLAN — so plan is necessary."""
     return Box(90, 60, 25) - Pos(20, 15, 0) * Cylinder(4, 30)
+
+
+@pytest.mark.parametrize(
+    "part_fn",
+    (_plain_box, _z_hole),
+    ids=("plain-box", "z-hole"),
+)
+def test_direct_detection_matches_built_model_features_for_view_planning_parts(part_fn):
+    built = tuple(build_drawing(part_fn()).model().features)
+    detected = tuple(detect_part_model(part_fn()).features)
+    assert detected == built
 
 
 def _x_hole():
@@ -213,7 +229,7 @@ class TestTheAsymmetricCounterexample:
         from draftwright.model.ir import RequestedDimension
 
         part = _z_hole()
-        detected = build_drawing(part).model()
+        detected = detect_part_model(part)
         assert detected is not None
         envelope = next(feature for feature in detected.features if feature.kind == "envelope")
         hole = next(feature for feature in detected.features if feature.kind == "hole")
@@ -287,7 +303,7 @@ class TestAnExtentMovesOrIsReported:
     def test_the_planner_record_itself_re_homes_instead_of_only_the_renderer(self):
         from draftwright.model.planner import plan_dimensions
 
-        model = build_drawing(Box(90, 60, 25)).model()
+        model = detect_part_model(_plain_box())
         assert model is not None
         envelope = next(
             group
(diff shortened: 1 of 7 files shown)


pzfreo:
Independent exact-SHA review: **ACCEPT** at `ffa55f69cb8a09173b8c780df718685706db3911` against exact current main `0d0ab60e12a37a396b0aea45f3ebbd6e1926b9e7`. No findings.

The reviewer confirmed via `git range-diff` that the rebased commit is identical to the previously accepted candidate. The diff remains exactly seven test modules; every distinct converted geometry retains a fresh-solid exact `tuple(features)` gate; public lifecycle/drawing assertions remain; and there are no conflict artifacts, mutable sharing, or private-surface expansion.

Validation: affected non-slow tests 314 passed / 2 deselected; private import/attribute canaries 4 passed; Ruff, formatting, diff and clean-worktree checks passed. Slow tests were not run.

PR #1324, opened 2026-08-24
Share read-only sheet-emitter drawings

### LINKED ISSUE #1309: Tests rebuild identical drawings within one module, costing over 500 s of suite time
Part of #1307.

## Problem

Test modules rebuild the same drawing from the same part with the same kwargs,
many times over, inside a single module.

Census over the fast tier (a plugin keying each `build_drawing` call by a geometry
signature — volume, bbox, face and edge counts — plus its normalised kwargs):

- 2539 recorded builds, **1362 distinct keys**
- **625 s** in repeat builds of an already-built key (38 % of recorded build time)
- **506 s** of that is repeated *within one module*, i.e. reachable by a
  module-scoped fixture with no cross-file coupling

Worst offenders:

| module | reclaimable |
|---|---|
| `test_issue_1215_no_approved_tolerance_is_dropped.py` | 137.7 s |
| `test_make_drawing.py` | 80.1 s |
| `test_issue_1166_cross_pass_feature_leaders.py` | 24.4 s |
| `test_issue_1143_hole_completeness.py` | 23.3 s |
| `test_sheet_emit.py` | 22.0 s |
| `test_issue_1144_transactional_hole_table.py` | 21.7 s |
| `test_slanted_blind_step.py` | 17.2 s |
| `test_issue_1202_observed_drawing_consumer.py` | 15.1 s |
| `test_layout_property.py` | 14.1 s |
| `test_pmi.py` | 13.7 s |

Individual keys are stark: one 20×15×4 plate at `auto_dims=False, page='A4'` is
built **97 times**; a 140×60×12 plate **26 times**; `nist_ctc_01_asme1_ap242.stp`
with `pmi='annotate'` **3 times at 15.4 s each**.

The suite has 155 test files and **37 fixtures total**, only 29 of them
module-scoped, against 1197 literal `build_drawing(` / `make_drawing(` call sites.
Sharing is the exception, not the rule. CLAUDE.md already states the intended
convention — *"a critique-style test should share a module-scoped built drawing,
not mint a new dense fixture"* — it is simply not applied in these modules.

## Scope

Convert the read-only repeats in the listed modules to module-scoped fixtures,
highest-value first. Take them one module per PR so each carries its own before/after
measurement.

The constraint that decides each case: **only share where the tests do not mutate
the drawing.** A critique-style test that reads `registry`, `lint_summary()`,
`iter_annotations()` or `model()` is safe. Anything that adds a dimension, runs
repair, or exports must keep its own build, or take a copy — and a deep copy of an
OCC-backed drawing is not obviously cheaper than rebuilding it, so measure rather
than assume.

Note also that `--dist loadscope` (what CI uses) keeps a module on one worker, so
module-scoped fixtures are actually shared there. Under `--dist worksteal` (see
the sibling issue) they are rebuilt per worker, which caps the gain. Worth
measuring both once a few modules are converted, since it may change which dist
mode is right.

## Gates

- Each converted module records its own measured before/after wall time in the PR.
- Every converted test still asserts its own precondition — a shared fixture must
  not become a fixture whose defect nobody re-checks.
- No converted test mutates the shared drawing. Demonstrate this, do not assert it
  by reading: mutating the shared object in one test and watching a named sibling
  fail is the proof.

## Measurement caveat

The census patched `builder.build_drawing` and `draftwright.build_drawing` but not
`sheet.build_drawing`, which `sheet.py` binds at import. Builds routed through the
`Sheet` facade (305 `Sheet(` call sites in tests) are therefore **missing** from
these totals — the real reclaimable figure is higher.

pzfreo:
Closes #1309

This focused slice shares the 22 automatic drawings consumed by the mirror-parity and furniture-accounting readers in tests/test_sheet_emit.py. Both readers and the named pristine-state sibling deliberately live in one TestTheDimensionMirror class so CI loadscope schedules one fixture instance.

Measured exact CI-shaped focused path (-n 4 --dist loadscope, 24 cases):
- base: 44 public automatic builds across the two readers
- candidate: 22 builder.build_drawing calls, 22 Sheet builds, 44 internal build passes
- result: exactly 22 duplicate public full builds removed

Safety gates:
- every shared case retains its expected feature-kind precondition
- both consumers remain read-only
- a deliberate annotation removal in the furniture reader makes the following named sibling fail on channel under the same loadscope command; the mutation was removed and the 24-case run returned green
- full non-slow module: 394 passed, 1 deselected, 2 xfailed
- repository static preflight, formatting, private canaries, and diff hygiene pass
- independent exact-SHA review of 669fc24339e13f31c5421a2554a58923f8671457 on a1d2cf3e838386f290c27cd3283880f7c61b096a: ACCEPT, no findings

Slow tests were not run.

### DIFF
--- a/tests/test_sheet_emit.py
+++ b/tests/test_sheet_emit.py
@@ -2568,7 +2568,9 @@ class TestTheDimensionMirror:
         assert "location_off_axis" in roles, f"got {sorted(roles)}"
 
     @pytest.mark.parametrize("name", sorted(_corpus()))
-    def test_the_mirrored_script_draws_what_the_automatic_drawing_draws(self, name):
+    def test_the_mirrored_script_draws_what_the_automatic_drawing_draws(
+        self, name, mirror_automatic_drawings
+    ):
         """The acceptance, per fixture: annotation-set EQUALITY.
 
         On drawn annotations, not emitted text — a text assertion passes on a script that
@@ -2576,11 +2578,7 @@ class TestTheDimensionMirror:
         #932. A corpus rather than one part, because single-fixture guards are how whole
         paths stayed uncovered twice in this series (#925 had no holes, #934 had one ladder).
         """
-        part = self._corpus()[name]
-        model = detect_part_model(part)
-        # Same title/number on both sides — the signature includes the title block, and a
-        # mismatch there would read as a dimension difference.
-        automatic = build_drawing(part, model=model, title="T", number="N")
+        part, model, automatic, _initial = mirror_automatic_drawings[name]
         src = emit_sheet_script(model, "part", "s", title="T", number="N")
         regenerated = self._run(src, part)["sheet"].build()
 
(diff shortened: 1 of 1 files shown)


pzfreo:
Independent exact-SHA review: ACCEPT, no remaining findings.

Reviewed head 669fc24339e13f31c5421a2554a58923f8671457 against current-main base a1d2cf3e838386f290c27cd3283880f7c61b096a. The initial review found a P1 loadscope grouping bug (three scheduling scopes caused 66 automatic builds and disconnected the mutation guard). The replacement co-locates both readers and the sibling in one class. Independent revalidation reproduced 24 focused passes with 22 builder.build_drawing, 22 sheet.build_drawing, and 44 internal passes; full non-slow module 394 passed / 1 deselected / 2 xfailed; private canaries and static/diff gates green. No slow tests run.

PR #1326, opened 2026-08-24
Measure machined feature leaders analytically

### LINKED ISSUE #1308: The machined-feature leader pass builds throwaway OCC geometry for every candidate it measures
Part of #1307.

## Problem

Exactly two sites construct a `FeatureLeaderJob`:

- `annotations/holes.py:2805` — passes `analytical_geometry=_analytical_geometry`,
  built on `_common.leader_callout_geometry` and a callout box measured once.
- `annotations/from_model.py:1943` (`_leader_callout_pass`) — passes nothing.

The second serves every machined-feature callout: chamfer, fillet, flat, pocket,
groove, boss, polygonal boss, polygonal stock, and the Y-axis step diameter. For
each of those, `leaders._measure` calls `job.build(tip, elbow, feature)`, which
constructs a real OCC `Leader` — an `Arrow`, a swept shelf rectangle and a text
sketch, fused — purely to read back a label box and some segments, then discards
it and tries the next candidate position.

Measured over the fast tier: **37 223 of 47 532 `_measure` calls take this path,
costing 551 s — 32 % of all build time.** Per candidate that is ~14.8 ms against
~0.1 ms for the analytical path.

On CTC-01 AP242 the same effect is visible in one build: `Compound.make_text` is
called **674 times for 54 distinct strings** — the label `7× R50` is rasterised
**192 times** — and `BRepAlgoAPI.Build` runs 1666 times, almost all of it
`Shape.fuse` inside these throwaway leaders.

This is not a new mechanism. ADR 0014 already describes it as the settled pattern:

> Ordinary-hole alternatives lower label/shaft geometry analytically and
> materialise only the selected OCC `Leader`; the rendered survivor is checked
> against that geometry before it is committed.

`_leader_callout_pass` simply never adopted it.

## Evidence that it works

A 12-line prototype (patch attached below) reuses the two helpers that already
exist — `_common.leader_callout_geometry` and the `lru_cache`d
`_core._text_size` — to supply `analytical_geometry` for the generic pass.

Per-fixture, best of three after a warm build:

| fixture | before | after | |
|---|---|---|---|
| `nist_ctc_01_asme1_ap242.stp` | 7.31 s | **3.83 s** | −48 % |
| `grm03_thumbwheel_drive_screw.step` | 2.43 s | **1.13 s** | −53 % |
| `issue_1058_wheel_rh.step` | 4.35 s | 4.33 s | unchanged |

The third is the control and worth keeping: that part's leaders are all *hole*
callouts, which already had the analytical path, so it correctly shows no gain.
The win is specific to machined-feature callouts.

Whole fast tier, `-n auto --dist worksteal`: **152.4 s → 106.8 s (−30 %)**.

Output is unchanged where it matters. On all three fixtures the produced drawing
is identical — same annotation set, same per-annotation bounding boxes to 3 dp,
same `lint_summary()["by_code"]`.

## Why it is not a drop-in

Applying the prototype fails **exactly 8 tests**, all candidate-budget guards:

```
test_issue_740_leader_assignment.py::test_candidate_budget_is_global_across_jobs
test_issue_740_leader_assignment.py::test_candidate_budget_fallback_restores_consumed_prefix_and_lazy_tail
test_issue_740_leader_assignment.py::test_grouped_job_candidate_budget_precedes_occ_construction
test_issue_740_leader_assignment.py::test_grouped_job_at_candidate_budget_still_uses_joint_assignment
test_issue_740_leader_assignment.py::test_pre_drain_y_diameter_keeps_the_policy_b_fallback_tail
test_issue_1166_cross_pass_feature_leaders.py::test_cross_pass_candidate_budget_precedes_collect_all_geometry
test_issue_1166_cross_pass_feature_leaders.py::test_fixed_obstacle_probe_budget_precedes_joint_geometry
test_issue_1166_cross_pass_feature_leaders.py::test_fixed_probe_product_budget_replays_the_exact_producer_floor
```

4673 pass; nothing outside these two modules moves.

They are not reporting a placement regression. They are reporting that
*exploration depth changed* — e.g.

```
assert [item["candidates_tried"] for item in event["items"]] == [1, 4, 1]
E   assert [8, 8, 8] == [1, 4, 1]
```

The budgets are denominated in **candidates**, on the premise that a candidate is
expensive because it builds OCC geometry. Once candidates are analytical that
premise is false: the same budget now buys ~150× more exploration, so more
candidates are tried before it bites.

This is squarely ADR 0014 Amendment 4 territory — *"a work budget must bound
measured work, not predict it"*. The caps need re-deriving for a tier whose unit
of work changed, and the amendment must be written in the same PR rather than the
guards being relaxed to match the new numbers.

## Scope

1. Give `_leader_callout_pass` an `analytical_geometry` built from
   `leader_callout_geometry` + `_text_size`, and confirm the analytical label box
   models a plain `Leader(label=<str>)` as faithfully as the callout-sketch form
   it was written for. `leaders._rendered_ink_matches` already validates the
   selected survivor against its analytical contract — keep that gate load-bearing
   for the new path, and prove it is by mutating the prediction and watching a
   named test fail.
2. Re-derive the candidate / fixed-obstacle-probe budgets so they bound measured
   work for both tiers, and update the eight guards to assert the new contract
   rather than the old constants.
3. Amend ADR 0014 (status, reasoning, date) to record that the analytical tier now
   covers machined-feature callouts and what that does to the budget unit. Note
   0014 already carries 4 amendments — if this reads as a fifth, weigh a
   superseding ADR instead.

## Gates

- The three fixtures above keep byte-identical annotations, bounding boxes and
  lint codes.
- A control part whose leaders are all hole callouts shows no change.
- `_rendered_ink_matches` is shown to fail when the analytical prediction is
  deliberately wrong.
- The eight guards assert measured work, and each is shown to fail when its cap is
  removed.

## The prototype

Applies cleanly to `43733dc`. Sizing evidence only — deliberately **not** a
proposed patch, because it leaves the eight budget guards failing and does nothing
about the ADR.

```diff
diff --git a/src/draftwright/annotations/from_model.py b/src/draftwright/annotations/from_model.py
--- a/src/draftwright/annotations/from_model.py
+++ b/src/draftwright/annotations/from_model.py
@@ -79,6 +79,7 @@ from draftwright.annotations._common import (
     carve_free_position,
     dim_footprint,
     full_strip_message,
+    leader_callout_geometry,
     place_strip_candidates,
     register_corridor,
     strip_free_span,
@@ -1930,6 +1931,16 @@ def _leader_callout_pass(
                     geom_clear=_geom_clear,
                 )
 
+            from draftwright._core import _text_size as _ts_probe
+
+            _w, _h = _ts_probe(str(label), float(dwg.draft.font_size))
+            _label_box_dims = (0.0, 0.0, _w, _h) if _w > 0 and _h > 0 else None
+
+            def _analytical(tip, elbow, _feature, *, _box=_label_box_dims):
+                if _box is None:
+                    return None
+                return leader_callout_geometry(tip, elbow, dwg.draft, callout_box=_box)
+
             collect_feature_leader(
                 ctx,
                 FeatureLeaderJob(
@@ -1939,6 +1950,7 @@ def _leader_callout_pass(
                     label=label,
                     candidates=_lane_candidates(),
                     build=_build,
+                    analytical_geometry=_analytical,
                     measurement=tuple(measurement),
                     noun=noun,
                     drop_code=drop_code,
```

The local-import and `_`-prefixed closure names are prototype shortcuts, not the
intended shape.

pzfreo:
Closes #1308

This moves every compatible machined-feature leader family onto the existing shared FeatureLeaderJob analytical solve. Candidate exploration lowers label, shaft, shelf, and rendered ink arithmetically; OCC Leader geometry is materialised only for selected survivor attempts and each rendered survivor is checked against its analytical contract before commit. Six late adapters join the canonical inventory, while required immediate consumers use the same producer floor. There is no second placement mechanism.

Budgets are re-derived as measured work: analytical and OCC candidate costs have explicit work units; fixed inventory/component/probe, pair, job, provisional, and state bounds preserve deterministic fallback floors. The existing guard families were updated to the measured-work contract, not weakened or deleted. ADR 0014 records the analytical tier, six adapters, survivor validation, and budget semantics.

Exact current-main warm-build medians (import solid once, warm once, median of three):
- CTC-01 AP242: 7.091135 -> 3.735585 s (-47.3%)
- GRM-03: 2.434761 -> 1.125315 s (-53.8%)
- hole-only wheel control: 4.026990 -> 4.068073 s (+1.0%, flat/noise)

Output/safety gates:
- full annotation name and rendered-type inventories, every per-annotation ink box to 3 decimals, and lint-code summaries are identical on CTC, GRM, and wheel
- named Y, boss, polygonal, and generic paths prove only selected analytical survivors construct OCC
- deliberately wrong machined analytical ink fails the retained rendered-survivor gate
- measured-work and fixed/pair/state fallback guards remain load-bearing with provenance
- focused current-head run: 38 passed, 3 slow tests deselected; repository static/docs/types clean
- independent exact-SHA review of 5ba3de64e74ff651e133b97c654ff35fe4a3bedb on 3d9edb80b1fc323ffe691f11d7cb1c971d17ac0d: ACCEPT, no findings; reviewer separately ran 89 focused, 7 lint/material, and 16 architecture/private checks

Slow tests were not run.

### DIFF
--- a/docs/adr/0014-collect-then-solve-annotation-placement.md
+++ b/docs/adr/0014-collect-then-solve-annotation-placement.md
@@ -93,7 +93,7 @@ selected OCC `Leader`; the rendered survivor is checked against that geometry
 before it is committed.
 
 The inventory is bounded before candidate OCC construction by job and
-expanded-candidate caps. Fixed rendered ink is itself component-bounded while
+measured candidate-work caps. Fixed rendered ink is itself component-bounded while
 it is lowered, cached once across joint/fallback use, and then guarded by the
 candidate×fixed-obstacle probe cap. Quadratic candidate geometry has a separate
 pair cap, and the exact solve retains its existing state cap. A cap
@@ -237,6 +237,35 @@ The rule this ADR now records:
 This is a placement-layer instance of the same discipline ADR 0001 applies to solvers:
 the bound has to be explainable, and "the estimate said it might be expensive" is not an
 explanation a user can act on.
+
+**Amendment 5 (2026-08-24, #1308) — machined-feature leaders use the existing
+analytical tier, and candidate budgets buy measured work.** The shared
+`FeatureLeaderJob` seam now lowers the plain-text label box, shaft and shelf for
+chamfer, fillet, flat, pocket, groove, boss, polygonal boss/stock and Y-axis step
+diameter callouts through the same `leader_callout_geometry` contract already used
+by ordinary holes. The generic pass measures each distinct label once through the
+cached renderer-faithful `_text_size`; candidate exploration constructs no OCC
+`Leader`. Only a selected survivor is materialised, and its rendered label and ink
+must match the analytical candidate before commit. A mismatch remains a
+validation-stage failure and replays the canonical producer tail. This extends the
+one shared solve; it does not introduce another placement mechanism.
+
+The old 512-candidate cap treated every candidate as equally expensive. Measurement
+on the #1308 corpus showed an OCC candidate at about 14.8 ms and an analytical one
+at about 0.1 ms. The joint-inventory guard is therefore denominated in 0.1 ms work
+units: analytical measurement costs one, OCC measurement costs 150, and the
+per-view allowance is 76,800 units. That preserves the former worst-case allowance
+of 512 OCC probes while allowing the cheap tier to explore more alternatives for
+the same measured-work ceiling. Trace records the projected joint work by view and
+the per-view limit, so a fallback states the capability it lost. Fixed-ink inventory
+components and candidate×component tests remain directly counted work (one unit per
+component/probe) under their 100,000-unit cap; they were renamed as work rather than
+relaxed. Pair and search-state guards retain their own direct-operation units.
+
+This does not supersede ADR 0014: the collect → solve → emit decision and all prior
+compatibility boundaries remain current. The change completes the analytical tier
+described by Amendment 2 and applies Amendment 4's budget rule to the now-observed
+two-tier cost model.
 
 ## Context (short — the full story is 0009's)
 
--- a/docs/adr/0014-collect-then-solve-annotation-placement.md
+++ b/docs/adr/0014-collect-then-solve-annotation-placement.md
@@ -28,7 +28,7 @@ unification remains separate work (#1166).
 
 **Amendment 2 (2026-08-15, #1166) — compatible same-view feature leaders
 share one late inventory.** Automatic and deferred sparse ordinary side/plan
-hole callouts and the five post-drain machined-feature leader families no longer
+hole callouts and the six post-drain machined-feature leader families no longer
 commit in separate local passes. They collect semantic jobs into
 `PlacementContext.feature_leaders`; the one `"feature_leaders"` stage after the
 corridor drain and before section/detail composition lowers them through
@@ -189,7 +189,7 @@ downstream furniture and table semantics. Recorded so it is not re-derived.
 
 The material term reaches every placer that weighs routing, not only the shared
 inventory: the within-pass assignment and the legacy first-clear scope in
-`_leader_callout_pass` take the same preference through the same
+`place_machined_leader_jobs` takes the same preference through the same
 `material_penalty_units`, so boss diameters and the machined families cannot drift from
 the critique either (#1187). The pre-drain diameter and grouped pattern consumers keep
 their pure legacy selection, as Amendment 1 requires.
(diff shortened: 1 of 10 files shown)


pzfreo:
Independent exact-SHA review: ACCEPT, no findings.

Reviewed head 5ba3de64e74ff651e133b97c654ff35fe4a3bedb against exact current-main base 3d9edb80b1fc323ffe691f11d7cb1c971d17ac0d across all five commits. The reviewer confirmed patch identity to the previously accepted stack, all scoped families on the shared FeatureLeaderJob path, six late adapters plus guarded producer-floor consumers, analytical geometry fidelity, survivor-only OCC, rendered-ink rejection, provenance/fallback behavior, measured-work budgets, and ADR 0014. Independent validation: 89 focused passes, 7 lint/material passes, 16 architecture/import/private canary passes, Ruff/format/diff clean. No slow tests run.

PR #1327, opened 2026-08-24
Reject disjoint convex AABBs before SAT

### LINKED ISSUE #1312: Two cheap engine costs: no broad-phase reject in convex overlap, and repeated text rasterisation
Part of #1307.

Two small, self-contained engine costs found while profiling. Neither is
architecturally interesting; both are cheap to fix and were measured, not guessed.
Filed together because separately each is too small to justify a PR.

## 1. `_convex_polygons_overlap` has no broad-phase reject

`_geometry.py:450` runs a full separating-axis test on every polygon pair it is
handed, with no cheap early-out. On one CTC-01 build it is called **365 805
times** (1.04 s cumulative), from three sites:

- `_geometry._convex_polygon_overlaps_box` — 136 013
- `annotations/leaders.py:601` genexpr — 119 532
- `annotations/leaders.py:235` genexpr — 110 260

Most pairs are far apart and rejected only after building the axis list and
projecting every vertex twice.

An AABB separating test first — strict, matching the existing "touching edges are
clear" convention — measured:

| fixture | before | after | |
|---|---|---|---|
| `nist_ctc_01_asme1_ap242.stp` | 7.77 s | 7.34 s | −5 % |
| `grm03_thumbwheel_drive_screw.step` | 2.51 s | 2.47 s | −2 % |

Annotation output identical on both.

~6 lines. The only thing to be careful about is the strictness convention: the
function's docstring commits to touching being *clear*, so the prefilter must use
`<=`, not `<`, or it changes behaviour at the boundary.

## 2. Text geometry is rasterised for strings already rasterised

`Compound.make_text` on one CTC-01 build: **674 calls for 54 distinct argument
tuples — 92 % redundant.** `7× R50` is built 192 times, `2× C5` 60 times. On
grm03 it is 219 calls for 38 distinct (83 %).

A memo keyed on the full argument tuple measured **−12 %** on CTC-01.

**Do not act on this before the leader work.** Nearly all of these calls originate
inside `leaders._measure` building throwaway annotations; if the analytical
lowering lands, they largely disappear and this memo becomes close to worthless.
Re-measure after, and only then decide whether the remainder justifies caching
inside a dependency's classmethod — which is the part that needs care, since a
cached `Compound` is mutable and shared, and the prototype paid a `deepcopy` per
hit to stay safe.

Note `_core._text_size` already does exactly this correctly for *measurement*
(`lru_cache(maxsize=512)`, pinned to the vendored font path per ADR 0006). The gap
is only in construction.

## Gates

- Identical annotations, bounding boxes and lint codes on at least three fixtures
  for each change.
- The prefilter's boundary convention is asserted by a test that fails if `<=`
  becomes `<`.
- Item 2 is re-measured after the leader work and dropped if the remaining win
  does not justify it.

pzfreo:
Closes #1312

Adds the measured strict-disjoint AABB broad phase before convex SAT. All four comparisons use `<=`, preserving the existing positive-area-overlap contract where touching edges and vertices are clear.

The named boundary test uses numbers whose subtraction raises, proving touching pairs return before SAT-axis construction; a genuine-overlap control proves real intersections still reach and pass SAT.

Post-#1308 same-host medians (import once, warm once, median 3):
- CTC-01 AP242: 3.630616s -> 3.368317s (-7.2%)
- GRM-03: 1.102130s -> 1.064957s (-3.4%)
- all-hole wheel control: 3.952388s -> 3.995277s (+1.1%, flat/noise)

The text-cache half is deliberately dropped: after analytical leader lowering its residual ~2.6% gain does not justify shared mutable dependency objects plus a deepcopy on every hit.

Validation (slow tests not run):
- 197 relevant overlap/leader/material tests pass
- all three #1308 annotation/type/3-decimal bbox/lint goldens pass
- 100% changed-line coverage
- ruff, format, mypy, codespell, strict docs pass
- deliberate `<=` -> `<` mutation fails the named boundary test
- exact-head independent review ACCEPT with no findings

Exact reviewed head: `22d9c0ad68554256a0766b38a9d7911531c9605f`
Exact base: `ba4db4f86db510b7210e1e97b7575580341c4661`

### DIFF
--- a/src/draftwright/_geometry.py
+++ b/src/draftwright/_geometry.py
@@ -456,6 +456,15 @@ def _convex_polygons_overlap(left, right) -> bool:
 
     if len(left) < 3 or len(right) < 3:
         return False
+    left_x, left_y = zip(*left, strict=True)
+    right_x, right_y = zip(*right, strict=True)
+    if (
+        max(left_x) <= min(right_x)
+        or max(right_x) <= min(left_x)
+        or max(left_y) <= min(right_y)
+        or max(right_y) <= min(left_y)
+    ):
+        return False
     axes: list[tuple[float, float]] = []
     for points in (left, right):
         axes.extend(
--- a/tests/test_issue_1312_engine_costs.py
+++ b/tests/test_issue_1312_engine_costs.py
@@ -1,0 +1,25 @@
+"""Measured small engine costs from #1312."""
+
+from draftwright._geometry import _convex_polygons_overlap
+
+
+class _AabbOnlyNumber(float):
+    """Fail if a touching-box rejection leaks through to the expensive SAT."""
+
+    def __sub__(self, other):
+        raise AssertionError("touching AABBs must return before building SAT axes")
+
+
+def test_convex_overlap_broad_phase_rejects_touching_aabbs_before_the_sat():
+    number = _AabbOnlyNumber
+    left = ((number(0), number(0)), (number(1), number(0)), (number(0), number(1)))
+    right = ((number(1), number(0)), (number(2), number(0)), (number(1), number(1)))
+
+    assert not _convex_polygons_overlap(left, right)
+
+
+def test_convex_overlap_broad_phase_keeps_real_overlap_for_the_sat():
+    left = ((0.0, 0.0), (2.0, 0.0), (0.0, 2.0))
+    right = ((0.5, 0.5), (2.5, 0.5), (0.5, 2.5))
+
+    assert _convex_polygons_overlap(left, right)


pzfreo:
Independent exact-SHA review: **ACCEPT — no findings**.

Reviewed exact head `22d9c0ad68554256a0766b38a9d7911531c9605f` against exact base `ba4db4f86db510b7210e1e97b7575580341c4661`; clean one-commit, two-file range.

Independently verified all four `<=` directions, touching-is-clear semantics, deliberate `<` mutation failure, genuine-overlap SAT control, absence of text-cache/mutable-object changes, and the #1308 annotation/type/3dp-box/lint goldens. Reviewer validation: 8 focused/output-identity tests, 15 import/private/architecture canaries, Ruff/format/diff hygiene; all green. Slow tests not run.

Issue #129, opened 2026-06-21
Place pitch-dim labels clear of vertical centerlines at creation time (not via the repair loop)


pzfreo:
## What

When the #92 pitch dimensions (`dim_pitch_*`, the `(n-1)× pitch` labels for `LinearArray`/`RectGrid` rows) land with their **label text on a vertical centerline**, offset the label off the centerline **at creation time** in the pitch-dim placement (`_place_pitch_dim` / `_add_grid_pitch_dims`), so `label_centerline_overlap` never fires. Reuse the helpers' `_compute_label_offset_x` clearance solver, passing the resulting `label_offset_x` into `_dim(...)`.

## Why this, not the repair loop (supersedes #119)

#119 tried to clear these in the bounded lint→repair loop after the fact. That approach is the wrong layer and no longer works:

- **ADR 0002**: the repair loop is a *safety net*, not the primary placement mechanism — placement belongs in the annotation passes.
- The repair matched dims to lint issues **by label text**, which is unreliable now: on CTC-02 the 15 `label_centerline_overlap` warnings are mostly **empty labels** (~8), **duplicate pitch-dim labels** (`2× 75` appears on several dims), and a non-dim section marker (`A`). The label lookup can't resolve any of those, so the #119 branch cleared **0 of 15**.
- Fixing at creation avoids the shared repair path entirely (lower regression surface) and the whack-a-mole of shifting a label only to land it on another line.

## Scope

- Most of CTC-02's 15 overlaps are pitch-dim labels → handle those in `_place_pitch_dim`/`_add_grid_pitch_dims`.
- Some overlaps come from other auto-placed dims (location/step dims) with empty labels — evaluate whether the same creation-time offset helper should be applied where those dims are placed.
- These are **warnings, not errors** — purely a legibility win. Low priority.

## Context

- Blocker is gone: helpers#161 (lint O(n²) → ~60× faster) shipped in the 0.12.0 we now depend on; a full CTC-02 build with auto-repair is ~122s (the ~70s *first* lint is the separate perf item).
- The drafting rule: a dimension *line* crossing a centerline is fine; it's the **label ink** on the centerline that hurts legibility — which is exactly what `label_centerline_overlap` flags.

pzfreo:
Closing as implemented by PR #551 (`7b250a2`). Pitch-dimension labels are cleared from centerlines/bolt circles at creation time, with regression coverage in `test_make_drawing.py`.

PR #1335, opened 2026-08-25
Prevent same-batch annotation ink collisions

### LINKED ISSUE #1334: Prevent annotation ink collisions in the corridor solve
Direct prevention follow-up to the Stage 1 detector landed by #1322. This issue does not depend on #1333: repair remains a useful safety net for residual or externally introduced collisions, but prevention can be implemented and measured independently.

## The point

Stage 1 reliably detects line-work crossing annotation text. This issue stops those collisions being placed. If the solver knows about foreign line-work and arrowheads, the finding never exists: nothing to report, repair, or undo.

#1333 remains open for collisions that cannot be prevented, labels constrained by pins or span clamps, and post-placement edits.

## What is missing today

The corridor solve is collect-then-solve (ADR 0014): it builds an obstacle model and places against it before lint runs. Lint is the audit afterwards, and layout never consults it. The late `reconcile_witness_labels` pass is the one exception, but it only repairs a subset after placement.

Stage 1 demonstrates that arrowheads and dimension lines are not represented as blockers for foreign labels. A label may therefore be placed where another annotation line already runs.

## What this is NOT

Do not invoke the Stage 1 lint pair scan inside candidate evaluation. Stage 1 now uses deterministic segment clipping rather than OCC booleans, but running an audit scan for every candidate would still multiply pairwise work and couple placement to lint.

Extend the obstacle model with conservative extents for line-work and terminators that the placement pipeline already computes. Candidate rejection should remain cheap rectangle and segment arithmetic within the existing solve.

## The check becomes the verifier

Measure the Stage 1 finding count before and after on the corpus with repair disabled. The count should collapse because prevention, not repair, removed the defects. This matches ADR 0014 Amendment 3: budgets measure outcomes rather than predict them.

## Acceptance

- Stage 1 corpus findings drop substantially without repair running.
- Focused regressions prove labels avoid foreign dimension lines and arrowheads.
- Pinned or otherwise infeasible placements retain deterministic fallback behavior and lint evidence.
- No material build-time regression.
- Both supported build123d kernels agree on the placement verdict.

pzfreo:
## Summary

- add a bounded same-batch dimension-label selector using public label, segment, and terminator metadata
- run it across the complete corridor survivor batch and immediate turned step-length chains
- preserve pinned candidates and retain deterministic, lint-visible fallback when no clear alternative exists
- keep already-clean batches on an object-identical zero-construction fast path

## Results

- GRM-03 step-chain ink findings: 2 -> 0
- crowded flange corpus ink findings: 4 -> 2, removing the two same-batch crossings while retaining cross-stage findings

## Validation

- `uv run pytest -q -n0 tests/test_issue_1334_prevent_ink.py`
- focused corridor placement tests: 11 passed
- analytical fixture suite: 6 passed
- smoke tier: 91 passed
- Ruff, formatting, mypy, and `git diff --check`

Slow tests and the alternate-kernel CI matrix were intentionally left to CI.

Closes #1334

### DIFF
--- a/src/draftwright/annotations/_common.py
+++ b/src/draftwright/annotations/_common.py
@@ -21,11 +21,13 @@ from build123d_drafting.helpers import DEFAULT_FONT_PATH, Dimension, Note, SafeD
 from draftwright._core import (  # noqa: F401 — _anno_box re-exported (#700)
     _anno_box,
     _decode_hole_location_fact,
+    _dim,
     _text_size,
     place_annotation,
 )
 from draftwright._geometry import (  # noqa: F401
     _boxes_overlap,
+    _segment_clip_extent,
     _segment_clips_box,
     _segment_crosses_box,
 )
@@ -38,6 +40,14 @@ from draftwright.linting.structural import (
 from draftwright.model.compiled import resolve_feature
 
 _log = logging.getLogger(__name__)
+
+
+# A line entering a label by less than this is a corner graze, not obscured text.  Keep
+# this equal to the Stage-1 ratchet's measured floor without importing that late critique
+# into the placement DAG: candidate selection owns its arithmetic and Stage 1 remains a
+# backstop, not the oracle invoked for every alternative (#1334).
+_LABEL_INK_CROSSING_MM = 0.5
+_LABEL_INK_CLEARANCE_MM = 0.25
 
 
 def _with_hole_location_coverage(annotation, coverage):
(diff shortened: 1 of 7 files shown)


pzfreo:
Addressed in bd7dfb1. Rebuilt label candidates now retain producer-attached semantic riders (`covers_*` and non-placement `_dw_*` metadata), while keeping the new `_dw_spec`. This fixes the hole-completeness and transactional hole-table regressions. I also updated the Policy-B regression from two ink findings to the intended remaining one.

Focused validation:
- all 15 retained-balloon transactional cases pass
- hole-table escalation rollback passes
- Policy-B crossing regression passes
- #1334 focused tests pass, including a new semantic-evidence preservation assertion
- Ruff, formatting, mypy, and diff checks pass

src/draftwright/annotations/_common.py
```diff
@@ -1458,6 +1468,280 @@ def box_within_page_and_clear(bb, page_box, obstacles) -> bool:
...
+            return dim
+        axis, _other, spec = info
+        direction = 1.0 if float(spec.p2[axis]) >= float(spec.p1[axis]) else -1.0
+        kwargs = dict(spec.kwargs)
+        kwargs["label_offset_x"] = (
+            kwargs.get("label_offset_x", 0.0) + (centre - natural) * direction
+        )
+        rebuilt = _dim(spec.p1, spec.p2, spec.side, spec.distance, spec.draft, **kwargs)
+        for attr in ("_dw_scale", "_dw_label_value"):
+            if getattr(dim, attr, None) is not None:
```
pzfreo: **Confirmed (verified end-to-end): rebuilt survivors silently lose coverage provenance.** `_rebuild` copies only `_dw_scale`/`_dw_label_value`, but corridor hole-location candidates attach `covers_hole_locations`/`covers_hole_centers` inside the build closure (`_with_hole_location_coverage` — from_model.py:673, holes.py:854/873/1040), and `_dw_spec` records only `p1/p2/side/distance/draft/kwargs`, so the spec cannot regenerate them. The rebuilt dim goes straight to `ctx.place` with no re-attachment. `hole_coverage.py` populates location evidence **only** from `getattr(annotation, "covers_hole_locations", ())` and has no geometric fallback for `location.*` parameters (unlike `coverage.py`, which does), so a conflicted location batch — exactly the #1334 trigger this PR exists for — ships a sheet that visibly carries the dimension while `lint_hole_coverage` reports `hole_requirement_missing` for it.

The two-attribute allowlist is also the second hand-maintained copy of the "metadata that must survive a rebuild" list (`repair._replace_dim` carries the same two, and its inline comment records `_dw_label_value` being missed at that very seam once). Suggest one shared rebuild-with-carry helper, with the carried-attribute roster including the coverage facts — or better, fail closed: copy every `_dw_*`/`covers_*` attribute rather than an enumerated pair.

src/draftwright/annotations/_common.py
```diff
@@ -1458,6 +1468,280 @@ def box_within_page_and_clear(bb, page_box, obstacles) -> bool:
...
+                        < tip[axis]
+                        < label[axis + 2] - _LABEL_INK_CLEARANCE_MM
+                        and label[other] - 1e-6 < tip[other] < label[other + 2] + 1e-6
+                    ):
+                        found.add(("arrow", source, tip_index, target))
+            for source, source_segments in enumerate(segs):
+                if source == target:
+                    continue  # the helper deliberately cuts its own line around its label
+                for segment_index, segment in enumerate(source_segments):
+                    if _line_depth(segment, label) >= _LABEL_INK_CROSSING_MM:
```
pzfreo: **Confirmed: the preventer and Stage-1 lint measure crossings differently, so the preventer can declare clean what lint still flags.** `_conflicts` gates each foreign segment independently (`_line_depth(seg, label) >= 0.5`), but `linting.ink_overlap.crossing_length` unions collinear segments joined at an endpoint **inside** the label before comparing the joined total to `MIN_CROSSING_MM`. A stroke entering a label as two joined 0.3 mm pieces is 0.6 mm to lint and invisible here — the batch passes through unshifted and `annotation_ink_overlap` still fires, i.e. the preventer declines to fix a fixable, linted defect. Same class of drift in the other direction: lint's `crossable_region` excludes diagonal-dim AABBs (a known false-positive source) while `_conflicts` has no such gate.

Reusing `crossing_length` + `crossable_region` here would eliminate both drifts at once — see the constants comment below for why that import is legal.

src/draftwright/annotations/_common.py
```diff
@@ -40,6 +42,14 @@
 _log = logging.getLogger(__name__)
 
 
+# A line entering a label by less than this is a corner graze, not obscured text.  Keep
+# this equal to the Stage-1 ratchet's measured floor without importing that late critique
+# into the placement DAG: candidate selection owns its arithmetic and Stage 1 remains a
+# backstop, not the oracle invoked for every alternative (#1334).
+_LABEL_INK_CROSSING_MM = 0.5
```
pzfreo: **The stated reason for duplicating the constant is factually wrong.** The comment keeps `_LABEL_INK_CROSSING_MM = 0.5` "equal to the Stage-1 ratchet's measured floor without importing that late critique into the placement DAG" — but `_LAYERS` in test_import_boundaries.py ranks `linting` at 2 and `annotations` at 4, so `from draftwright.linting.ink_overlap import MIN_CROSSING_MM` is a legal *downward* import; this module already imports `linting.issues`/`linting.structural` at the top, and this PR's own test imports `draftwright.linting.ink_overlap` to validate the result. `MIN_CROSSING_MM` carries a long measured-calibration why-trail; the moment corpus evidence moves it, this copy drifts silently and router and lint disagree — the exact hazard ADR 0014 Amendment 3 resolved for the material field by sharing one lowering. Import the constant (or pin equality with a guard test). Note also `_LABEL_INK_CLEARANCE_MM = 0.25` has **no** Stage-1 counterpart, so the comment's "keep them in step" framing points at a value that doesn't exist.

src/draftwright/annotations/from_model.py
```diff
@@ -3686,8 +3687,21 @@ def _clear(items):
...
-    # Room guard: if any dim would fall off the drawable page, place NONE.
     page = (_MARGIN, _MARGIN, dwg.page_w - _MARGIN, dwg.page_h - _MARGIN)
+    # The chain is one placement batch: until commit, no sibling's extension line or
+    # terminator exists in strip occupancy.  Select small along-line label offsets against
+    # the complete batch before the room guard (#1334).  Measurement provenance stays paired
+    # by name; only the rendered Dimension survivor changes.
+    measurements_by_name = {name: measurements for name, _dim_obj, measurements in candidates}
+    candidates = [
+        (name, dim, measurements_by_name[name])
+        for name, dim in prevent_dimension_label_ink(
```
pzfreo, line 3698: **Confirmed: the step-chain call site validates shifted labels against nothing but the page room guard.** Unlike the strip path (whose revalidation checks blockers/forbid/out_of_band, and whose in-band committed boxes are carved out of the free segments so a shift can't reach them — verified, no finding there), this call runs prevention with only `page=` and then places. `step_lengths` runs after hole callouts, locations, envelope and diameter passes, so committed OD/envelope witness lines and leaders are real neighbours; `_centres_for` can move a label up to span-end ± 0.5 mm — positions the pre-PR code could never occupy — onto that committed ink, manufacturing a *new* `annotation_ink_overlap` the natural placement didn't have. The prevention model can't see it (`_conflicts` enumerates the batch only). Suggest passing the committed-ink obstacle boxes into `prevent_dimension_label_ink` for this call site, as ADR 0014 Amendment 2's late stage does with its named committed blockers.

src/draftwright/annotations/_common.py
```diff
@@ -2258,16 +2543,45 @@ def _reject(name, reason):  # trace-only (#736): why this candidate left this se
...
+
+    if len(solved) > 1:
+        natural_solved = dict(solved)
+        adjusted = prevent_dimension_label_ink(
+            solved,
+            page=(
+                (0.0, 0.0, float(dwg.page_w), float(dwg.page_h))
+                if hasattr(dwg, "page_w") and hasattr(dwg, "page_h")
+                else None
+            ),
```
pzfreo, line 2597: **Confirmed: page-bounds mismatch between the two call sites.** This site passes the full physical page `(0, 0, page_w, page_h)`; `_draw_step_chain` passes the `_MARGIN`-inset rect. `_centres_for` uses `page` solely to clamp candidate centres, and nothing in the revalidation covers margins (out_of_band holds only committed annotation boxes; the title block isn't committed yet during corridor passes — the stated reason `forbid` exists). So when a view is packed against the margin and the only conflict-clearing centre is a span end, the shifted label can overhang into the 10 mm margin, clamped only at the sheet edge — where the same shift on the chain path stops at the margin line. Minimal fix: pass the same inset rect here.

src/draftwright/annotations/_common.py
```diff
@@ -1458,6 +1468,280 @@ def box_within_page_and_clear(bb, page_box, obstacles) -> bool:
...
+                choices.add(clipped[axis + 2] + half + _LABEL_INK_CLEARANCE_MM)
+        bounded = {min(max(value, lo), hi) for value in choices if math.isfinite(value)}
+        # Bound exceptional work.  Nearest candidates are the meaningful drafting
+        # alternatives; the two association bounds remain present explicitly.
+        nearest = sorted(bounded, key=lambda value: (abs(value - current_centre), value))[:16]
+        return tuple(dict.fromkeys([*nearest, lo, hi]))
+
+    # Deterministic steepest descent.  Every accepted move strictly improves the
+    # lexicographic objective, so it terminates even when the layout is infeasible.
+    for _iteration in range(max(1, 2 * len(current))):
```
pzfreo, line 1752: **Efficiency: the conflicted path does unbudgeted OCC work, against this file's own precedents.** Every trial calls `_objective(trial)` → `_conflicts(batch)`, which rebuilds labels/segments/tips for the **whole** batch (helpers' `label_bbox`/`segments` are recompute-on-access properties doing OCC transforms) and runs the full O(n²·s) scan, though exactly one dim changed — with steepest descent that's ~36·n²·s work per accepted move and O(n⁴·s) worst case. Worse, each novel `(index, centre)` cache miss in `_rebuild` pays a full helpers `Dimension` construction (OCC `Text` render through the pinned font, ~10–50 ms), and `_centres_for` re-derives centres from moving batch state so keys proliferate across iterations. `dim_footprint` exists in this file precisely because "a rejected candidate must not pay the rendering cost" (#602), and ADR 0014 Amendment 4 asks for a live counter on **measured** work, not a predicted iteration cap. Trial evaluation needs only the natural box translated by `(centre − natural)` along the axis — closed form; build real Dimensions only for the finally selected survivors (≤ n builds), and memoise per-object (box, segments, tips) keyed by `id(dim)`.

src/draftwright/annotations/_common.py
```diff
@@ -2258,16 +2543,45 @@ def _reject(name, reason):  # trace-only (#736): why this candidate left this se
...
+        )
+        # A label shift normally stays inside the dimension's measured span and
+        # therefore inside its already-validated footprint.  Keep the validation
+        # contract explicit nevertheless: if an unusual helper/style grows the real
+        # box into a hard obstacle, retain the natural survivor.  Stage-1 lint then
+        # records the infeasible ink conflict instead of this prevention step trading
+        # it for a structural collision.
+        solved = []
+        for name, dim in adjusted:
+            real = _geom_box(dim)
```
pzfreo, line 2612: Two cheap wins here: (1) on the overwhelmingly common clean path `prevent_dimension_label_ink` returns the identical objects, yet this loop still re-runs `_geom_box` (uncached OCC bounding_box) plus three `_box_hits` scans per survivor — all already computed at the accept phase above for these exact objects. Skip entries where `dim is natural_solved[name]`. (2) The three predicates are a copy-with-variation of the accept-phase checks ~40 lines up (same blockers/forbid/out_of_band trio, no per-reason trace entries this time); the next obstacle class added — as forbid (#481) and out_of_band (#679) each were — must be remembered at both sites. Extract one shared `_real_box_conflict(real) -> reason | None`. Also worth a one-line assert: `natural_solved = dict(solved)` silently assumes batch-unique names — a duplicate would restore an invalid shift with the *other* candidate's dim.

pzfreo (commented):
Adversarial review of the head (e8f4a7d). Eight findings survived verification — two mechanisms were checked end-to-end and confirmed by independent verification passes; one candidate finding (cross-batch blindness on the **strip** path) was refuted during verification and is *not* reported: `strip_obstacles`/`out_of_band`/`blockers` fully represent committed ink there, and a label shift along the dim line cannot change the stacking-axis extent, so the shifted label stays disjoint from in-band committed boxes exactly as the natural one was. The step-chain path is a different story (inline comment).

**Architectural note (no diff line to anchor on):** this PR adds a new placement stage — a post-solve same-batch re-selection that can rebuild survivors with shifted labels — with no ADR 0014 amendment. ADR 0014 is the placement model "as it exists in the code today"; its Emit step still says place_strip_candidates "builds each survivor once and re-validates its real box", which this PR makes false, and every prior late stage (Amendments 1–3) earned an amendment. CLAUDE.md: amend the ADR in the same PR rather than silently diverging. The 'Stage-1/Stage-3' vocabulary the comments and tests introduce also appears nowhere in docs/, so it can't be looked up.

Also noted while verifying: the local-search shape (`_centres_for` + steepest descent) is the per-obstacle candidate design that `clear_label_of_centerlines`'s own docstring (500 lines up, #129 second review) records as rejected in favour of the joint interval carve (`carve_free_segments`); reusing the carve to compute feasible-centre intervals directly would remove the iteration bound and the candidate-set heuristics.

src/draftwright/annotations/_common.py
```diff
@@ -1458,6 +1468,280 @@ def box_within_page_and_clear(bb, page_box, obstacles) -> bool:
...
+    shift can keep a shared witness out of the digits without making the label read as its
+    neighbour's.  If the bounded choices cannot improve the batch, the natural deterministic
+    placement survives and the normal ``annotation_ink_overlap`` lint remains explicit
+    evidence of the infeasible fallback.  Names in *immutable* are never shifted (pins win).
+
+    Returns ``[(name, dimension), ...]`` in input order.
+    """
+
+    original = list(dimensions)
+    if len(original) < 1:
```
pzfreo: Nit: the guard should be `len(original) < 2` — every conflict token requires two distinct batch members, so a singleton can never conflict, and the real precondition currently lives in one caller's `len(solved) > 1` wrapper while the `_draw_step_chain` site has no guard and pays the full `_axis_info`/`_conflicts` walk for a one-dim chain. Moving the `< 2` inside makes the contract self-contained and lets the strip-path wrapper condition shrink.

pzfreo:
Re-reviewed bd7dfb1 ("Preserve semantic evidence across label selection").

**Verdict: sound fix for the coverage-rider finding — the most severe one.**

- The prefix-based copy (`covers_*` + `_dw_*`, excluding `_dw_spec`) is the fail-closed shape the review asked for: nothing enumerable to forget as the rider vocabulary grows, and `_dw_spec` correctly stays the rebuilt placement's own. I hunted for riders outside those prefixes: the only family is `hole_representation`/`hole_representation_reason`, which is assigned solely in `_restore_annotation_transaction` on already-committed annotations — post-placement, so it cannot exist on a dim at `_rebuild` time. No leak.
- The `test_issue_1166` expectation change (annotation_ink_overlap 2 → 1) checks out as honest: the 2 predates this PR, main's own comment anticipated stage 3 removing the same-batch crossing, and the survivor is the cross-stage Policy-B leader conflict that same-batch prevention cannot address by design.
- Independently ran the focused tests on the head: `tests/test_issue_1334_prevent_ink.py` (including the new rider-survival assertion) and the 1166 Policy-B regression — 4 passed.

**Still open from the review** (not blockers on the rider bug's scale, but the first two are cheap and worth doing before merge):
1. ADR 0014 amendment for the new stage (its Emit text — "builds each survivor once" — is now false).
2. Import `MIN_CROSSING_MM` from `linting.ink_overlap` instead of the `_LABEL_INK_CROSSING_MM` copy; the comment's DAG rationale is factually wrong (linting is rank 2, below annotations).
3. Per-segment vs unioned crossing measurement drift against Stage-1 lint.
4. Step-chain shifts validated against nothing but the page room guard (committed ink invisible).
5. Margin-clamp mismatch: strip path passes the physical page, chain path the `_MARGIN` inset.
6. Efficiency: per-trial OCC Dimension builds / whole-batch recompute in the conflicted path; redundant `_geom_box` revalidation on the clean path.

pzfreo:
Addressed the review findings in `930b158`, with two suggestions deliberately not applied verbatim.

Fixed:
- prevention now imports `MIN_CROSSING_MM` and uses lint's `crossable_region` + connected-stroke `crossing_length`; the joined 0.3 + 0.3 mm regression is pinned
- immediate step-chain label moves cannot introduce contact with committed annotation ink
- both entry paths clamp to the common `_MARGIN` drawable page
- singleton and duplicate-name invariants are explicit
- label/segment/tip metadata is cached per built alternative
- unchanged clean-path survivors skip the redundant `_geom_box` revalidation; the repeated hard-obstacle checks are one helper
- the remaining flange expectation is corrected from 2 to the measured 1 cross-producer finding
- ADR 0014 Amendment 6 documents the new Emit refinement, its bounds, fallback, evidence preservation, and exceptional rebuilds

I did **not** replace trial geometry with a translated natural label box. `label_offset_x` also moves the helper's dimension-line cutout, so the source segments change; that coupled change is what clears GRM-03. Box-only scoring would accept/reject against geometry that will not be emitted. For the same reason, a one-shot interval carve over fixed obstacles is not equivalent here: moving one label changes the obstacle line-work seen by its peers. The bounded full-batch rescore stays, with per-object metadata and candidate construction caches. A genuinely analytical lowering would need to model the helper's cutout/terminator semantics, not only translate an AABB.

Validation (no slow tier):
- 275 smoke/unit tests passed
- focused #1334, connected-segment, committed-ink, step-chain, Policy-B and round-trip regressions passed
- import boundaries passed
- Ruff, format, mypy and diff checks passed

PR #1345, opened 2026-08-27
Reproducible SVG/DXF/PDF export, opt-in via reproducible=

clairbee:
## What this is

Two exports of one drawing differ in bytes, which makes a written drawing useless
as something to check in and diff. This adds `reproducible=` to make them agree.

The first two commits on this branch were a first attempt. They broke every DXF
export: `_export_shape` was handed a pre-decomposed **list**, but it only calls
`add_shape` directly on the `ExportSVG` branch — for `ExportDXF` it falls through
to `_elements(shape)`, which opens with `shape.faces()`. 40 tests failed on the
fast tier. The third commit fixes that and finishes the feature.

## The three sources of drift, and what settles each

| | |
|---|---|
| **Element order** | Not stable between runs. DXF ordered *going in* (`_elements(ordered=True)`); SVG settled *coming out* (`canonicalize_svg`). |
| **The clock** | ezdxf's fixed-metadata option, plus the creation marker it misses; reportlab's `invariant`. |
| **A set iteration** | ezdxf builds CLASSES by iterating a `set[str]` — pre-seeded sorted. |

The SVG/DXF asymmetry is deliberate. `ExportDXF` writes an entity per element as
it converts and the handles follow, so nothing after the fact can fix it. An SVG
is sorted in the written file, so it is handed its shapes unordered even when the
flag is on — ordering it too would pay the expensive half twice for a result
`canonicalize_svg` already reaches alone.

## Off by default, because ordering is not free

Interleaved, 9 runs each, 358-part sheet:

```
reproducible=False   0.452s      (origin/main: 0.485s)
reproducible=True    0.597s      +32%
```

Off costs what it did before the option existed. The metadata pinning is the cheap
half (~1 ms); both hang off the one flag, because a caller wanting a stable file
wants both and should not have to know which one costs.

## API

```python
dwg = build_drawing(part, reproducible=True)      # default the Drawing carries
dwg.export("out", formats=("dxf",))
dwg.export("out", formats=("dxf",), reproducible=False)   # override per call
```

`build_drawing` is the seam `partcad/partcad-render-draftwright` needs: it forwards
configured options there, then calls `drawing.export(stem, formats=(fmt,))` with no
keywords, so the setting has to survive the build. Verified with their exact call
shape across two interpreters — on, the DXF is identical; off, it differs at no
cost. They need one line on their side: `"reproducible"` in `DRAWING_PARAMETERS`.

## Defects found and fixed along the way

- **`canonicalize_svg` was not canonical.** It sorted each element's `tail` — the
  whitespace *after* it — with the element, so the last slot's shorter indent rode
  along with whichever element sorted last. Three permutations of one three-element
  layer produced three different files. Latent on real sheets, so only the unit
  test catches it.
- **`CREATED_BY_EZDXF` still carried a live timestamp.** ezdxf's option cannot
  reach it: it is stamped when the document is built, long before an export asks.
- **A process-wide `ezdxf.options` flag was left flipped** for every other ezdxf
  user in the interpreter. Now restored in a `finally`.
- **`_geometry_key` was not a total order.** Hidden-line removal emits overlapping
  silhouettes as reversed duplicate pairs agreeing on bounding box, geometry type,
  edge count, length and vertex set. On the front hidden layer of the test fixture
  **all 12 tie-groups held genuinely different geometry**, so their order fell back
  to the kernel's. The key now carries directed endpoints — what the exporter
  actually writes.
- **Duplicated decomposition removed.** `_canonical_parts` in `drawing.py` was
  `_elements` plus a sort, one DAG rank too high. `drawing.py` is now a 4-line diff
  against main.

## A corrected claim

The premise holds, but not for the reason originally given. The kernel's walk order
does vary (1 of 17 shapes reorders on a plain plate) but does **not** reduce to
`PYTHONHASHSEED` — with the sort removed on purpose, a fixed seed pair caught it in
3 of 4 trials. Docstrings now say "not stable between runs", and the end-to-end gate
is labelled a sampling detector. It never false-fails with the fix (6 of 6).

`AGENTS.md` said byte-identical output "is not a goal". That conflated two claims and
is now split: output is still not stable across versions or layout changes (ADR
0004/0012), but `reproducible=True` promises two exports of one drawing, on one
version, agree.

## Verification

- **Fast tier: 4752 passed.** 30 new tests.
- **16 mutations, one per guard, all caught.** One initially survived — flipping
  `Drawing.__init__`'s default is invisible through `build_drawing`, which passes the
  value explicitly — so the public constructor's default got its own test.
- **Changed-line coverage 97%** (gate 93%). The 2 uncovered lines are the defensive
  `except` that keeps a metadata failure from breaking an export.
- `pr-check --static` clean.

### Pre-existing failures, not from this branch

All reproduce with these files reverted to `main`:

- `test_workflows.py::test_version_updater_...` ×3 — the local `uv` rejects
  `uv version --no-sync`. Environmental.
- `test_export_dxf_curves.py::test_hole_table_dxf_conversion_avoids_bulk_curve_arrays`
  — 5s cap, 3.4–3.8s standalone on both main and this branch; fails only under
  `-n auto` load.
- `test_slanted_blind_step.py::test_crowded_step_warning_...` — 120s cap, ~41s
  standalone on both (main 41.2–42.4s, branch 40.5–40.7s); exceeded it only under
  `--cov` plus 16-way parallelism.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -104,6 +104,20 @@ for issue in dwg.lint():
 paths = dwg.export("out", formats=("svg", "dxf", "pdf", "png"))   # -> {format: path}
 ```
 
+`reproducible=True` writes files that do not carry the run that produced them, so two
+exports of the same drawing are byte-identical and a written drawing can be diffed or
+checksummed to see whether its content actually changed. Set it per call, or once for
+the drawing via `build_drawing(..., reproducible=True)`:
+
+```python
+dwg = build_drawing(part, reproducible=True)     # every export from this drawing
+paths = dwg.export("out", formats=("dxf",))
+paths = dwg.export("out", formats=("dxf",), reproducible=False)   # override per call
+```
+
+Off by default because it is not free: ordering the DXF entities costs roughly a third
+of DXF export time again. Off costs exactly what it did before the option existed.
+
 ## What NOT to do
 
 - ❌ Raw page coordinates for feature annotations; `Drawing.place_dim(...)` (deprecated raw
@@ -111,7 +125,10 @@ paths = dwg.export("out", formats=("svg", "dxf", "pdf", "png"))   # -> {format: 
 - ❌ `dwg.add(Dimension/Leader/FeatureControlFrame(...))` / raw `Note` objects — `add` is the
   engine's low-level placement primitive (being made private, #817). Use the verbs: `note()` for
   free text, `add_table()`/`add_hole_table()` for tables, the feature verbs for callouts/dims.
-- ❌ Bypassing recognition / assuming byte-identical output (it is not a goal — ADR 0004/0012).
+- ❌ Bypassing recognition / treating byte-identical output as a *stability* guarantee across
+  versions or layout changes — it is not one, and pinning a byte digest to catch regressions is
+  the wrong instrument (ADR 0004/0012). `reproducible=True` is a narrower promise: two exports
+  of one drawing, on one version, agree. It does not promise the next version draws it the same.
 
 ## Source of truth
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -272,6 +272,35 @@ IR, generation, and drawing code must not depend on benchmark expectations or sc
   `Drawing.export(out, *, formats=("pdf",)) → {format: path}` is the front door
   (the legacy `svg=`/`dxf=` tuple form + `export_pdf` are back-compat/deprecated
   wrappers). Sits below `make_drawing.py`, above `_core.py`.
+
+  **`reproducible=` — byte-identical exports, opt-in.** On, two exports of one
+  drawing are identical, so a checked-in drawing diffs cleanly and a caller can
+  see when its output really changed. Three things are settled to get there: the
+  clock and GUIDs an exporter stamps (ezdxf's fixed-metadata option — restored
+  after the save, it is process-wide state — plus the creation marker that option
+  misses, and reportlab's `invariant`), ezdxf's CLASSES section (built by
+  iterating a `set[str]`, so it is pre-seeded sorted), and the order the kernel
+  hands parts over, which is not stable between runs and does **not** reduce to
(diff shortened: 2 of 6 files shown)


pzfreo:
Hi Roman — thanks for this, and welcome. I'll come back with review notes on the change itself; one housekeeping item first.

draftwright is dual-licensed: AGPL-3.0 plus separate commercial terms, so contributions need to be covered by the [CLA](https://github.com/pzfreo/draftwright/blob/main/CLA.md), which grants me the right to relicense contributed code (§2). It's accepted by conduct — opening a PR indicates agreement — but as this is the project's first outside contribution I'd like it on the record explicitly.

You're contributing from `rkuz@partcad.org` and this delivers the seam `partcad/partcad-render-draftwright` needs, so a comment covering both capacities would settle it. Something like:

> I have read draftwright's CLA and agree to it for this Contribution. I am authorised to act for PartCAD in this respect and agree to it on PartCAD's behalf as well.

If PartCAD's structure means you'd rather not bind the entity in a GitHub comment, email works equally well — paul@fremantle.org.

Separately, the branch is behind `main` — a rebase or merge from upstream will get CI running cleanly.

clairbee:
Hi Paul!

Happy to help!
I have read draftwright's CLA and agree to it for this Contribution.

The seam you mentioned is working just fine today and makes draftwright available on the PartCAD platform (though the required PartCAD revision is not released just yet). This Contribution is for this repository only, to allow dropping draftwright from the blacklist of known non-reproducible tools in PartCAD examples.

PartCAD and its seams have nothing to do with this CLA. I'd be happy if you take ownership of that seam and be in control of how your tool is presented to PartCAD users.

pzfreo:
Thanks, Roman — understood. We’ll leave ownership of the PartCAD seam with you and focus on the draftwright side.

Many thanks for the contribution and for getting draftwright unblocked for reproducible PartCAD examples. The PR is now merged — very much appreciated!

PR #1354, opened 2026-08-27
Preserve authored dimension display precision

### LINKED ISSUE #1349: Authored feature dimensions need precision-preserving display control
## Summary

A feature-linked authored dimension can carry the correct numeric nominal but still print an unsafe rounded value. The Gramel CNC quotation exercise on Draftwright 0.4.15 exposed this on legitimate manufacturing nominals including `13.55`, `6.25`, `4.75`, `0.85`, and `1.875` mm.

The current workaround is `measured_dimension(..., value=..., label="13.55", ...)`. That restores the printed value, but it restates the number and gives up the main benefit of the referential DSL: the displayed requirement is no longer read from the feature/parameter that owns it.

## Why this is a trust issue

A drawing can compile, lint and look plausible while changing a CNC nominal through formatting. `13.55 -> 13.6` or `0.85 -> 0.9` is not presentation-only when the PDF is the quotation/manufacturing authority.

This is distinct from tolerance rendering. The raw nominal remains correct; the unsafe change happens at the display boundary.

## Desired surface

One or both of:

```python
sheet.dimension(envelope, "width.length").format(decimals=2)
sheet.dimension(envelope, "width.length").display("13.55")
```

The second spelling, if supported, should retain the numeric/reference identity and treat the string only as approved display text.

An automatic policy could preserve the precision of an authored nominal or parameter source, but an explicit per-dimension override is still needed.

## Acceptance

- Feature-linked authored dimensions can request display precision without becoming measured/restated dimensions.
- Regression coverage includes `13.55`, `6.25`, `4.75`, `0.85`, and `1.875`.
- Numeric reconciliation, tolerances, suppression and provenance continue to operate on the numeric parameter identity, not by parsing the label.
- Generated Sheet scripts round-trip the display policy.
- Automatic/detected drawings retain their current default formatting unless a policy is explicitly changed.

Related: #927 (renderer label provenance), but this issue is the user-facing precision contract rather than renderer reformatting ownership.

pzfreo:
## Symptom and evidence

- User-visible problem: feature-linked authored dimensions rounded manufacturing nominals such as `13.55` to `13.6`, even though the underlying numeric parameter remained exact.
- Fixture/reproducer: `tests/test_issue_1349_precision_display.py` covers `13.55`, `6.25`, `4.75`, `0.85`, and `1.875`, plus compound hole callouts, turned dimensions, ladders, live edits, and generated scripts.
- Before/after result: `sheet.dimension(feature, "parameter.id").format(decimals=n)` now preserves requested trailing places while `ApprovedDimension.value` and `DimensionId` remain authoritative.
- Mutation that makes the guard fail: removing the precision policy at the intent, planner/compiler, renderer, reservation, or script-emission boundary makes the corresponding end-to-end regression fail; the public acceptance regression fails on `main` because `DimensionIntent.format()` is absent.

## Contract and scope

- Smallest contract this change tests: a referential dimension may control printed decimal places without restating its nominal or changing placement.
- Relevant ADRs: 0011 (declared IR), 0014 (collect then solve), 0015 (compiler pipeline), and 0016 (declared dimensioning intent and compiler-owned label text).
- Explicitly deferred: arbitrary display strings and precision policy for compound `location` intent; automatic/detected formatting remains unchanged.

## Validation

- [x] `scripts/pr-check --quick`
- [x] `scripts/pr-check` before final review (`4916 passed, 5 skipped, 2 xfailed`; 95.13% total coverage; 94% diff coverage)
- [x] The named mutation was observed failing before the implementation passed it
- [x] Automatic, declared, and generated-script paths were checked where affected
- [x] Recognition and repeated-lint call counts were checked where affected (N/A for recognition: no recognizer code or behavior changed; repeated lint remains covered by the full suite)

## Stop check

- [x] New prerequisites were split or the work was explicitly reclassified as discovery (N/A: no new prerequisite or discovery work)

- [x] I have read the [CLA](https://github.com/pzfreo/draftwright/blob/main/CLA.md) and agree to it. If I am contributing on behalf of an organisation, I am authorised to agree on its behalf.

Closes #1349

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ### Added
 
+- Referential Sheet dimensions can preserve an explicit number of decimal places with
+  `sheet.dimension(feature, "parameter.id").format(decimals=n)`. The policy changes only
+  printed display text: numeric reconciliation, tolerances, suppression and dimension
+  provenance continue to use the feature parameter, and generated Sheet scripts round-trip the
+  policy (#1349).
 - New lint code `annotation_ink_overlap`: reports another annotation's line-work —
   a dimension line, an extension line or a leader shaft — drawn through a label's
   text, which `annotation_overlap` cannot see because it compares label boxes to
--- a/docs/reference/sheet.md
+++ b/docs/reference/sheet.md
@@ -46,6 +46,23 @@ resolved result.
     options:
       filters: public
 
+## Dimension intent handle
+
+`dimension(feature, parameter_id)` and `add_dimension(feature, parameter_id)` return a
+referential `DimensionIntent`. The handle never carries a replacement nominal and never chooses
+page coordinates. Use `format(decimals=n)` to preserve between 0 and 15 decimal places in the
+printed nominal while reconciliation, tolerance, suppression and provenance continue to read the
+numeric parameter from the feature. Trailing zeroes are intentional manufacturing display text:
+
+```python
+sheet.authored_dimensions()
+sheet.dimension(envelope, "width.length").format(decimals=2)  # 13.5 prints as 13.50
+```
+
+::: draftwright.sheet.DimensionIntent
+    options:
+      filters: public
+
 ## Hole handle
 
 ::: draftwright.sheet._Hole
--- a/src/draftwright/_geometry.py
+++ b/src/draftwright/_geometry.py
@@ -337,11 +337,18 @@ def _axis_direction_is_aligned(axis: str, direction, *, tol: float = 1e-3) -> bo
     )
 
 
-def _fmt(v: float) -> str:
+def _fmt(v: float, decimals: int | None = None) -> str:
     """Format a float as integer string if whole, otherwise 1 dp. The one number
     formatter the IR (:mod:`draftwright.model.ir`) and the drawing layers
     (:mod:`draftwright._core`) share (#700 — the two copies had already begun to
-    drift on ``-0``)."""
+    drift on ``-0``).
+
+    ``decimals`` is an explicit referential-dimension display policy (#1349). It preserves
+    trailing zeroes by design; ``None`` retains the established automatic formatting.
+    """
+    if decimals is not None:
+        normalized = 0.0 if round(float(v), decimals) == 0 else float(v)
+        return f"{normalized:.{decimals}f}"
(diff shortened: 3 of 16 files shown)


pzfreo:
ADR merge-gate review: PASS

- ADR 0011: the public declaration remains referential. `.format(decimals=n)` carries display policy only; the nominal is still read from the declared/detected feature parameter, and no recognizer or alternate producer path was added.
- ADR 0014: no annotation coordinates, strip choice, or placement mechanism were added. The existing solve still owns placement, and compound-callout reservation now measures the same precision-aware text the renderer emits.
- ADR 0015: the policy crosses the existing IR waist (`RequestedDimension` → planner → compiled plan/render intent). Numeric geometry stays numeric; renderer and estimator consumers receive the approved display policy through the established paths.
- ADR 0016: `DimensionIntent.format()` preserves the reference-not-restate contract, semantic identity, suppression, tolerance, and numeric reconciliation. General renderers consume compiler `value_text`; the already-documented legacy compound-hole seam remains inventoried by the label-provenance ratchet and is not widened. Generated scripts mirror the intent with chained `.format(...)`, independent of placement outcome.

Fail-closed checks cover conflicting policies, location-intent ambiguity, identity after feature reorder, invalid precision, and missing renderer/reservation propagation. No ADR deviation or new ADR is required. No recognizer change is present, so no upstream issue is needed.

pzfreo:
Final ADR merge-gate refresh after review/coverage fixes:

- ADR 0011: PASS — the public surface remains declarative and feature-referential; the added grid round-trip verifies canonical discriminated identities.
- ADR 0014: PASS — no raw coordinates or placement authority were added; Y-chain placement still flows through the shared solve/detail request path.
- ADR 0015: PASS — values and display policy still cross the compiled-plan waist; no alternate engine or renderer-owned measurement was introduced.
- ADR 0016: PASS — the final cleanup removes an unreachable duplicate renderer-side repeat-label decision, reducing rather than expanding the formatting seam. Numeric values and DimensionId remain unchanged.

Independent review after the final patch reports no blockers, majors, or minors (only one optional test-coupling nit). Full local gate: 4,918 passed, 5 skipped, 2 xfailed; 95.21% project coverage; 100% changed-line coverage. No recognizer code changed.

Issue #1274, opened 2026-08-22
Lint misses dimension-text collisions while reporting overlaps for empty labels


pzfreo:
Draftwright 0.4.9. Two symptoms of the same area — the overlap checks — found while comparing two drawings of the same part (GRM-03 thumbwheel drive screw, 28.7 × 10 × 10 mm, 16 faces).

## 1. Colliding dimension text is not detected

A sheet rendered at A4 2:1 scored **0.95 with a single warning** (`axial_length_missing`). No overlap warning of any kind.

The rendered PNG shows, in the lower axial chain, the dimension texts **"3.2" and "3" printed on top of one another** — the digits overlap directly, and the value cannot be read. Separately, five diameter leaders (⌀4, ⌀6, ⌀10, ⌀5, ⌀3) converge into a crossing bundle at the collar, where ⌀6, ⌀10 and ⌀5 cannot be told apart.

Both are exactly the kind of defect a reader notices immediately, and lint reported the sheet as essentially clean.

The practical consequence: an automated caller that trusts `lint_summary` as its quality signal treats this sheet as acceptable. There appears to be no check for **dimension text against other dimension text**, only `label_centerline_overlap`.

## 2. `label_centerline_overlap` reports empty labels

On a second drawing of the same part, Draftwright reported **six `label_centerline_overlap` warnings, every one naming an empty label** (`label ''`). Inspection of both the native PNG and a PNG rendered back from the PDF found no corresponding visible text at those locations.

So the two failures point opposite ways: the check that exists fires on labels that are not there, and no check fires on text that visibly collides.

## Why it matters beyond aesthetics

`lint_summary` is increasingly consumed programmatically rather than read by a person. A score that is high because a collision class is unchecked is worse than a lower score, because it is actively trusted. Empty-label warnings push the other way — they train callers to dismiss overlap warnings as noise, which is precisely the category that should be believed.

## Suggested direction

- Add an overlap check between placed dimension/annotation text bounding boxes, not only text against centrelines.
- Either suppress `label_centerline_overlap` for empty label text or fix whatever produces the empty entries, so the warning that survives means something.

Happy to supply the two Sheet scripts and rendered PNGs if useful.

pzfreo:
Second instance, on a drawing that is otherwise sound — which makes it cleaner evidence than the first, since nothing else about the sheet is wrong.

[...]

Issue #1257, opened 2026-08-21
Lint passes a visible isometric/orthographic view overlap on GRM-03 (false negative)


pzfreo:
## Summary

A fresh Draftwright 0.4.9 GRM-03 review produced both kinds of critique disagreement:

1. **false negative:** a visible isometric/orthographic view overlap passed lint;
2. **false positive:** a clean A4 drawing retained six `label_centerline_overlap` warnings for `label ''`.

Grouping these because the common contract is: lint should describe the rendered sheet a reviewer sees.

## Case A — visible overlap not reported

An A3 5:1 strict candidate contained the GRM-03 orthographic views, isometric view, diameter/axial dimensions, consolidated short-feature notes, and an overall reference dimension.

Build result:

```
scale=5.0
page=420 x 297
errors=0
geometry_issues=0
placement_drops=0
by_code={"label_centerline_overlap": 6}
```

Visual inspection of the exported PNG showed the isometric wheel/body intruding over the upper orthographic shaft. No `view_overlap` issue was emitted.

The candidate was rejected manually despite passing the available layout critique.

## Case B — empty-label false positives remain

The selected A4 2:1 drawing has no visible text/centreline collision. Native PNG inspection and a render back from the one-page PDF agree.

Lint still returns six warnings such as:

```
label_centerline_overlap:
label '' overlaps centerline by 7.5×75.0 mm
label '' overlaps centerline by 119.9×5.2 mm
```

This resembles #1142, which is closed, but still reproduces on the current GRM-03 rotational furniture path in 0.4.9.

## Impact

- A layout that is visibly unsuitable can pass.
- A visibly clean layout produces repeated warnings.
- Automated candidate selection cannot trust the critique in either direction.

## Acceptance

- Stroke/geometry bounds of every placed view, including isometric versus orthographic, participate in one final overlap check after repack.
- The GRM-03 A3 5:1 candidate reports the real view overlap or is repacked so it does not occur.
- Empty semantic labels never enter text-label overlap lint; use the actual rendered text object/bounds where text exists.
- The final GRM-03 A4 2:1 drawing produces no empty-label centreline warnings.
- Add raster/SVG-backed regression evidence so view-block metadata alone cannot certify a visibly overlapping sheet.
- Preserve #1216's cross-scale pairwise checks without reintroducing duplicate-pair score inflation.

Related: #1130, #1216, #1147, #1142.

pzfreo:
[...]

PR #1451, opened 2026-09-03
fix: reconcile same-face boss and blend ownership

### LINKED ISSUE #1450: GRM-04: Ø8 callout points at the web and duplicates the R4 lower profile
## Observed

GRM-04 produces both `Ø8` and `R4` for the same 8 mm lower circular profile, while the `Ø8` leader does not land on that profile. It terminates in the tapered web/blank interior, so the reader cannot tell what cylindrical feature it is meant to identify.

The `R4` leader correctly identifies the lower outside arc. Since `Ø8 == 2 × R4`, the sheet presents two different semantic declarations for the same physical circular boundary and makes the incorrect `Ø8` callout look like a separate feature.

This reproduces from the repository fixture on current `main` at `6b67498` and with released Draftwright 0.4.17:

```bash
uv run draftwright tests/fixtures/grm04_drive_plate.step \
  --out /tmp/grm04 --title 'GRM-04 DRIVE PLATE' --number GRM-04 \
  --page A3 --scale 5 --scale-policy strict --zones \
  --projection third --script
cd /tmp
uv run --project /path/to/draftwright python grm04.py
```

The emitted declarative source contains both owners:

```python
boss1 = sheet.diameter(
    diameter=8,
    height=4.5,
    span=((-1.5, 0, 0), (3, 0, 0)),
    at=(3, 0, 0),
    axis="x",
)
blend2 = sheet.blend(
    axis="x",
    radius=4,
    at=(0.287, 0, -4),
    side="convex",
    axis_direction=(1, 0, 0),
)

sheet.dimension(boss1, "boss.diameter")
sheet.dimension(blend2, "blend.radius")
```

On the rendered front view:

- `R4` points to the lower outside arc;
- `Ø8` points diagonally into the tapered web rather than touching the 8 mm circular boundary; and
- both values describe the same lower-profile size.

`drawing.lint()` reports centreline warnings and the unrelated `16.5`/`Ø2.4 THRU` ink crossing, but reports neither this duplicate physical measurement nor the detached/mistargeted `Ø8` leader.

## Expected

1. One physical circular fact should have one semantic owner. Emit either the appropriate `Ø8` diameter or `R4` radius for this profile, not both.
2. If `Ø8` is retained because a genuine cylindrical boss exists, its leader must terminate on the projected boundary of that boss rather than the tapered web.
3. Completeness/dedup should reconcile diameter and blend candidates that resolve to the same physical circular geometry.
4. Lint should fail closed when a retained callout tip does not touch the feature it names.

## Why this matters

This is worse than harmless over-dimensioning: the duplicate values suggest two distinct features, and the `Ø8` leader visually assigns one of them to the wrong area. A human or drawing agent can see the ambiguity, but the current coverage and lint signals treat the sheet as complete on this axis.

Related: #1378 covers the generic leader-tip-in-air failure. This issue adds the exact GRM-04 fixture and the cross-feature `diameter`/`blend` ownership conflict that makes the rendered result especially misleading.

pzfreo:
A provider-proved same-face boss/blend pair now has one drafting owner. On GRM-04, the drawing retains R4 and the 4.5 envelope while removing the detached duplicate Ø8 callout.

The join requires unique same-run face evidence, matching circular geometry, and full-envelope axial coverage. Detection and independent lint share the envelope predicate and diameter tolerance, including the same declared turned-profile input. Generated-script physical lint follows that ownership decision. The existing 0.005 mm envelope tolerance remains conservative: a pair outside it is not absorbed.

Closes #1450.

## Architectural fit

Ownership remains at the recognition-to-IR boundary and uses provider face evidence. The shared predicate lives below detection and lint, preserving their import boundary. Existing compiled measurements and solver placement remain authoritative; no raw annotation placement or additional recognition run is introduced. No ADR text changes are required.

## Validation

On head `8de08e5c975e94c07039020527130421bdc217b6`, updated with the merged development-version bump:

- `scripts/pr-check --full` exited 0.
- 6,747 passed, 5 skipped, 2 expected failures.
- Total coverage 95.94%; changed-line coverage 97% (77 changed lines, 2 uncovered).
- Ruff, formatting, mypy, spelling and strict documentation build passed through the same gate.
- Earlier rendered GRM-04 evidence and mutation results are recorded in this PR's review history.

Hosted checks for this head must pass before merge. The local gate excludes the slow tier, which remains the existing post-merge main safety net.

### DIFF
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -3718,6 +3718,12 @@ class Drawing:
                 assembly=self.assembly,
                 holes=holes,
                 bosses=bosses,
+                blends=recognition.blends,
+                recognition_evidence=self._build.recognition_evidence,
+                # The same profile set `detect` used, so the boss/blend absorption decision
+                # is one decision rather than two that can disagree. `prof_kw` is empty only
+                # where detect also falls back to the aggregate.
+                **({"turned_profiles": prof_kw["profiles"]} if "profiles" in prof_kw else {}),
                 registry=self._registry,
             )
             issues += lint_axial_coverage(
--- a/src/draftwright/linting/coverage.py
+++ b/src/draftwright/linting/coverage.py
@@ -58,6 +58,11 @@ from draftwright.recognition_frame import (
     groove_owns_turned_step_band,
     profiles_owning_axial_band,
 )
+from draftwright.recognition_ownership import (
+    BOSS_BLEND_DIAMETER_TOL,
+    boss_blend_owner_pairs,
+    envelope_is_emittable,
+)
 from draftwright.view_plan import VIEW_AXES
 
 _UNSET = object()  # sentinel: distinguishes "not supplied" from a valid prof=None
@@ -269,6 +274,9 @@ def lint_feature_coverage(
     assembly=None,
     holes=None,
     bosses=None,
+    blends=None,
+    recognition_evidence=None,
+    turned_profiles=_UNSET,
     registry=None,
 ) -> list:
     """Coarse completeness check: report part diameters with no callout (#80).
@@ -303,6 +311,11 @@ def lint_feature_coverage(
     ``cyls`` accepts a precomputed ``analyse_cylinders(part)`` result so
     repeated lint runs need not re-scan the solid.
 
+    When same-run ``recognition_evidence`` proves that a recognised boss and blend share the
+    same defining face and circular value, the boss diameter is excluded from this coarse
+    physical inventory. The blend radius is the truthful callout for that face; counting its
+    equivalent diameter again would recreate the duplicate this lint is meant to detect.
+
     Counts are checked too (#92): the part's holes (via ``recognise_holes``) give
     a required count per diameter (each bore, counterbore, and spotface
     occurrence counts one), and structured callouts declare how many holes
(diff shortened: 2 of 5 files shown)


pzfreo:
## Review after rebase onto `main` (cd4795f1)

Rebased and pushed as `b2022215`. It needed four repairs just to run, and I found two design issues I did **not** fix because they are your call.

### Design issues — not fixed

**1. `envelope_emittable` is now computed twice, by two different formulas, from two different profile sets.**

- `model/detect.py:1834` — `not profiles and not _is_round(bbox, bosses_d) and not polygonal_stock`
- `linting/coverage.py:339` — `not turned_profiles and not polygonal_stock and not any(abs(boss.diameter - bbox.size.X) <= 0.5 and ...)`

The second inlines `_is_round`'s body (`detect.py:2699`) with its `0.5` hardcoded rather than calling it, and — the part that matters — reads `recognition_evidence.result.turned_profiles`, while detect's `profiles` is **not always** the recognition inventory. At `detect.py:1793-1795`, when a caller supplies `prof`/`profiles` explicitly, `profiles` is the caller's value:

```python
if prof is _UNSET and profiles is _UNSET:
    profiles = recognition.turned_profiles
...
if profiles is _UNSET:
    profiles = () if prof is None else (prof,)
```

So on a declared build the two can disagree. If detect says emittable and lint does not, the IR drops the boss diameter while lint still requires it — `feature_not_dimensioned` against a diameter the plan deliberately removed. The converse hides a real duplicate. That is the render/lint disagreement class already tracked in #1126, #1274 and #1257, and this change introduces a second site where it can arise.

**I did not reproduce it** — it needs a declared turned profile alongside a boss/blend pair, which I could not construct quickly. The mechanism is what I am reporting, not an observed failure. One shared helper taking the profile set as an argument would remove the question.

**2. `_BOSS_ENVELOPE_SPAN_TOL = 0.005` (`recognition_ownership.py:22`) is 5 µm** against imported STEP coordinates. A boss that spans the envelope to within 6 µm silently stops being absorbed and the sheet changes back, with no diagnostic saying why. `_DIA_TOL` next door is 0.15. Deliberate, or should it be looser?

**3. Two independent `0.15` literals** must stay equal or lint and detect disagree about whether a boss and blend state the same circle: `_DIA_TOL` (`detect.py:262`) and `lint_feature_coverage(tol=0.15)` (`coverage.py:267`). They match today; nothing binds them.

### Fixed during the rebase

**4. `Blend.axis` no longer exists.** recognisers 0.4.14 (#1459, shipped in v0.4.18) moved the direction into `StraightBlendPath`/`CircularBlendPath`. Three of the four tests failed with `AttributeError`. Now reads `blend_provider_key(blend)`, which is the one fail-closed reader of that released shape — re-deriving the axis here would be a second unvalidated copy of the provider boundary.

**5. mypy: 7 errors.** `boss_spans_envelope(boss: object, bbox: object)` reads both by attribute, which `object` cannot express. Now `Any`, matching sibling `recognition_frame`. The PR's stated verification ran `ruff` and a test subset, so the static gate had never run against this branch.

**6. `part.bounding_box()` on every lint of every part.** `bosses`/`blends` are inventories, so an empty one is not `None` — the `is not None` test measured the solid again even with zero bosses, breaking `test_relint_recomputes_no_annotation_boxes`, which budgets exactly one solid measurement on the re-lint path. Verified failing here and passing on clean `main`. Changed to truthiness.

**7. Diff coverage was 88% against a 93% gate.** The uncovered lines were the fail-closed arms this PR's own description advertises — "an evidence-less build, an ambiguous face claim ... fails closed" — so nothing established that a missing join refuses rather than absorbs, which is the whole safety property. Added tests for both, and for the empty-inventory skip. Now 97%.

**8. Dropped "Amendment 21" to ADR 0017.** That record was archived by #1463 and is stamped frozen history. Per CLAUDE.md, recording ownership for one more record type belongs in the PR body, not an ADR. The text is preserved in the commit; nothing else about the change is affected.

### State

`scripts/pr-check --full` exit **0** — 6745 passed, 5 skipped, 2 xfailed, diff coverage 97%. Still a draft; items 1–3 are yours to decide.

pzfreo:
## Finding 1 fixed: one shared decision

Pushed. The two `envelope_emittable` computations are now one predicate with one input.

**`recognition_ownership.py`** (rank 0, already imported by both sides) gains:

- `envelope_is_emittable(*, bbox, bosses, turned_profiles, polygonal_stock)`
- `boss_fills_footprint(...)` — `detect._is_round` now delegates to it, so its `0.5` has one home and `detect.py:2400`'s other caller stays consistent
- `BOSS_BLEND_DIAMETER_TOL = 0.15` — replaces the two independent `0.15` literals (finding 3). `lint_feature_coverage`'s `tol=` no longer reaches this decision, which is the point: detect never saw the caller's `tol`.

**The input is unified too**, which is the half that mattered. `lint_feature_coverage` takes `turned_profiles`, and `drawing.py` passes the same set it gave detect (`prof_kw["profiles"]`), falling back to the aggregate only where detect also falls back. A declared profile now reaches both sides.

### On the guard

My first attempt asserted that the shared names *appear* in both modules. Under mutation that was worthless: making lint re-derive the profile set and ignore its argument — the original defect, restored — left it green. Replaced with a behavioural test that passes a declared profile and asserts the answer changes, plus the direction (declaring a profile stops the absorption, so the boss diameter returns to the inventory).

Mutation results now: lint re-deriving the set → **caught**; the shared predicate ignoring `turned_profiles` → **caught**; lint re-inlining the roundness formula → **caught**; the tolerances drifting apart → **caught**. Dropping `polygonal_stock` from the predicate is inert on GRM-04 and is caught elsewhere, by `test_issue_1382_through_step_semantics::test_complement_owner_requires_an_emitted_envelope[polygonal-stock]` — verified by mutation rather than assumed.

### Still yours

**Finding 2** — `_BOSS_ENVELOPE_SPAN_TOL = 0.005` (5 µm) is untouched. Changing a tolerance changes which parts absorb, and that is a drafting judgement, not a refactor.

`scripts/pr-check --full` exit **0** — 6747 passed, 5 skipped, 2 xfailed, diff coverage 97%.

Issue #1571, opened 2026-09-10
label_vs_measured fires on a correctly rounded label (⌀118 boss height, 4.84 shown as 4.8)


pzfreo:
Found while verifying #1132 on CADGenBench 109, a part that until today could not produce a
drawing at all — so this had no way to surface before.

draftwright at `d5b91b6` + #1132's branch, quiddity 0.2.8.

## What is reported

```
[label_vs_measured] Dim '4.8': label value 4.800 differs from measured path length 2.420
                    (÷0.5 = 4.840) by 0.8% — possible axis swap or wrong endpoint
```

`label_vs_measured` is a **fidelity** code — `_FIDELITY_CODES` in `linting/quality.py` — meaning
"the drawing asserts something false". It is one of the few codes that lowers the honest quality
component rather than the legacy score alone, so a false one is expensive.

## Why I think it is a false positive

The annotation is `m_bossheight_z2`, and its claim resolves cleanly:

```
claims: [{'feature': 'boss@(0.000,0.000,347.966)/z[diameter=118.000]',
          'parameter_id': 'boss_height.length'}]
```

So the label is bound to a real boss-height measurement — not a swapped axis or a stray
endpoint, which is what the message proposes. The drawn path measures 4.840 mm in world units
and the label reads `4.8`, which is what 4.840 becomes at one-decimal display precision. The
0.8% "discrepancy" is exactly the rounding.

If that reading is right, the check is comparing the **displayed** label against the **exact**
path length, so any value whose last displayed digit rounds will trip it once the ratio exceeds
the tolerance. Small dimensions are the exposed case: at 1 dp, 4.84 → 4.8 is 0.83%, while the
same absolute rounding on a 48.4 mm dimension is 0.08% and passes.

## What I have not established

I have not confirmed the boss's true height is 4.84 from the geometry, only that the rendered
path measures it. Nor have I found the check's tolerance or confirmed it has no
display-precision allowance — I stopped at the point where this stopped being #1132's business.
So treat the diagnosis as a hypothesis with the measurements attached, not a conclusion.

## Reproduction

```
generate_sheet_script(
    "<cadgenbench>/build123d-mcp-0.3.85.dev0-opus-5-xhigh-complete81-r1/109/output.step")
# sha256 301cb3ce914fa28c…, 1 solid, bbox 130.940 x 130.940 x 420.000
# then run the emitted script and read drawing.lint()
```

Filed separately rather than folded into #1132, which changes only where the turned-span check
[...]

Issue #1564, opened 2026-09-10
Mini-epic: make the hosted AI drawing loop start, see, and tell the truth


pzfreo:
Seven issues selected for their effect on the hosted AI drawing loop
(pzfreo/draftwright-io), which begins by executing a generated `Sheet` script and
steers on Draftwright's structured output. Each is tracked in its own issue; this
one only holds the set and its order.

### The agent cannot start

- [x] #1132
- [x] #1555
- [x] #1135

### The agent is told nothing, or something false

- [x] #1563
- [x] #1126

### The drawing states something the source never said

- [x] #1157
- [x] #1230

No work is scoped here. Sequence, evidence and acceptance stay in the sub-issues.

pzfreo:
## Outcome — all seven have one

Four resolved, three blocked with named blockers. No sub-issue was left without a verdict.

### Resolved

| issue | outcome | PR |
|---|---|---|
| #1157 | Fixed. `None` is the new public default; the title block states `UNSPECIFIED`, `""` stays the deliberate blank, an explicit string round-trips. Also dissolves #501 — no sentinel left to dedupe. | #1566 (green) |
| #1563 | Fixed. `Sheet(part, source=…, pmi=…)`; the census credits its document; two new checks (`pmi_unreconciled`, `pmi_source_unknown`). | #1567 |
| #1230 | Already fixed by #443 and #1225's provenance threading — verified by running the ratchet's own sweep (`escapes: []`, `unconfirmed: []`). The issue's diagnosis was also mistaken. One stale ledger comment corrected. | in #1567 |
| #1126 | Inversion gone: dropping declarations now lowers the score 0.80 → 0.45 → 0.35 → 0.00 where 0.4.2 measured 0.50 → 0.85 → 0.90. Test-only PR adding the guard nobody had written. | #1568 |

Every fix is mutation-checked, each mutation failing a *named* test. Two corrections went on the
record rather than being quietly fixed: a width rationale I invented for #1157 (measured: every
candidate fits), and a negative mutation result for #1126 (monotonicity alone passes even with
the mechanism removed, so the three assertions are load-bearing only together).

### Blocked — the recogniser chain

| issue | mechanism, measured | blocker |
|---|---|---|
[...]

Issue #1555, opened 2026-09-09
Plate edge blends recognised as turned steps: spurious ⌀50 and ⌀7 callouts on a part with no turned features


pzfreo:
Draftwright 0.4.27.dev0. A 3 mm sheet-metal bracket — a part with **no turned features
whatsoever** — produces two `TurnedStep` occurrences, and both reach the sheet as
diameter callouts.

## What the recogniser reports

```
turned_steps:1  TurnedStep represented  {'axis': 'z', 'diameter': 50.0, 'hi': -50.8,     'lo': -53.8}
turned_steps:2  TurnedStep represented  {'axis': 'z', 'diameter':  7.0, 'hi': -16.867,   'lo': -19.867}
```

Both are exactly **3.0 mm tall — the sheet thickness**. Neither is a turned diameter:

- **⌀50** is the R25 blend on the strap flank. A flank blend through a 3 mm plate is a
  cylindrical face of radius 25, and the turned-step recogniser accepts it.
- **⌀7** is the pair of R3.5 rounded corners on the hanger plate. These are *already*
  recognised as fillets — `fillets:8` at `[1.025, -24.085, -18.367] r3.5` and `fillets:9`
  at `[16.975, -24.085, -18.367] r3.5`, mid-plane `z = -18.367`, which is exactly the
  midpoint of the ⌀7 step's `hi`/`lo`. So one piece of geometry is stated twice, in two
  different vocabularies.

## What reaches the sheet

The compiled plan approves both as `[step]` groups in the front view:

```
  [step] view=front
      length     step     3 (3.000)
      diameter   step    50 (50.000)
  [step] view=front
      length     step     3 (3.000)
      diameter   step     7 (7.000)
```

On the rendered drawing:

- the **⌀50** callout sits at the bottom-left of the front elevation with **a leader
  arrow pointing into empty space** — there is no ⌀50 anywhere on the part, and nothing
  under the arrow;
- the **⌀7** callout duplicates geometry the plan view already states as `3× R3.5`;
- each contributes a `3` step-length dimension, correct only by coincidence (it is the
  sheet thickness).

An arrow pointing at nothing is the kind of defect a reader spots instantly, and lint
does not mention it.

## Why it matters

The failing predicate looks like it will accept *any* rounded outer corner or edge blend
on a plate: on sheet-metal and plate parts every one of them is a full cylindrical face
whose height is exactly the material thickness. That makes this a systematic
false-positive class on a whole family of parts, not a one-off.

[...]

Issue #1135, opened 2026-08-12
Sheet.build() does not terminate on a turned part with 9 step declarations (CADGenBench 109, >1h)


pzfreo:
## Current triage checkpoint — 2026-09-06

The reported reproduction used Draftwright 0.4.2. Retain this issue until the supplied CADGenBench source and emitted script are rerun against the current released version and the pending reviewed release candidate. Record the exact version, entry point, failure/timeout and fixture identity before deciding whether a code fix or closure is warranted; no current reproduction was established by the tracker audit.

<details><summary>Original report and reproduction evidence</summary>

## Summary

Building the emitted starter script for a threaded cylindrical connector does not complete
within **one hour**. Comparable parts build in 40–90 seconds.

```
emit  ok in 59s
build no completion in 3600s   (also tested at 600s)
```

Recognition and script emission are fine — 13 features in 59 s. It is `Sheet.build()` that
never returns. There is no error, no partial output, and no progress signal; the process
simply has to be killed.

## The part

CADGenBench fixture 109, "threaded cylindrical connector", core tier. Recognised inventory:

```
hole 1, step 9, diameter 1, rotational 1, envelope 1
```

## Likely related to #1132

That issue reports a *flanged tubular spool* whose emitted starter is rejected by
`build()`:

> `step() declares a segment of a turned profile, but the declared steps don't span this
> part's full height`

Its inventory is `pattern 2, hole 1, step 5, rotational 1, envelope 1`.

Both failures are turned/rotational parts carrying many `step` declarations alongside a
`rotational`. One fails closed with a clear message; this one does not terminate. They may
well be the same underlying problem in turned-profile handling — one path detecting the
inconsistency and one path not.

Of ten CADGenBench fixtures baselined, these are the only two that produced no drawing, and
both are from the six-part core tier.

## Why an unbounded build matters beyond this part

A build that fails is recoverable; a build that never returns is not. Anything holding a
worker for it — CI, a batch import, or the server-side drawing agent in draftwright-io#405,
which explicitly requires that waiting never holds CAD capacity and that idle
infrastructure can scale to zero — is stuck until something external kills it. A wall-clock
ceiling inside `build()`, or a solver iteration bound that fails closed with a diagnostic,
would turn this into the #1132 case: bad news, but news.

## Reproduction

- draftwright **0.4.2**
- CADGenBench fixture 109, submission
  `pzfreo_opus5-xhigh-official-baseline-prompt-nom_20260811-034944`
[...]

PR #1572, opened 2026-09-10
Generate and warn instead of refusing a partial turned profile (#1132)

### LINKED ISSUE #1132: Emitted starter script fails to build: step() chosen for features build() rejects (CADGenBench 132)
## Current triage checkpoint — 2026-09-06

The reported reproduction used Draftwright 0.4.2. Retain this issue until the supplied CADGenBench source and emitted script are rerun against the current released version and the pending reviewed release candidate. Record the exact version, entry point, failure/timeout and fixture identity before deciding whether a code fix or closure is warranted; no current reproduction was established by the tracker audit.

<details><summary>Original report and reproduction evidence</summary>

## Summary

For a flanged tubular spool, the starter script Draftwright emits **fails to build under
Draftwright**. `emit` &rarr; `Sheet.build()` should be self-consistent; here the emitter
chooses `step()` for a set of features that `build()` then rejects, and the error message
recommends the verb the emitter should have used.

```
ValueError: step() declares a segment of a turned profile, but the declared steps don't
span this part's full height — it is not a turned (rotational) body. For a boss (an
external cylinder on a prismatic part) use .boss() — it renders its own ø and height.
```

The user-facing effect is a dead end: the part imports, a script is generated, and there
is no drawing and no obvious way forward without hand-editing generated code.

## The emitted script

```python
sheet = Sheet(part, title='cgb-132', number='DWG-001')
pattern1 = sheet.pattern(hole(diameter=10, at=(0, 0, 12),  axis="z", depth=12),  kind="bolt_circle", count=6,  bcd=90)
pattern2 = sheet.pattern(hole(diameter=8,  at=(0, 0, 140), axis="z", depth=113), kind="bolt_circle", count=12, bcd=110)
hole1 = sheet.hole(diameter=62, at=(0, 0, 0), axis="z", spotface=(70, 10), depth=130)
step1 = sheet.step(diameter=20,  length=10,  at=(0, 0, 5))
step2 = sheet.step(diameter=20,  length=2,   at=(0, 0, 11))
step3 = sheet.step(diameter=130, length=5,   at=(0, 0, 29.5))
step4 = sheet.step(diameter=70,  length=103, at=(0, 0, 83.5))
step5 = sheet.step(diameter=130, length=5,   at=(0, 0, 137.5))
rotational1 = sheet.rotational(od=130, bores=(70, 62,), at=(0, 0, 70), axis="z")
envelope1 = sheet.envelope()   # envelope 130 × 140 × 130
```

The declared step lengths total 125 against a part height of 140, which is exactly the
condition `build()` rejects. Note also that `step1`/`step2` are ø20 features on a part with
od=130 — plausibly the bosses the error message is pointing at, rather than segments of the
turned profile.

## Why this is worth more than the individual part

`emit` producing a script that `build` refuses is a self-consistency failure, and it is the
one failure mode a consumer cannot route around: any automated pipeline — the drawing
agent in draftwright-io#405, a CI check, a batch import — starts from the emitted script.
If that script is not guaranteed to build, every such pipeline needs a repair path before
it can begin, and an agent's first action is fixing Draftwright's output rather than
improving it.

A property test over the fixture corpus — *for every part, `emit()` then `build()`
succeeds* — would catch this class directly.

## Context

Found while baselining ten CADGenBench fixture parts (paired reference drawing + STEP).
Eight of ten built. This is one of the two that did not, and both are from the six-part
"core" tier whose geometry is closest to the benchmark reference. The other, a threaded
cylindrical connector, exceeded a 600 s build timeout and is filed separately.

## Reproduction

- draftwright **0.4.2**
- CADGenBench fixture 132 (flanged tubular spool), submission
  `pzfreo_opus5-xhigh-official-baseline-prompt-nom_20260811-034944`
- `emit` the starter, then `Sheet.build()`

Happy to share the STEP file.

</details>

pzfreo:
Closes #1132. Part of #1564. **Stacks on #1570** — its evidence needs quiddity 0.2.8, so review
that one first; this PR's own diff is the last four commits.

## The defect

`generate_sheet_script` settles its layout through `builder._assemble`, so the z-turned span
check raising there meant a part whose recognised profile does not tile produced **no script and
no drawing**. An entry point that returns nothing is the one failure a consumer cannot route
around — every automated pipeline starts from the emitted script.

## The change

The check moves to `linting.coverage.lint_turned_profile_span`, beside
`lint_declaration_reconciliation` — the same declared-vs-geometry shape, computed from the model
at lint time. The predicate is unchanged in substance: same tiling walk, same groove handling,
same single-solid and orientation gates. Only its severity and its home moved.

**It could not stay in `_assemble`, and my first attempt to leave it there failed silently.**
Recording a `LintIssue` on the drawing's registry looked correct and disappeared: the annotate
pass calls `reset_issues()` so a repack cannot duplicate drop records, which discards anything
the builder recorded first. That version built the part and reported *nothing* — worse than
either raising or warning. I found it only by checking `by_code` rather than trusting that the
build succeeded.

## Result: all three CADGenBench fixtures produce drawings

None of them did this morning.

| fixture | before today | now |
|---|---|---|
| **114** bracket | `EMIT FAILED` in 1.6 s | builds clean (false steps fixed upstream, #1555) |
| **132** spool | `EMIT FAILED` in 4.8 s | **builds and warns** — 3 real steps cover `z 0..113` of 140 |
| **109** connector | >3600 s hang, then `EMIT FAILED` | **emits in 42 s, builds in 93 s, warns** — 7 real steps leave 23.69 mm (`z 343.13..366.82`) undescribed |

109's uncovered stretch is the threaded region, correctly no longer a step chain after
quiddity#587. Both warnings are true statements about the parts.

## Corrected after independent review

An adversarial review of this PR found three claims that failed on execution. All three are
fixed in the last commit; recording them here rather than silently amending the text above.

1. **The reported gap overstated by 3×.** A single forward walk stopped at the first gap and
   named everything above it — `343.13..420` on fixture 109, where a declared step covers
   `366.82..420`. The real shortfall is **23.69 mm**. The spans are now merged and every
   uncovered interval is reported with its total. I had repeated the wrong figure on #1132 and
   #1135 as well; both are corrected.
2. **The docstring gave the wrong reason warning is sufficient.** The real one is stronger:
   `_compile_overall_height` retains the overall-height rung whenever
   `_step_chain_covers_extent(...)` is false — the same tiling question, asked independently of
   the approved plan. So #631's dropped height *cannot co-occur* with this warning. It was
   already prevented elsewhere, and raising was belt and braces.
3. **The `_model_declared` gate's stated reason was false.** I wrote that the detection path
   "cannot under-declare against" its own profile. Measured on fixture 132, a plain
   `build_drawing(<step file>)` detects z-steps covering `0..113` of a 140 mm body and reports
   nothing, while this predicate on that same model warns. The gate stays — it is where the
   raise it replaces lived, so this alters no automatic drawing — but the comment now says what
   is true, and widening it is named as a separate question.

The review's mutation pass also found two unguarded properties: dropping the
`_GEOMETRY_AWARE_CODES` entry **and** downgrading severity to `info` together left the full fast
tier green. `tests/test_issue_1132_turned_profile_span_is_reported.py` pins both, plus the gap
arithmetic and a malformed-feature case. Four mutations, each failing a named test — including
one that restores the original first-gap-only walk.

## #631's contract changes, deliberately

A hand-authored `.step()` misuse now warns rather than raising. There is no third option: a
generated script is indistinguishable from hand-written code when re-run, so it is raise for
both or report for both, and raising for both is what produced #1132.

#631's substance survives, because its complaint was **silence** — reaching for `.step()` on a
boss flipped the height ladder into turned-suppression and dropped the overall height with no
signal. The misuse is still detected and still named; it is now a warning instead of an
exception. Seven tests encoded the old behaviour (three for #631, four in the plural-profile
suite). Each now asserts the warning, with the reason recorded at the site rather than the test
being quietly renamed to match.

Maintainer sign-off obtained for this loosening before implementing.

## Registration

`turned_profile_not_spanned` is in `_GEOMETRY_AWARE_CODES` — the shortfall is about the part,
not about where an annotation landed — and classified **unscored** in `quality.py` beside the
other `*_missing` codes, because an omission is not a false statement. The `_UNSCORED_CODES`
audit passes.

## Evidence

- Fast tier: **8450 passed**, 5 skipped, 13 xfailed.
- Slow tier: **62 passed, 2 xfailed** (`-n auto --dist worksteal`).
- `ruff format --check` and `ruff check` exit 0 over `src/` and `tests/`.

## Found while verifying, filed not folded

#1571 — `label_vs_measured` fires on a correctly rounded label on fixture 109 (a ⌀118 boss
height of 4.84 displayed as `4.8`). That is a fidelity code, so a false one costs the honest
quality component. It is unrelated to this change and has its own issue.

https://claude.ai/code/session_01N3VevbCf1UWuPzj3G927ef

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,12 @@
 
 ### Changed
 
+- Updated the exact Quiddity runtime and declared-inspection contract to **0.2.8**, which stops
+  a modelled thread being read as a stack of turned shoulders and coalesces contiguous
+  equal-diameter bands into the one span they physically are. Measured on the repo corpus,
+  NIST CTC-05 is again the only fixture that moves: its ⌀304.8 region was three adjacent
+  rungs and is now one, in both STEP encodings (#1135, quiddity#587).
+
 - Updated the exact Quiddity runtime and declared-inspection contract to **0.2.7**. Unlike the
   0.2.5 adoption below, this one changes drawings: the provider now requires circumferential
   cylinder support to establish a turned-profile axis line, so rounded plate corners and parallel
--- a/docs/reference/recogniser-capabilities.md
+++ b/docs/reference/recogniser-capabilities.md
@@ -70,9 +70,9 @@ issue. The contract test derives package record outputs, converter registries, l
 and emitter branches independently, so copying a new name into the declaration alone cannot make CI
 pass.
 
-### Quiddity 0.2.7 runtime contract
+### Quiddity 0.2.8 runtime contract
 
-The runtime pins the published `quiddity==0.2.7` package, following the 0.2.4 adoption in
+The runtime pins the published `quiddity==0.2.8` package, following the 0.2.4 adoption in
 [#1486](https://github.com/pzfreo/draftwright/pull/1486) and the migration in
 [#1471](https://github.com/pzfreo/draftwright/issues/1471). 0.2.7 is a patch adoption: it requires
 circumferential cylinder support to establish a turned-profile axis line, so rounded plate corners
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
     # published wheels, not only uv's development resolver constraints.
     "cadquery-ocp-novtk!=7.9.3.1.1 ; python_full_version >= '3.13'",
     "build123d-drafting-helpers>=0.15.1",
-    "quiddity==0.2.7",
+    "quiddity==0.2.8",
     "numpy>=1.24",
     # PNG raster export (SVG→PDF→PNG). Both permissively licensed (Pillow HPND; pypdfium2
     # Apache-2.0 wrapping Google PDFium BSD-3) and ship pre-built wheels with NO native system
--- a/src/draftwright/builder.py
+++ b/src/draftwright/builder.py
@@ -75,7 +75,6 @@ from draftwright.linting.coverage import lint_axial_coverage
 from draftwright.model import (
     Datum,
     Feature,
-    GrooveFeature,
     PartModel,
     StepFeature,
     build_pmi_features,
@@ -565,62 +564,13 @@ def _assemble(
         # measurements actually reach the page is settled downstream and reported by lint
         # (`axial_length_missing`). Guard on the tiling condition rather than a classifier
         # proxy (is_rotational / prof both have blind spots).
-        z_steps = [f for f in pm.features if isinstance(f, StepFeature) and f.frame.axis == "z"]
-        # The global-envelope check is meaningful only in its legacy single-solid domain;
-        # caller-owned group strings cannot turn one solid into several physical bodies.
-        # Multiple solids can be parallel, axially disjoint, or accompanied by unrelated
-        # compound members, so their steps need not tile the compound bbox (#1357).
-        owns_global_envelope = any(feature.kind == "envelope" for feature in pm.features)
-        if (
-            z_steps
-            and len(a.part.solids()) == 1
-            and (pm.orientation == "z" or owns_global_envelope)
-        ):
-            tol = 1e-3 * max(a.z_size, 1.0)  # small absolute float epsilon, floored
-            # Tiling means the segments run end to end — a single reach to each end isn't
-            # enough, since an interior gap is a stretch of part no declared step describes.
-            # Walk the spans low→high, extending coverage.
-            # A GROOVE counts as one of those segments (#953): it is machined INTO the turned
-            # profile, so a shaft that has one is still a turned profile, which is the only
-            # thing this guard is asking. Without it the emitted script for any grooved shaft
-            # RAISED — the detector produces exactly this model (step, groove, step) and the
-            # direct build draws it, so the engine rejected its own detector's output the
-            # moment that output was declared rather than detected. Grooves are the only
-            # detected feature that splits the chain: a chamfered or filleted shoulder leaves
-            # the two step spans meeting (checked), and a bore or cross-hole carves no axial
-            # interval. #631's own repro still raises.
-            # What this does NOT do is verify the declaration against the solid, and it never
-            # did: `.step(diameter=20, length=48, at=…)` fabricating one full-extent segment
-            # already defeats it on its own, groove or no groove (checked). Declared spans are
-            # the author's statement of intent, which ADR 4 (was 0011) takes at face value rather than
-            # re-deriving, so this catches the honest mistake #631 reported — declaring the
-            # boss you can see and getting a worse drawing — not a fabricated coverage claim.
-            # Hardening it into a real verb-domain check is #956.
-            # NOT claimed here: that the resulting drawing dimensions the height. On this very
-            # fixture the step chain drops at placement (crowded shoulders) and the height,
-            # suppressed at compile time on the premise the chain conveys it, is then conveyed
-            # by nothing — identically on the detected path. That is a real defect, tracked
-            # separately (#955); raising here would not fix it, only hide it from one of the
-            # two front doors.
-            spans = sorted(
-                [(min(p0[2], p1[2]), max(p0[2], p1[2])) for f in z_steps for (p0, p1) in [f.span]]
-                + [
-                    (f.frame.origin[2] - f.width / 2, f.frame.origin[2] + f.width / 2)
-                    for f in pm.features
-                    if isinstance(f, GrooveFeature) and f.frame.axis == "z"
-                ]
-            )
-            covered = a.bb.min.Z
-            for lo, hi in spans:
-                if lo <= covered + tol:
-                    covered = max(covered, hi)
-            if spans[0][0] > a.bb.min.Z + tol or covered < a.bb.max.Z - tol:
-                raise ValueError(
-                    "step() declares a segment of a turned profile, but the declared steps "
-                    "don't span this part's full height — it is not a turned (rotational) "
-                    "body. For a boss (an external cylinder on a prismatic part) use .boss() "
-                    "— it renders its own ø and height."
-                )
+        # The z-turned span check that used to RAISE here now reports from
+        # `linting.coverage.lint_turned_profile_span` instead (#1132). Raising made
+        # `generate_sheet_script` return nothing at all for a part whose recognised
+        # profile does not tile, and an entry point that produces no script and no
+        # drawing is the one failure a consumer cannot route around. The check itself is
+        # unchanged in substance; only its severity and its home moved, and the annotate
+        # pass's `reset_issues()` is why it cannot be recorded from here.
         # PMI (STEP AP242) is likewise detection-sourced, so a declared / emitted-script model
         # carries none. When PMI annotation is on, synthesise the same imported drafting
         # annotations detection would (render_pmi reads them off the model, gated on a.pmi_mode)
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -4474,7 +4474,14 @@ class Drawing:
                 # Declared-vs-geometry in the axial direction (#1132): a z-turned profile that
                 # leaves part of the body undescribed. Gated with the reconciliation check
                 # above because both are only meaningful for a caller-declared model — the
-                # detection path builds its own profile and cannot under-declare against it.
+                # detection path is NOT immune — it is scoped this way for a different reason.
+                #
+                # Measured on CADGenBench 132: a plain `build_drawing(<step file>)` detects
+                # z-steps covering 0..113 of a 140 mm body and reports nothing here, while this
+                # predicate applied to that same detected model returns the warning. Same part,
+                # same shortfall, one door silent. It is scoped to the declared model because
+                # that is where the raise it replaces lived, so this change alters no automatic
+                # drawing. Widening it is a real question and a separate one.
                 issues += lint_turned_profile_span(
                     features,
                     (self._analysis.bb.min.Z, self._analysis.bb.max.Z),
--- a/src/draftwright/drawing.py
+++ b/src/draftwright/drawing.py
@@ -138,6 +138,7 @@ from draftwright.linting import (
     lint_round_bottom_blind_slot_coverage,
     lint_slot_coverage,
     lint_through_step_coverage,
+    lint_turned_profile_span,
     pmi_stage_summary,
 )
 from draftwright.linting.angular import lint_angular_supports, lint_profile_angle_coverage
@@ -227,6 +228,10 @@ _GEOMETRY_AWARE_CODES = frozenset(
         "authored_dim_degenerate",
         "authored_dim_source_unresolved",
         "axial_length_missing",
+        # A turned profile that leaves part of the body undescribed (#1132). Registered
+        # here for the same reason as `axial_length_missing` beside it: the shortfall is
+        # about the part, not about where an annotation landed.
+        "turned_profile_not_spanned",
         "flat_requirement_suppressed",
         "flat_requirement_missing",
         "flat_requirement_unverifiable",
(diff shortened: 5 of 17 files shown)


pzfreo:
## Adversarial review — #1572 (commits after `d5b91b6` only)

**Verdict: land with changes.** The core move is right and I could not find a case where warning
instead of raising costs the drawing anything — but I got that answer for a reason the PR does not
give, and three of the PR's own factual claims are wrong when executed. Two of the corrections are
one-liners; one is a design question.

Everything below was run. Fixture work used a local CADGenBench copy (see *Reproduction* at the end),
since those parts are not in this repo.

---

### 1. The stated reason for the `_model_declared` gate is false on the PR's own headline fixture

`drawing.py` gates the new call with:

> *"the detection path builds its own profile and cannot under-declare against it."*

Measured, on CADGenBench **132** (`z 0..140`, 1 solid) through `build_drawing(<132>.step)` — the
plain detection path, no model:

```
model_declared: False   orientation: z   solids: 1
detected z-step spans: [(0.0, 5.0), (5.0, 108.0), (108.0, 113.0)]
DETECTED-BUILD lint by_code: {'hole_requirement_missing': 4, 'diameter_leader_target_unverifiable': 2}
passed: True   geometry_issues: 4
lint_turned_profile_span(<the detected model>) -> [('warning', 'turned_profile_not_spanned')]
   "...leave 113..140 of this part's height undescribed"
```

The detection path *does* under-declare against its own part — by exactly the 27 mm the PR's table
cites — and the predicate fires when handed the detected model. Same on **109**: detected steps leave
`343.13..` uncovered, `by_code` has no `turned_profile_not_spanned`, `geometry_issues: 3`.

Running the emitted script over the same 132 bytes reports it:

```
by_code: {'hole_requirement_missing': 4, 'diameter_leader_target_unverifiable': 2,
          'turned_profile_not_spanned': 1}
passed: True  errors: 0  warnings: 7  geometry_issues: 5
```

So the identical physical shortfall on the identical part is reported through one front door and
silent through the other. This is not a regression — the old `raise` was also inside
`if model is not None:` — but the PR's fixture table reads as if `draftwright 132.step` now warns,
and it does not.

It also contradicts the PR's own registration argument. `turned_profile_not_spanned` went into
`_GEOMETRY_AWARE_CODES` because *"the shortfall is about the part, not about where an annotation
landed"*. If it is about the part, how the model arrived is not a reason to skip it — and ADR 5
invariant 2 has lint judging the placed drawing and the geometry, not the provenance of the IR.

**Ask:** either drop the `_model_declared` gate (the check needs only `features` + `bbox` +
`orientation` + `single_solid`, all available on the detected path, and `pm` is the same object in
both cases), or replace the comment with the real reason. What is not acceptable is a justification
the PR's own evidence refutes.

### 2. The argument that makes this safe is missing from the PR

You asked whether the drawing still loses the overall height. **It does not, anywhere I could make it
happen** — and the reason is structural, not luck:

`model/compiled.py::_compile_overall_height` suppresses the overall-height rung only when
`_step_chain_covers_extent(model, groups_out, "z")` is true:

```python
if model.orientation == "z":
    if step_chain_approved:
        return None, ApprovedContingency("step_length", ladder, inactive), [inactive]
    return ladder, None, []          # <- chain does not tile: dim_height is KEPT
```

That is the same tiling question this lint asks, asked independently from the approved plan. So a
part that trips the warning keeps `dim_height`. Measured, `dim_height` is present in every fixture in
the converted tests and on the real part:

| fixture | misuse | `dim_height` |
|---|---|---|
| `.step(boss)` on `Box(90,64,38)+Cylinder` | yes | `48` |
| stepped boss on plate | yes | `40` |
| interior-gap cylinder (`build_drawing(model=…)`) | yes | `40` |
| CADGenBench 132, emitted script | yes | `140` |

For contrast, the *correctly* declared full-span version (`step(length=40)` on the `z 0..40` cylinder)
draws **no** `dim_height` — `m_steplen0: '40'` carries it. The suppression premise is doing its job;
#631's symptom cannot co-occur with this warning.

That is the reason a warning is sufficient, and it is verifiable. The PR argues instead from *"there
is no third option"*, and `coverage.py`'s docstring ends on *"Raising made that visible"* — which
leaves a reader believing the warning still guards a dropped height. Please state the compiler fact
in the docstring; it is the load-bearing half.

### 3. The reported gap is wrong when coverage resumes after it — and the PR body repeats the wrong figure

The walk stops at the first break and then reports `covered..hi_z`. On 109 that is measurably an
overstatement:

```
tol = 0.42
all spans: [(0,57.42),(57.42,67.42),(67.42,87.42),(87.42,93.42),(93.42,113.42),
            (113.42,119.42),(113.42,343.13),(366.82,420.0)]
covered -> 343.13   hi_z: 420.0
REPORTED gap: 343.13..420
spans lying INSIDE the reported gap: [(366.82, 420.0)]
```

`366.82..420` is described by a declared step, yet the message says that stretch is undescribed and
"dimensioned by nothing". The PR body carries the same figure — *"7 real steps leave z 343.13..420
undescribed"* — and it is 23.69 mm, not 76.87 mm. Report the first genuine gap (`covered..next_lo`),
or say "from 343.13" without claiming the far end.

### 4. Neither the severity nor the `_GEOMETRY_AWARE_CODES` entry is load-bearing

Mutation, both at once — removed `"turned_profile_not_spanned"` from `_GEOMETRY_AWARE_CODES`
(`drawing.py`) **and** changed `severity="warning"` to `"info"` in `coverage.py`:

```
uv run pytest -n auto --dist worksteal -q
8445 passed, 5 skipped, 13 xfailed in 248.62s     exit 0
```

Full fast tier green. All seven converted tests assert only `by_code[...] == 1`, and no test in the
repo mentions the code otherwise (`grep` over `tests/`: seven hits, all `by_code`). So "loudly in
lint" currently rests on two facts nothing pins. One assert in the #631 test — that
`lint_summary()["geometry_issues"]` rises, or that the issue's `severity == "warning"` — closes both.

**The guards that *are* load-bearing** (asked for, verified):

- `lint_turned_profile_span` → `return []` unconditionally: **all seven** converted tests fail
  (3 in `test_make_drawing.py`, 4 instances in `test_issue_1357_plural_turned_profiles.py`, counting
  both `order` params). None is decorative.
- `"turned_profile_not_spanned"` removed from `_UNSCORED_CODES`:
  `TestEveryLintCodeIsClassified::test_no_lint_code_scores_on_an_unstated_axis` fails. Confirmed the
  audit really sees it: `_emitted_codes()` returns it among 110 literals, `stages == {None}`.

"Unscored" is the right call, and for a better reason than the PR gives: scoring it on completeness
would derive a denominator from the IR, which **ADR 5 invariant 2** forbids. Worth saying in the
`quality.py` comment — "an omission is not a false statement" only rules out fidelity.

### 5. Two comments in the converted tests are now false

- `tests/test_make_drawing.py:9343` — *"the steps cover only the boss, not the plate below, so the
  overall height would be dropped"*. Measured on that exact fixture: `dim_height: '40'` is drawn.
- `tests/test_make_drawing.py:9367` — *"…so this must still raise"*, in a test that now asserts a
  warning.
- `:9316` — *"silently dropped the overall height (net −1 annotation, no height dim)"* reads
  present-tense over a fixture that draws `dim_height: '48'` and 8 annotations. Historical, but it is
  the sentence a future reader will act on.

You corrected the *headlines* of these tests and left the bodies of the comments describing the old
world. `sheet_emit.py:25` has the same problem from the other side: *"Fidelity: the script reproduces
a lint-clean drawing of the same features"* — the emitted script for 132 now deliberately produces a
drawing carrying `turned_profile_not_spanned`.

### 6. Which paths reach it, measured

Reaches (all confirmed to report `turned_profile_not_spanned: 1`):

- `Sheet(...).build().lint_summary()`
- `build_drawing(part, model=…)`, including `scale=0.5` and `scale=2.0`
- the drawing built by running a `generate_sheet_script` output (132: emit 3.6 s, build+export 17.2 s)
- after `dwg.repair()`, and on repeated `lint()` calls — it is recomputed from `_part_model` +
  `_analysis` on every call and never stored, so `orchestrator.reset_issues()` cannot reach it.
  **Your second-attempt failure mode is genuinely closed**, and a repack cannot defeat it for the same
  reason (the check reads final single-owner build state, not a recorded issue).

Does **not** reach:

- the automatic/detection build — §1
- `lint(physical=False)` — measured empty. Correct (repair's allowlist), but it means the repair loop
  never sees it.
- `Drawing.report()` on a declared build: raises
  `ReportUnavailableError: accepted occurrence ownership is unavailable for this drawing; raw
  automatic recognition is required by report schema version 3`. Pre-existing, but it means the CLI
  JSON sidecar cannot carry this warning on the only path that produces it.
- `cli.py` — it never calls lint at all (`grep -n "lint" src/draftwright/cli.py` → no hits).

So today the warning reaches a human only if they type `drawing.lint()`; the emitted script's trailer
merely suggests it. That is the honest scope of "reports instead of refuses" and it belongs in the PR
body in place of "loudly".

### 7. Predicate equivalence — I could not find a disagreement, with two caveats

- `kind == "step" / "groove" / "envelope"` is a bijection with `StepFeature` / `GrooveFeature` /
  `EnvelopeFeature`: `kind` is a `ClassVar`, each string is unique in `model/ir.py`, and the only
  `class X(Y)` lines in that file are the two `Protocol`s — no feature subclasses exist. Equivalent
  to the old `isinstance`.
- `tol`: `a.z_size` is `bb.max.Z - bb.min.Z` in both branches of `analyse()` (`analysis.py:963`, and
  the `_reuse` copy at `:840`), so `1e-3 * max(hi_z - lo_z, 1.0)` is the same float. Equivalent.
- The `envelope` fallback, the single-solid gate and the groove handling are byte-for-byte the same
  logic in a different order. Equivalent.
- **Caveat A:** the new code evaluates `f.frame` and `f.span` before the kind check narrows the type
  (`getattr(f.frame, "axis", None)` still dereferences `f.frame`). A duck-typed feature with
  `kind="step"` and no `.frame`/`.span` now raises *inside lint*, where the old `isinstance` ignored
  it. Not reachable via `Sheet`/`declare`; a note, not a defect.
- **Caveat B:** `self._analysis.bb` is dereferenced unguarded, inside a block whose own comment says
  `a` may be `None` ("a manually-built Drawing, or lint called mid-build"). Safe only because
  `_build.analysis` is set at `builder.py:598` and `_model_declared` at `:619`, so True implies
  non-None. A `getattr` costs nothing and makes it not depend on that ordering.

### 8. Nit

`bbox` means a build123d `BoundBox` everywhere else in `linting/coverage.py`
(`lint_prismatic_coverage(..., bbox=a.bb)`, `_principal_boundary_plane(face, bbox)`,
`_radial_outer_arc_count(part, bbox)`). Here it is a `(lo_z, hi_z)` float pair. `z_extent` would stop
the next caller passing `a.bb`.

---

### Architecture

Good fit, and better than the raise was. ADR 5 invariant 7 is explicit that an approved measurement
that cannot be placed *"is reported against that measurement, without gating the build"* — moving a
diagnostic out of the build gate and into lint is the pattern, not an exception to it. ADR 5's
"refuses rather than shrinks" is scoped to the machine-readable report (invariant 9), not the sheet,
so no record needs to change. `linting/` still imports no `draftwright.model` (fast tier green
includes `test_import_boundaries`). The one architectural loose end is §1: a geometry-aware code
gated on IR provenance.

### Evidence

- Fast tier: `8445 passed, 5 skipped, 13 xfailed in 244.71s`, **exit 0**.
- Slow tier (`-m slow -n auto --dist worksteal`): `62 passed, 2 xfailed in 227.73s`, **exit 0**.
- `ruff format --check src tests` **exit 0**; `ruff check src tests` **exit 0**.
- Mutations: 4 applied and reverted, each result above. Working tree returned clean on
  `fix/1132-generate-and-warn` (`git status --porcelain` empty).

### Reproduction note

The CADGenBench fixtures are not in this repo, so the PR's headline table is not reviewer-reproducible
from `pzfreo/draftwright`. I used a local copy —
`.../results/build123d-mcp-0.3.85.dev0-opus-5-xhigh-complete81-r1/{132,109}/output.step`, `z 0..140`
and `z 0..420`, one solid each, both importing cleanly through `analysis._import_step` — and the
figures matched yours (`113..140`, `343.13`). `coverage.py`'s docstring sentence about *"the two
CADGenBench fixtures that still reach this"* cites evidence nobody else can run and will rot; either
name where it lives or drop it.

Also confirmed in passing: 109's detected build carries `label_vs_measured: 1` — #1571 is real.

https://claude.ai/code/session_01N3VevbCf1UWuPzj3G927ef

Issue #1602, opened 2026-09-13
Classify lint by veracity / coverage / legibility, and fail CI only on veracity


pzfreo:
## Review checkpoint — 2026-09-13

Review the proposed gate separately from the factual taxonomy. `plan_incomplete` is not the only error-severity code: current tests and code also use errors such as `scale_not_stated`, `pmi_source_unknown` and material `label_vs_measured` contradictions. The acceptance text referring to open #1596 is also historical; #1606 merged, with the remaining square-pattern and refused-pattern-ledger cases in #1611 and #1607.

The default `--fail-on veracity` proposal needs an explicit maintainer policy decision against ADR 5's honest completeness contract. It must not make an incomplete drawing appear ready or remove coverage findings. #1608 is still open and the review reproduced a remaining false positive for `.format(decimals=0)`; resolve that before treating the rounding check as fully repaired. This checkpoint changes no ADR or gate.

<details>
<summary>Earlier scope and evidence (retained for context)</summary>

Part of #1595. Depends on #1600 (`label_vs_measured` false positive) — a hard-fail register cannot carry one.

## The problem

`[error] plan_incomplete` and `score 0.000`, and the CLI exits **0**. CI cannot distinguish a good sheet from a bad one.

The obvious fix — exit non-zero on error severity — is wrong, and the reason is worth keeping: on any complex part the lint will never be clean, so it swaps a constant green for a constant red. Severity is not the discriminator.

## The proposal

What distinguishes the #1595 findings is **what kind of claim each defect is**:

| class | examples | scales with complexity? | gate |
|---|---|---|---|
| **veracity** — the sheet asserts something untrue about the solid | #1596 fictitious bolt circle, `label_vs_measured` | no — never acceptable at any size | **hard fail** |
| **coverage** — the sheet does not say everything | #1597 unconstrained seat, #1598 dropped hex pockets, `plan_incomplete`, `location_ref_dropped` | yes, unavoidably | ratchet |
| **legibility** — the sheet cannot be read | #1601 ink overlap, 16× `feature_leader_crossing` | yes | ratchet |

A drawing stating a datum the part does not have is wrong whether the part has 6 features or 600. A drawing that omits something is more likely to omit something as parts grow.

## This is less work than it reads

`linting/quality.py` already has most of the machinery:

- `_FIDELITY_CODES` — essentially the veracity register today (`label_vs_measured`, `claimed_value_absent`, `declared_feature_absent`, `scale_not_stated`, …);
[...]

Issue #1595, opened 2026-09-13
Epic: what a real printed part exposes — a missing mating dimension, an invented datum, and a lint that cannot see either


pzfreo:
> **Broken out into an epic.** Every finding below now has an issue; this one is the umbrella and the record of the part. Claims were re-verified by execution against current `main` (post-#1591) before splitting — one did not fully reproduce, noted in #1599.

## The work

**Veracity — the sheet says something untrue.** Never acceptable at any part size.

| | | |
|---|---|---|
| #1596 | Refuse a bolt-circle callout that only four holes can support | A2 |
| #1600 | `label_vs_measured` misreports display rounding as a topology error | A5 |

**Coverage — the sheet does not say everything.**

| | | |
|---|---|---|
| #1597 | Lint cannot see geometry no recogniser proposed | A1 (draftwright's half) |
| #1598 | Six functional hex pockets refused and undimensioned | A3 |
| pzfreo/quiddity#594 | Open semicylindrical trough not in the vocabulary | A1 (recogniser's half) |
| pzfreo/quiddity#593 | Six coaxial bores across air gaps reported as one | A4 |

**Legibility — the sheet cannot be read.**

| | | |
|---|---|---|
| #1599 | The same feature is dimensioned twice | B1 |
| #1601 | Overlap, view overflow, broken rendering, inconsistent formatting | B2–B6 |

**Policy and process.**

| | | |
|---|---|---|
| #1602 | Classify lint by veracity/coverage/legibility; fail CI only on veracity | C1 |
| #1603 | Outputs written to the working directory, not beside the input | C2 |
| #1604 | Section view by default for a part whose detail is internal | B7 |

## Suggested order

**#1596 first.** Not the worst outcome — #1597 is — but it is a veracity defect with a small, bounded fix, and it is the cheapest thing here that stops the tool asserting something false.

**#1597 is the structural finding.** Its two halves have different owners: the trough vocabulary gap is quiddity#594, behind the recogniser boundary; the lint blind spot is ours and is the more important half. Fixing quiddity#594 fixes *this part*; it does not fix the class. ADR 5 casts lint as the independent judge, and on coverage it currently shares the recognisers' blind spots by construction.

**#1600 blocks #1602** — a hard-fail register cannot carry a false positive.

**#1603** is small, self-contained, and wrote files into a git repo.

## Verified by execution

[...]

Issue #1176, opened 2026-08-15
PPP0101 omits defining radii and dovetail dimensions while lint scores 1.0


pzfreo:
## Problem

A controlled development benchmark using the build123d v0.11.0 Too Tall Toby PPP0101 bearing-bracket fixture produces a visually clean automatic drawing with:

- lint score **1.0**
- passed: **true**
- warnings/errors/infos: **0**

The drawing is not definitionally comparable with the fixture's reference drawing. The automatic Sheet script contains a hole, slot, envelope, a two-level prismatic ladder, and four R6 fillets, but it does not communicate several profile-defining requirements present in the pinned reference:

- upright outer profile R26
- web-to-base root R9
- upright thickness 12 mm
- dovetail flank angle 60 degrees
- dovetail height 8 mm
- dovetail foot width 16 mm
- dovetail centre gap (18 mm reference)

The retained STEP has a 115 × 50 × 68 mm BREP envelope. Its generated step-ladder declaration already carries the 8 and 15 mm levels and the ±13.619 mm supports, so at least part of the omitted dovetail intent reaches the declarative model without becoming a complete dimension basis.

## Evidence

Source: build123d v0.11.0, fixture `ttt-ppp0101` ("Party Pack 01-01 Bearing Bracket").

Reference PNG SHA-256:

```
3f97d66a301f298a5a22c62d09f4645a6808b3884e434235a070801cd0e3f6bb
```

A separate reference-assisted Sheet revision could render R26, 16 mm, and (18 mm) declarations from the same STEP. That demonstrates that these are not merely unavailable source data. R9 was declared but then reported as `fillet_dropped`, which is separately relevant to #1166.

This issue is about the automatic Stage 1 completeness claim: unsupported or unrecognised requirements must not disappear behind a perfect lint score.

## Expected architecture

- Recognise the upright/root radii and the defining dovetail profile, or report the exact geometry as unrecognised/unverifiable.
- Plan a non-redundant profile dimension basis.
- Surface missing/dropped defining dimensions through the existing completeness outcomes.
- Never convert absent recognition evidence into a clean completeness result.

This should be a synthetic public regression derived from the geometry, not a committed copy of the third-party PNG/STEP.

## Acceptance criteria

- [ ] A synthetic bearing-bracket fixture exposes structured R26/R9 and dovetail-profile requirements, or explicit unsupported/unverifiable outcomes.
[...]

PR #1608, opened 2026-09-13
Stop reporting the drawing's own rounding as a topology error (#1600)

### LINKED ISSUE #1600: label_vs_measured misreports display rounding as a topology error
## Review checkpoint — 2026-09-13

Earlier report #1571 is consolidated here as a duplicate; retain its CADGenBench 109 / 4.84 → 4.8 evidence. PR #1608 remains open.

The 2026-09-13 review reproduced an additional supported case on PR head d1ecfb92: `Sheet(Box(4.45, 30, 12), page="A3", scale=5)` with authored `width.length.format(decimals=0)` prints `4` but still receives a 10.1% `label_vs_measured` topology-error diagnosis. Full reproduction and the required distinction between explicit coarse precision and trimmed automatic labels are posted on #1608. Keep this acceptance visible until corrected; #1610 separately owns location formatting.

<details>
<summary>Earlier scope and evidence (retained for context)</summary>

Part of #1595. Blocks #1602 (the veracity gate), because a hard-fail class cannot carry a false positive.

## What happens

```
[warning] label_vs_measured: Dim '4.5': label value 4.500 differs from measured path
length 4.450 by 1.1% — possible axis swap or wrong endpoint
```

The measurement is 4.450. The label is `4.5` because the sheet rounds to 1 dp. Nothing is wrong with the geometry, the endpoints or the axis — the drawing states 4.450 to the precision it prints at. The lint then compares its own rounded output against the unrounded source and reports the difference as a suspected topology error.

## Why it matters more than one noisy warning

`label_vs_measured` is in `_FIDELITY_CODES` — the register of codes meaning *the sheet says something untrue*. #1595's C1 proposes hard-failing CI on exactly that register. A code that fires on correct rounding cannot be in it: the gate would be red on any part whose dimensions do not land on the display grid, which is most of them.

So this is a prerequisite, not a nice-to-have.

## Suggested fix

Compare the measurement against the label **at display precision**: format the measured value with the same decimals the label uses, then compare. A genuine axis swap or wrong endpoint survives that (it differs by far more than half an ulp of the display grid); rounding does not.

Keep the existing materiality threshold for the genuinely-close cases — this is about not treating the rounding step itself as evidence.

## The related half

#1595's A5 also notes the part carries 0.05 mm clearances while the sheet prints 1 dp, so the drawing cannot express its own design intent. That is a separate question — per-part or per-feature display precision — and worth its own issue if it is wanted; this one is only about the lint's false positive.

## Acceptance

The fixture builds with no `label_vs_measured` for the `4.5` dimension. A deliberately mislabelled dimension still reports one — add it as a counterexample so the fix cannot be "stop checking".

</details>

pzfreo:
Closes #1600. Part of epic #1595. **8522 passed, 0 failed, 14 xfailed**, real-part canary 8 passed, `scripts/pr-check --static` clean.

> **Reviewed independently and revised.** The review found two regressions I introduced and one misleading message; all three are fixed in `2bf7442`. Summary at the bottom.

## The defect

A sheet at one decimal place prints 4.450 as `4.5`. The check then compared `4.5` against 4.450 and reported:

```
[warning] label_vs_measured: Dim '4.5': label value 4.500 differs from measured path
length 4.450 by 1.1% — possible axis swap or wrong endpoint
```

Nothing is wrong with the geometry, the endpoints or the axis.

## Why it fired

The comparison is **relative**: `|label − measured| / measured > 0.005`. Rounding to one place moves a value by at most 0.05 mm, which is 0.05% of a 100 mm dimension and 1.1% of a 4.45 mm one. **The check fired on smallness, not on wrongness** — the same rounding passed silently on the long dimension and was reported as a topology error on the short one.

## The fix

A label states its own precision by being written: `4.5` shows one place, `4.45` two, `12` none. A difference within **half of the last displayed place** is the label doing its job. Beyond that it is still reported — an axis swap or a wrong endpoint misses by far more.

Reading the precision off the label rather than from `decimal_precision` matters: a placed dimension does not carry a draft, and the sheet's precision will not be one number once `.format(decimals=…)` reaches automatic dimensions.

## Why it is a prerequisite, not just noise

`label_vs_measured` is in `_FIDELITY_CODES` — the register of codes meaning *the sheet says something untrue* — and #1602 proposes hard-failing CI on exactly that register. A code that fires on correct rounding cannot be in it.

## Evidence

On `tests/fixtures/whistle_frame_reference.step`, **7 of 25** numeric dimension labels differ from the model value purely by rounding (up to 0.05 mm). The warning is gone; nothing else on the sheet changes.

| mutation | result |
|---|---|
| rounding allowance removed | 1 named test fails |
| displayed places hardcoded to 3 | 7 named tests fail |

## What this does *not* do

It stops the tool misreporting rounding. It does not change the rounding. See the note on the issue about display precision — on that same part six labels state a nominal the model does not have, and 1 dp is currently a hardcoded constant with no policy behind it. That is a drafting decision, not a bug fix, so it is written up separately rather than decided here.

https://claude.ai/code/session_01BeWRqP9V6ZrAnHNW2Cibn6

## Independent review, and what it changed

Reviewed in a fresh context with no knowledge of the author's reasoning. It returned **request changes**, correctly, on three findings:

1. **The allowance was 10× too loose where it mattered.** `_fmt` trims a trailing `.0`, so a 1 dp sheet prints 40.0 as `40` — zero places in the string, 0.05 mm of real precision behind it. Reading places off the string granted **half a millimetre** of slack to every whole-numbered label, so `40` on a 40.45 mm path passed as ordinary rounding. That is a 1.1% mislabel the check used to catch, moved out of the fidelity register and into an `info` asserting it is ordinary — the opposite of what this PR exists to enable. My stated reason for not using the sheet's precision was also false: an engine-built dimension carries it on `_dw_spec.draft`. The sheet is now the floor; a bare numeric label may raise it.

2. **`nominal_rounded` counted genuine mislabels as rounding**, and being the largest shift one became the "worst case" named — displacing the thing the info exists to surface. Now gated on the rounding test. The test named for this never tested it.

3. **The remedy did not work for the dimensions it named.** Six of seven offenders on the fixture are location dimensions, and `.format()` refuses a location intent outright. The message no longer prescribes something that raises `ValueError`; the gap is #1610.

It also found that the test stand-ins built raw `Dimension`s carrying no `_dw_spec`, so they exercised the no-draft fallback rather than the path every real sheet takes — which is why nothing caught (1). They now go through `_core._dim`, and a standing guard pins the real fixture's behaviour.

Two prose corrections: comments saying "a few microns" where the measurements are 0.05 mm and 0.027 mm, and this PR's own test count, which had been measured on the first commit and not re-run.

What it confirmed: "7 of 25 numeric labels" exact, "silent on every golden fixture" exact, both stated mutations reproduce, `_UNSCORED_CODES` is the right register, and the `issues[0]` → select-by-code change is a real improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,22 @@
 
 ### Fixed
 
+- **`label_vs_measured` no longer reports the drawing's own rounding as a topology error
+  (#1600).** A sheet at one decimal place prints 4.450 as `4.5`; the check then compared
+  `4.5` against 4.450 and said "possible axis swap or wrong endpoint". The comparison was
+  *relative*, so the same 0.05 mm of rounding passed silently on a 100 mm dimension and was
+  an error on a 4 mm one — it fired on smallness, not on wrongness.
+
+  A label now states its own precision: `4.5` shows one place, `4.45` two, `12` none. A
+  difference within half of the last displayed place is the label doing its job; anything
+  beyond it is still reported, and an axis swap or wrong endpoint misses by far more.
+
+  This is a prerequisite for #1602's proposed CI gate: `label_vs_measured` sits in the
+  register of codes meaning *the sheet says something untrue*, and a code that fires on
+  correct rounding cannot be hard-failed on.
+
+### Fixed
+
 - **A required dimension with nowhere to go now re-plans the sheet instead of being
   reported and left (#1590).** The automatic recovery ladder — larger scales on the
   selected page, then dropping the optional pictorial, then a larger page — already
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,28 @@
   register of codes meaning *the sheet says something untrue*, and a code that fires on
   correct rounding cannot be hard-failed on.
 
+### Added
+
+- **`nominal_rounded` says where the sheet's precision loses the model value.** Silencing
+  the false alarm above left nothing reporting it at all, and `4.5` for 4.450 is still a
+  real 0.05 mm between drawing and model — on a part carrying 0.05 mm clearances, the whole
+  clearance. One `info` per sheet, naming the worst case:
+
+  > 7 dimension(s) print a nominal the model does not have, rounded to the sheet's
+  > precision; the largest is '15.7' for 15.6500 (0.0500 mm). Within a general tolerance
+  > this is ordinary; where a fit depends on it, raise that dimension's places with
+  > `.format(decimals=...)`
+
+  It does **not** guess, and does not change any rounding. `4.450` locating a hinge axis is
+  design intent; `71.595` as an overall envelope is parametric fallout no drafter would
+  print in full — and telling them apart needs to know what the part is *for*, which a STEP
+  file does not carry (the same gap as #1597). So the engine reports and the author decides:
+  `.format(decimals=…)` already sets places per dimension and survives the `--script` round
+  trip.
+
+  Measured: 7 of 25 numeric dimension labels on `whistle_frame_reference.step`; silent on
+  every golden fixture, all of which are built from round numbers.
+
 ### Fixed
 
 - **A required dimension with nowhere to go now re-plans the sheet instead of being
--- a/src/draftwright/linting/quality.py
+++ b/src/draftwright/linting/quality.py
@@ -312,6 +312,13 @@ _UNSCORED_CODES = frozenset(
         "authored_omission",
         "axial_length_missing",
         "boss_height_missing",
+        # Where the sheet's precision prints a nominal the model does not have (#1600). Not
+        # fidelity: the label IS the measurement written to the places the sheet uses, and a
+        # general tolerance covers the difference — the check that treated it as a false
+        # statement was the defect. Not legibility either; the sheet reads perfectly. It is a
+        # statement about precision POLICY, which has no register of its own until #1602
+        # builds one.
+        "nominal_rounded",
         # The `*_withheld` family (#1216), kept together and in place alphabetically. Each is a
         # claim the drawing declined to make because it could not make it honestly, and said
         # so: a jittered pattern pitch, a step height the page cannot carry, and an `n×` mark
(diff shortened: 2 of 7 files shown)


pzfreo:
Review finding on `d1ecfb92d239dbdef77c1f1ed9bcc22cf608a56a`: **P2 — honor explicit coarser display precision before closing #1600.**

`_displayed_decimals` uses `max(sheet_precision, places_in_label)`. This repairs trimmed default `.0` labels but ignores the supported `.format(decimals=0)` override. The sheet draft still has precision 1, so a correct whole-number label receives only 0.05 mm allowance.

Executed on this PR head through the public Sheet API:

```python
from build123d import Box
from draftwright import Sheet

s = Sheet(Box(4.45, 30, 12), page="A3", scale=5)
e = s.envelope()
s.authored_dimensions()
s.dimension(e, "width.length").format(decimals=0)
d = s.build()
print(d.get_annotation("m_env_width").label)  # 4
print([(i.code, i.message) for i in d.lint()
       if i.code in {"label_vs_measured", "nominal_rounded"}])
```

Result: `label_vs_measured` reports **10.1% — possible axis swap or wrong endpoint** for the correctly formatted `4`. At decimals=1 the same declared geometry prints `4.5` and gets only `nominal_rounded`; decimals=2 prints `4.45` and gets neither. The public API accepts 0 through 15, so this is an existing supported path, not an invalid declaration.

Please carry/use the actual approved per-dimension display policy at the annotation/lint boundary, with a regression for explicit zero places as well as trimmed automatic whole numbers. Removing the sheet floor globally would reopen the mislabel regression your prior review caught. This is ADR 4 authored-format fidelity and ADR 5 truthful diagnosis; no new formatting policy is required.

Validation: the PR's structural lint, #1349 display-precision and #1176 fidelity modules pass **157 tests** locally on Python 3.14, but do not catch this public-path counterexample. #1571 is consolidated into #1600 as an earlier fixture; #1610 remains the separate location-addressing limitation.

PR #1731, opened 2026-09-20
Persist generated replay assessments (#1715)

pzfreo:
## Symptom and evidence

- User-visible problem: a generated Python edit could be replayed, but no exact post-build artifact bound that script, source, outputs, declarations, layout, and lint result together.
- Fixture/reproducer: generated script replay tests, including edited, failed, stale, foreign, and deterministic replay cases.
- Before/after result: successful generated replays atomically write a versioned assessment; failures cannot leave stale success evidence.
- Mutation that makes the guard fail: alter the script, fail build/export/report, or place foreign content at the derived destination.

## Contract and scope

- Smallest contract this change tests: exact same-run replay provenance and report persistence.
- Relevant ADRs: ADR 1, ADR 3, ADR 4, ADR 5.
- Explicitly deferred: comparison policy and fixed-denominator scoring are the next stack layer.

## Validation

- [x] Focused replay assessment and generated-script tests
- [x] Ruff, formatting, mypy, codespell, strict docs
- [x] Failure and stale-artifact mutations observed
- [x] Generated-script path checked end to end
- [x] Recognition runs only during generation; replay uses declared IR

## Stop check

- [x] No repeated review round invalidated the persistence strategy
- [x] Comparison remains isolated in the next PR

## Contributor License Agreement

- [x] I have read the CLA and agree to it.

Closes #1715

PR #1732, opened 2026-09-20
Compare replay assessment evidence vectors (#1712)

### LINKED ISSUE #1712: Define a baseline-to-candidate drawing assessment and no-regression contract
## Parent

Child of #1708; builds on #1277, #1710 and #1711.

## Problem

An agent needs to decide whether an edited script produced a better drawing. The legacy diagnostic score cannot answer that: it ignores info findings, saturates at zero, and mixes neither semantic coverage nor fidelity. Averaging or taking the minimum of quality axes would still hide the reason for change and contradict ADR 5.

## Outcome

Provide a deterministic comparison of two assessments over the same source authority, expressed as independently reviewable deltas rather than one score.

The comparison should report:

- resolved, introduced and unchanged lint findings;
- recognised requirement transitions by physical owner and parameter;
- placed carrier changes, value/tolerance/member/axis meaning and representation substitutions;
- fidelity and target-validation changes;
- layout changes for the selected defect;
- unscored/unclassified findings;
- unavailable or incomparable evidence;
- intentional authored changes supplied by the caller.

## No-regression policy

A candidate cannot be automatically preferred when it:

- loses, substitutes or weakens a fixed expected physical requirement;
- introduces missing, dropped, unsupported or unverifiable outcomes;
- introduces a fidelity contradiction or unclassified lint code;
- resolves a layout issue by deleting its required annotation;
- compares different source hashes, incompatible schema/provider versions or irreconcilable identities;
- converts unavailable evidence into an implied pass.

Restraint and manufacturing readiness remain independently unavailable until their own evidence exists.

## Acceptance criteria

- Same-count/same-kind and equal-value owner substitutions are detected, reusing the delivered semantic comparison foundations from #1006/#1277.
- Comparison refuses different STEP hashes and reports incompatible run options/schema versions.
- Caller-authorised changes are separated from incidental regressions.
- The output identifies why a candidate is preferred, rejected or incomparable; no composite quality scalar is introduced.
- A fixed-denominator canary catches a shared omission present in both baseline and candidate.
- A layout fixture proves that removing a crossing while preserving all expected measurements is preferred, while deleting the crossed dimension is rejected.
- Output ordering and identifiers are deterministic for byte-identical inputs on one version.

pzfreo:
## Symptom and evidence

- User-visible problem: completeness and lint scores could look better after deleting or retargeting an annotation, so one scalar could mislead an autonomous editor.
- Fixture/reproducer: pure assessment comparisons covering deletion, restoration, owner substitution, engineering meaning, carrier and pin changes, lint multisets, denominator shrink, and incompatible provenance.
- Before/after result: comparison returns separate semantic, completeness, fidelity, layout, unscored, unavailable, restraint, and manufacturing-readiness evidence with a fail-closed policy decision.
- Mutation that makes the guard fail: delete a fixed requirement, swap physical occurrences, change value/tolerance/axis/member/pin, shrink the denominator, or change source/producer/run provenance.

## Contract and scope

- Smallest contract this change tests: pure JSON comparison of compatible replay assessment v2 documents.
- Relevant ADRs: ADR 3, ADR 4, ADR 5.
- Explicitly deferred: the real-part two-build canary is the top stack layer; physical completeness and manufacturing readiness stay explicitly unavailable.

## Validation

- [x] 80 focused comparison/replay tests
- [x] 98 percent changed-line coverage before final pin-state addition; focused pin tests also pass
- [x] Ruff, formatting, mypy, codespell, strict docs
- [x] Every named mutation is rejected or separated as an intentional change
- [x] Pure comparison adds no CAD builds

## Stop check

- [x] Review fixes preserved the same declaration plus occurrence association strategy
- [x] Real-part proof remains isolated in the next PR

- [x] I have read the CLA and agree to it.

Closes #1712


pzfreo:
Upstack CI review found one parent-test compatibility issue: the new v1 replay test double did not implement the public model()/measurement_snapshot() surfaces added by this v2 layer. The fix stays in #1731, supplies empty aligned contracts, and was verified locally on both branches. This PR’s replacement run has passed lint, changed-line coverage, canaries, and nine Linux shards; the final shard is still running.

PR #1744, opened 2026-09-20
Reject dirty automatic page and scale proposals

pzfreo:
## Symptom and evidence

- User-visible problem: automatic page/scale planning could prefer a proposal with hard overlap lint merely because it retained more annotation outcomes.
- Fixture/reproducer: focused proposal-verdict tests plus the generated object-reference workflow and the CTC01 real-part canary.
- Before/after result: invalid automatic proposals are rejected; generated examples report the settled page; an already-resolved overlap no longer receives stale agent-improvement credit.
- Mutation that makes the guard fail: restore outcome-count-first proposal ranking or remove the hard-validity verdict.

## Contract and scope

- Smallest contract this change tests: hard layout validity precedes required-outcome completeness when comparing automatic proposals.
- Supporting bound: expanded feature-leader inventories are measured and searched with conservative broad phases, conflict components, and exact dominance pruning; bounded fallback keeps the established producer floor.
- Relevant ADRs: ADR 2 and ADR 5.
- Explicitly deferred: default interior eligibility for additional feature families, whole-annotation dimension candidates, GD&T candidates, and a public `region=` DSL.

## Validation

- [x] `scripts/pr-check --static`
- [x] Focused page/scale verdict tests
- [x] Leader assignment, candidate-region, and cross-pass fallback clusters
- [x] Generated object-reference documentation canary
- [x] CTC01 agent-assessment real-part canary (56 s locally, below its 120 s bound)
- [x] AP242 generated-sheet test (72 s locally, below its 300 s bound)

## Review standard

Reviewed against Google's code-review guidance for correctness, design, complexity, tests, naming, comments, documentation, and every changed line. The exact exterior-only budget expectation remains distinct from the later interior-enabled expectation.

## Stop check

- [x] New prerequisites were kept within the bounded shared-solver contract

- [x] I have read the CLA and agree to it.


pzfreo:
Fresh CI review update: Python 3.10/3.12 lanes exposed a planning-cost regression in the AP242 generated-sheet canary (300s timeout). The hard-validity ladder was compiling two smaller and two larger scales before page escalation—four complete drawing compiles. I reduced that bounded policy to the nearest ISO scale in each direction, then the existing standard-page tail, and added a unit guard that pins exactly those two scale probes. The same AP242 test passes locally and fell from 68s to 44s (35% faster) without changing its verdict or generated semantics. Static checks pass; fresh CI is running.

Issue #1738, opened 2026-09-20
Allow annotations to use proven whitespace inside drawing views


pzfreo:
## Problem

Draftwright currently treats an orthographic view's complete rectangular extent as unavailable to feature-label placement, even where the projected drawing contains large regions with no visible edges or annotation ink. Feature leaders therefore consume exterior strips unnecessarily, increasing crossings and causing otherwise viable smaller sheets to drop dimensions.

The reference drawing discussed alongside the [Xometry technical-drawing guide](https://xometry.pro/en/articles/prepare-technical-drawing/) uses whitespace inside the object view for callouts. Draftwright should be able to do the same without exposing arbitrary annotation coordinates or creating a second placement engine.

## Desired model

Interior placement should be another candidate family in the existing shared placement solve:

```text
Sheet / Drawing intent
  -> semantic annotation producer
  -> interior + exterior candidates
  -> shared placement solver
  -> completeness and lint validation
```

The core rule remains unchanged: callers declare what to annotate and may constrain or rank natural candidates, but they do not hand-place feature annotations at page coordinates.

An interior candidate is eligible only when:

- its complete measured label box fits the intended view region;
- the box clears projected visible/hidden edge ink by the normal drafting clearance;
- its leader geometry satisfies the existing material/routing policy;
- it clears all already placed annotation ink;
- it participates normally in pairwise candidate conflicts; and
- page and title-block bounds remain hard constraints.

Policy B must not deliberately retain an interior candidate through fixed annotation ink. Exterior compatibility behavior can remain unchanged.

## DSL relationship

This does not inherently conflict with `view=`, `side=`, `priority`, or pinning. Those already represent bounded placement intent while the solver retains coordinate authority.

A possible eventual vocabulary is:

```python
s.dimension(feature, "diameter", region="interior")
dwg.callout(feature, region="interior")
```

with:

- `region="auto"`: interior and exterior candidates may compete;
- `region="interior"`: only proven-clear interior candidates are eligible;
- `region="exterior"`: traditional exterior corridors only;
- `side="left"` / `"right"` / `"above"` / `"below"`: a particular supported exterior corridor.

This is placement/display intent, not physical feature geometry. Infeasible explicit intent must fail or drop honestly; it must not silently move to an incompatible region. The API decision can be deferred until automatic placement is robust, but the implementation should use a typed internal policy rather than an environment-variable branch.

## Prototype evidence

[...]

PR #1747, opened 2026-09-20
Complete gusset requirement integration

pzfreo:
Repairs the merged gusset-family integration contracts: fail closed when provider evidence is unavailable, preserve one exact requirement row per physical rib occurrence, and update the explicit family/assignment rosters. This is a prerequisite for #1738 because current main does not pass the full suite without it.


pzfreo:
Review disposition: reviewed the incremental diff under Google’s design/functionality/complexity/test/naming/comment/style criteria. I checked the absent-provider-evidence boundary, exact one-row-per-rib-occurrence cardinality, grouped spacing semantics, and every explicit family roster update. The tests fail if cardinality is collapsed or the family is omitted. No blocking findings remain; this PR’s CI is green.

PR #1748, opened 2026-09-20
Preserve recognition evidence in document sheet reports

pzfreo:
Document members are authored sheets over one source-owned recognition run. Preserve schema 8 layout evidence while also projecting the exact shared recognition requirement ledger, so live member edits remain auditable without losing declared drawing evidence.


pzfreo:
Review disposition: reviewed the incremental production diff and its report-schema test. I checked that schema 8 declared layout evidence remains authoritative, recognition is projected only when both evidence and ownership exist, the requirement snapshot is reused without changing drawing content, and source identity comes from the retained snapshot. Review previously caught the stale schema-3 expectation and fixed it to schema 8. No blocking findings remain; fresh CI is running.

PR #1750, opened 2026-09-20
Preserve semantic layout regression contracts

pzfreo:
Refreshes layout regressions around the settled hard-validity policy: assert semantic completeness instead of stale coordinates, keep drop-path tests deterministic, and accept newly recovered off-axis locations. No production placement changes.


pzfreo:
Review disposition: reviewed every test-only layout change for semantic strength and determinism. The updates replace stale coordinate/scale outcomes with exact semantic completeness assertions, retain strict failure checks for unrecovered steps, and use a synthetic lint issue only where the test’s contract is the suggestion text itself. No blanket warning/error relaxation was introduced. No blocking findings remain; the current commit has a successful CI run.

Issue #1752, opened 2026-09-20
Stadium slots are dimensioned with no end-radius callout


pzfreo:
## Summary

A stadium (obround) through-slot is dimensioned width × length × position, with no
statement that its ends are radiused. The drawing of a 26 × 8 stadium slot is identical
to the drawing of a 26 × 8 rectangular slot.

Blocked on pzfreo/quiddity#708 — the provider proves the end caps and drops the radius,
so there is nothing to adapt yet.

## Reproduction

`build_drawing("draftwright-demo-bracket.step")` (the upright plate carries one obround
through-slot) gives the IR feature:

```
SlotFeature(frame=Frame(origin=(0.0, 0.0, 24.0), axis='x'), width_axis='z', long_axis='x',
            width=8.0, length=26.0, w_center=32.0, lo=-13.0, hi=13.0)
```

and places, for that slot, exactly:

```
m_slot0_width   m_slot0_length   m_slot0_pos
```

No radius annotation, on any view.

## Cause

`SlotFeature.parameters()` (`model/ir.py:1154`) returns two parameters and no more:

```python
return [
    DimParameter("length", "slot_width", self.width),
    DimParameter("length", "slot_length", self.length),
]
```

It cannot return a third, because `quiddity.Slot` carries no end-shape field for
`_convert_slot` (`model/detect.py:587`) to read. The geometric evidence does reach this
build — recognition returns the two caps as a cylinder group (concave, `u_extent = π`,
⌀8.0, axis `y`, at x = ±9.0, z = 32.0, spanning y 17.0–25.0) — but re-deriving
"these ends are radiused" from that substrate is geometry reasoning in the wrong repo
(ADR 3), so the fix is to consume a provider field.

## Work, once quiddity#708 lands

Round-trip per ADR 4 — recognise, emit **and** declare:

1. **Pin.** `quiddity==0.3.1` → the release carrying the field (`pyproject.toml:26`), and
   `("slots", "Slot")` from `_SCHEMA_1_RECORDS` to its new version in
   `recogniser_schema.py`.
2. **Adapter.** `_convert_slot` / `_member_slot` carry `end_radius` onto `SlotFeature`.
3. **IR.** `SlotFeature.end_radius: float | None`; `parameters()` appends
   `DimParameter("radius", "slot_end_radius", self.end_radius)` when it is set. Register
   `"slot_end_radius.radius"` in the `DimensionParameterId` literal
[...]

PR #1761, opened 2026-09-21
Keep circular leaders normal to their rims

### LINKED ISSUE #1760: Keep circular interior leaders normal to their rims
## Summary

Interior hole-callout candidates introduced by #1738 rotate label rays around a fixed arrow tip. For a circular target, that can make the selected shaft meet the circumference obliquely instead of along a radius.

This is distinct from #1752: that issue concerns a missing stadium-slot end-radius requirement; this issue concerns the geometry of leaders that already exist.

## Reproduction

```python

drawing = build_drawing("draftwright-demo-bracket.step", page="A3", pmi="off")
```

In the plan view:

- `hc_plan0`, `2× ⌀8 THRU`, has its tip on the right-hand hole rim but its interior shaft departs diagonally.
- `hc_plan1`, `⌀10 THRU`, has its tip on the right-hand bore rim but its interior shaft departs vertically.

For each leader, the 2D cross product of `(tip - circle_center)` and `(elbow - tip)` is nonzero.

## Cause

`interior_leader_candidates()` treats every feature attachment like a corner: it keeps the tip fixed and rotates the elbow ray through eight angles. That is valid for a point/corner attachment but not for a circle, where every alternative direction has a corresponding physical rim site and normal.

## Acceptance

1. Circular producers carry typed projected center/radius evidence into the shared interior-candidate adapter.
2. Every generated circular candidate keeps its tip on the proved rim.
3. Every circular shaft is collinear with the center-to-tip radius and points outward.
4. No raw coordinates or model-specific placement rule is added.
5. Grouped/split holes retain exact physical member ownership.
6. Exterior fallback, bounded-solver behavior, deferred replay semantics, and CTC01 interior placement remain covered.
7. The A3 demo bracket is lint-clean with `ø20`, `2× ⌀8 THRU`, and `⌀10 THRU` all at zero radial cross-product error.

pzfreo:
## Summary

- carry typed projected circle targets into the shared interior leader adapter
- move each rotated circular candidate to its corresponding rim site, keeping the shaft radial and outward
- preserve exact member ownership for grouped and split hole callouts
- update layout/replay contracts for solver-proven interior alternatives and bounded fallback

Closes #1760.

This does not address #1752, which remains blocked on the provider field needed for a stadium-slot end-radius requirement.

## Review notes

The change is shared geometry, not a demo-bracket special case: circular producers provide only center/radius evidence; the existing solver still chooses the candidate. Non-circular and profiled-hole candidates retain their prior behavior.

Dense circular inventories now contain genuinely different rim sites, so their connected assignment can honestly reach the existing state budget. The documented greedy fallback remains complete in the 12-callout regression and does not weaken the producer floor.

## Validation

- `uv run scripts/pr-check --static`
- `uv run pytest tests/ -q -n auto --dist worksteal`
  - 9,012 passed
  - 6 skipped
  - 14 expected failures
- A3 `draftwright-demo-bracket.step`: clean lint, 14/14 audited recognized requirements placed
- radial cross-product error is zero for `m_bossdia_z0`, `hc_plan0`, and `hc_plan1`


pzfreo:
Review pass completed against Google's code-review dimensions (design, functionality, complexity, tests, naming, comments, style, documentation, and every changed line): https://google.github.io/eng-practices/review/reviewer/looking-for.html

Findings resolved before submission:

1. A first implementation retargeted every candidate from one fixed tip but lost exact physical ownership for split declared holes. The final producer resolves the owner at each member location; the existing diameter-target lint regression now passes.
2. Constraining a circle to only its original normal avoided oblique arrival but caused the demo's central `⌀10` leader to fall outside and retain a Policy-B crossing. The final typed target exposes eight actual rim sites, each with its own outward normal, so the shared solver retains useful freedom without arbitrary coordinates.
3. The denser physical candidate graph can reach the existing exact-search state budget. The trace/test now states that honestly and proves the bounded greedy incumbent still places all 12 dense callouts; the candidate-measurement floor remains unchanged.
4. Deferred callout-only replay and the automatic drawing have different fixed annotation inventories after #1738. The replay test now checks semantic carrier parity and honest drops instead of accidentally requiring identical page lanes.
5. The historical boundary-elbow test now accepts either the exterior boundary route or a solver-proven interior label, while retaining the actual no-silhouette-crossing contract.

No blocking findings remain. The remaining tradeoff is the documented, non-blocking bounded-search fallback for unusually dense connected circular inventories; it preserves completeness and reports `optimal=false` rather than claiming an unproved optimum.

PR #1764, opened 2026-09-21
feat: dimension obround slot end radii

### LINKED ISSUE #1752: Stadium slots are dimensioned with no end-radius callout

pzfreo:
## Symptom and evidence

- User-visible problem: a proved stadium/obround through-slot was dimensioned only by width, length, and position, so its drawing did not distinguish it from a rectangular slot.
- Fixture/reproducer: `/Users/paul/Downloads/draftwright-demo-bracket.step`, whose upright plate contains a 26 × 8 obround slot.
- Before/after result: the slot previously had no end-shape statement; it now carries a solver-placed `2× R4` leader that meets a proved semicircular cap normally. The A3 demo has clean lint.
- Mutation that makes the guard fail: removing `m_slot0_radius` changes `slot_end_radius.radius` from `placed` to `missing`; moving a radius tip off the proved rim or making its shaft oblique fails the geometry assertions.

## Contract and scope

- Pins Quiddity 0.3.2 and consumes `Slot.end_radius` schema 2 without re-deriving geometry in Draftwright.
- Carries the conditional parameter through IR, planner, automatic rendering, deferred feature edits, completeness, declared `Sheet.slot(end_radius=...)`, and generated Sheet scripts.
- Uses typed radial candidates in the shared feature-leader solve. `auto`, `interior`, and `exterior` remain document policies; there is no raw-coordinate or CTC-01-specific placement path.
- Refuses Quiddity 0.3.2 hole-pattern relations whose members collapse onto one end-view axis, preserving ordinary grouped-hole ownership rather than emitting an undrawable pitch.
- Relevant architecture: ADR 1 (compiler pipeline), ADR 2 (layout), ADR 3 (recognition boundary), ADR 4 (declared intent), ADR 5 (honest failure).
- Explicitly deferred: obround pockets and rounded-corner rectangular-slot radius callouts.
- Stacked on #1763, which supplies the public leader-region policy.

Closes #1752.

## Validation

- [x] `scripts/pr-check --quick` — 98 passed; static analysis, spelling, and strict docs build clean.
- [x] Full local suite — 9,033 passed, 6 skipped, 14 xfailed.
- [x] The named mutation was observed failing before the implementation passed it.
- [x] Automatic, declared, deferred-edit, and generated-script paths were checked.
- [x] Recognition and repeated-lint behavior were checked where affected.
- [x] A3 demo exported to PDF/PNG and visually inspected; `lint()` is clean.

## Stop check

- [x] No two consecutive review rounds invalidated the same core association strategy.
- [x] The Quiddity 0.3.2 compatibility prerequisite is contained and covered in this adoption PR.

- [x] I have read the [CLA](https://github.com/pzfreo/draftwright/blob/main/CLA.md) and agree to it. If I am contributing on behalf of an organisation, I am authorised to agree on its behalf.


pzfreo:
Final review pass (Google engineering-practices order: correctness, design, complexity, tests, naming/comments, docs):

- Correctness: updated CTC-01's independent requirement set from 78 to 79 and named the new `slot_end_radius.radius`; the real-part canary passes.
- API/round-trip: accepted Quiddity's one-unit 0.01 mm publication discrepancy between radius and doubled width (for example 6.35 / R3.17), while retaining fail-fast validation for materially invalid caps and too-short slots.
- Design/complexity: removed an unnecessary generic radial lane-expansion change. Obrounds now contribute a small, bounded set of normal-to-arc semantic candidates; the existing shared solver and document `auto` / `interior` / `exterior` policy remain unchanged.
- Tests/docs: automatic, declared, generated-script, deferred-edit, pattern, tolerance, completeness, policy, and real-part paths are covered. A3 demo lint is clean.

Final CI is fully green across lint, changed-line coverage, Python 3.10–3.14 shards, macOS/Windows canaries, and the real-part canary. No substantive or nit-level findings remain in this review pass.

